PR #193, opened 2014-11-13
[WIP] Move 'js: true' line from describe to it blocks in an attempt to address...

jkeck:
... intermittent travis failures.


jkeck:
I think this is a non-starter as it didn't seem to help at all.  Closing.

PR #234, opened 2015-02-13
Rescue error when an unembeddable object is requested (i.e. has no contentMetadata type)

jkeck:
This creates an error class that gets called when an object w/o contentMetadata (e.g. a collection) is embedded.

This error is rescued in the embed controller and returns a 400.


jchristo4:
Looks good :+1:

PR #238, opened 2015-06-19
Adds Teaspoon JavaScript testing framework to project.

mejackreed:
- Adds jQuery gem into project, just so it can be injected into Teaspoon tests
 - Modifies default rake task to run teaspon and rspec
 - Create js specs for image and file, that change for included modules to be defined

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,3 @@ rvm:
 env:
   global:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
-
-script: bundle exec rspec spec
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -8,4 +8,4 @@ env:
   global:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
 
-script: bundle exec rspec spec
\ No newline at end of file
+script: bundle exec rake
```
cbeer: This should be the default -- remove it?

mejackreed:
@cbeer thanks for that... should be good to go.

@jkeck what do you think?

jkeck:
:+1: this will definitely make it easier for you guys to start writing useful JS unit tests as you start making enhancements/changes.

jkeck:
@mejackreed needs a rebase.

PR #241, opened 2015-06-21
[DO NOT MERGE] Add React.js and required components in, currently only replacing the…

mejackreed:
… EmbedFooter

 - Adds es5-shim/sham for IE8 and phantomjs support
 - Adds react-rails gem to application
 - Ruby viewers have react-rails ViewHelpers (and dependencies) included for server rendering
 - Adds a 'react' route to aid in development of React.js components
 - Poltergeist js errors are raised to Ruby for better debugging
 - EmbedFooter is rewritten in React.js

Notice propTypes support in React components which allow for specifying types in supplied properties

Some quick file size comparisons. Note that the react library is included here. We can look into optimizing this by using the cdn or removing unused parts of react

filename | master | move-to-react
----------- | --------- | ------------------
file.js (unminified) | 62.7 kb | 792 kb
file.js (minified) | 9.2 kb | 172 kb (estimated)
embed payload | 12.1 kb | 11.5 kb

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,9 @@ gem 'filesize'
 
 gem 'bower-rails'
 
+# Provides reactjs support
+gem 'react-rails'
+
 # To use ActiveModel has_secure_password
 # gem 'bcrypt-ruby', '~> 3.1.2'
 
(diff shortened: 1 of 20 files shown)


aeschylus:
Can I close these? They will still be stored here for reference under the closed filter.

PR #260, opened 2015-07-07
use travis container infrastructure for faster build

mejackreed:


aeschylus:
Decreased coverage is not meaningful in this case.

PR #261, opened 2015-07-07
move `CommonViewer` from a module to a class

### LINKED ISSUE #219: Make CommonViewer a class
`CommonViewer` started off (probably acceptably) as a mixin.  However; it has really taken on the shape of a super-class for the viewers themselves.  I would recommend making `CommonViewer` a class and subclassing all of the other viewers from it instead of including it as a module.

https://github.com/sul-dlss/sul-embed/blob/master/lib/embed/viewer/common_viewer.rb#L4-L7

mejackreed:
Closes #219 

ping @jkeck

### DIFF
--- a/lib/embed/viewer/common_viewer.rb
+++ b/lib/embed/viewer/common_viewer.rb
@@ -1,6 +1,6 @@
 module Embed
   class Viewer
-    module CommonViewer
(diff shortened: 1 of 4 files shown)


jkeck:
:+1:

PR #273, opened 2015-07-09
Issue 271

aeschylus:


app/models/embed/purl.rb
```diff
@@ -13,9 +13,9 @@ def title
 
     def type
       @type ||= begin
-        contentMetadata = ng_xml.xpath('//contentMetadata').first
-        contentMetadata.attributes['type'].try(:value) if contentMetadata.present?
-      end
+                  contentMetadata = ng_xml.xpath('//contentMetadata').first
```
mejackreed: Do we need to modify this whitespace here?

app/models/embed/purl.rb
```diff
@@ -42,10 +42,10 @@ def license
 
     def embargo_release_date
       @embargo_release_date ||= begin
-        if embargoed?
-          ng_xml.xpath('//rightsMetadata/access[@type="read"]/machine/embargoReleaseDate').try(:text)
-        end
-      end
+                                  if embargoed?
```
mejackreed: Same about whitespace

app/models/embed/purl.rb
```diff
@@ -96,20 +102,20 @@ def purl_xml_url
...
-        end
-        raise ResourceNotAvailable unless response.success?
-        response.body
-      rescue Faraday::Error::ConnectionFailed => error
-        nil
-      rescue Faraday::Error::TimeoutError => error
-        nil
-      end
+                      conn = Faraday.new(url: purl_xml_url)
+                      response = conn.get do |request|
```
mejackreed: Are we moving this for a reason?

spec/models/embed/envelope_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "rails_helper"
+
+describe Embed::Envelope do
+  include PURLFixtures
+  describe 'to_bounding_box' do
```
mejackreed: Could we dry up these spec using rspec's `let`? We can go over an example.

app/models/embed/purl.rb
```diff
@@ -183,9 +189,14 @@ def has_image_data?
...
           @file.xpath('./location[@type="url"]').present?
         end
+        def has_geo_data?
+          @file.xpath('//*[name()="gml:Envelope"]').present?
+        end
         def preview_types
           ["image/jp2"]
         end
+        def bounding_points 
+        end
```
aalsum: Remove bounding_points function

app/models/embed/purl.rb
```diff
@@ -183,9 +189,14 @@ def has_image_data?
         def has_location_data?
           @file.xpath('./location[@type="url"]').present?
         end
+        def has_geo_data?
```
mejackreed: I think this method is noop too. Either should be moved and tested/used, or removed

mejackreed:
@aeschylus https://github.com/sul-dlss/sul-embed/pull/278 should fix the errors with the JS tests that are failing.

https://github.com/sul-dlss/sul-embed/pull/276 refactors/drys up some of the specs in this PR

aeschylus:
I'm not getting these failing tests locally, and I believe the travis updates have already been applied.

PR #274, opened 2015-07-09
Exchange Geo (Raster) druid between public/restricted

aalsum:
The public and restricted druid in the sandbox page was exchanged.


mejackreed:
:+1: good catch!

PR #276, opened 2015-07-09
refactor envelope specs

mejackreed:
@aeschylus @aalsum 

A quick refactor of the envelope spec. This refactor:
 - isolates and tests only the public methods on the `Envelope` class.
 - isolates tests within `PURL` that deal with `bounding_box`


mejackreed:
Note: the tests will not pass until this gets added in https://github.com/sul-dlss/sul-embed/pull/278 :frog:

Issue #275, opened 2015-07-09
Extract public and restricted status from Purl for geo objects


aalsum:
Public druid PURL xml has:

```
<rightsMetadata>
<access type="read">
<machine>
    <world/>
</machine>
</access>
</code>
```

Restricted druid  PURL xml has:
```
<rightsMetadata>
<access type="read">
<machine>
    <group>Stanford</group>
</machine>
</access>```

jkeck:
Check out https://github.com/sul-dlss/dor-rights-auth

Not sure if this kind of work has already been done there or not.

aalsum:
This issue is already closed by #277  pull request

PR #277, opened 2015-07-09
Add public? method to purl model to return true for public object

aalsum:
Close #277

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -60,7 +60,10 @@ module Embed
       "#{Settings.purl_url}/#{@druid}"
     end
 
-    # @return true if the object is publicy accessible
+    ##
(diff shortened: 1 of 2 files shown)


app/models/embed/purl.rb
```diff
@@ -60,6 +60,11 @@ def purl_url
       "#{Settings.purl_url}/#{@druid}"
     end
 
+    # @return true if the object is publicy accessible
```
mejackreed: Could the yardoc be written like?

```ruby
##
# Returns true if the object is publicly accessible
# @return [Boolean]
```

mejackreed:
Closes #275

mejackreed:
@aalsum just to note, it looks like this method may also be helpful: https://github.com/sul-dlss/sul-embed/blob/master/app/models/embed/purl.rb#L184-L186

PR #278, opened 2015-07-09
isolate javascript tests from eachother

mejackreed:
Isolation of the JavaScript tests is needed to prevent Teaspoon errors and conflicts.


spec/teaspoon_env.rb
```diff
@@ -38,7 +38,7 @@
 
     # Specify a file matcher as a regular expression and all matching files will be loaded when the suite is run. These
     # files need to be within an asset path. You can add asset paths using the `config.asset_paths`.
-    suite.matcher = "{spec/javascripts,app/assets}/**/*_spec.{js,js.coffee,coffee}"
+    suite.matcher = '{spec/javascripts,app/assets}/*_spec.{js,js.coffee,coffee}'
```
jkeck: Did you happen to try changing this line to be:

```
suite.matcher = 'spec/javascripts/**/*_spec.{js,js.coffee,coffee}'
```

So that app/assets isn't loaded when teaspoon runs and the app/assets files can be required per-test?  I may try this locally as well in sul-requests to see if we can address that (although it hasn't particularly been a problem for us so far).

mejackreed: Yeah I did, that doesn't solve the problem (you would think so right?).

jkeck: :+1: Yeah, totally.  Good to know though! Thanks.

PR #280, opened 2015-07-10
add coffee-rails for deployment dependency - teaspoon

mejackreed:
Needed for deployment to dev environment

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,9 @@ group :development, :test do
 
   # Have jQuery local for testing
   gem 'jquery-rails'
+
+  # Needed for deploying to dev/test
+  gem 'coffee-rails'
 end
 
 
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -31,6 +31,9 @@ group :development, :test do
 
   # Have jQuery local for testing
   gem 'jquery-rails'
+
+  # Needed for deploying to dev/test
+  gem 'coffee-rails'
```
jkeck: Really?  That's unfortunate if we're not actually using it. :frowning:

Rails--

mejackreed: #teaspoon

jkeck: Ahhhh, that makes sense why it wasn't needed before.  I redact my criticism of rails.  Rails is Life. Rails is Love.

mejackreed: :tropical_drink: keep drinking the KoolAid

jkeck: Oh yeah!

PR #284, opened 2015-07-10
fixes some spacing issues introduced

mejackreed:


app/models/embed/purl.rb
```diff
@@ -199,14 +199,9 @@ def has_image_data?
         def has_location_data?
           @file.xpath('./location[@type="url"]').present?
         end
-        def has_geo_data?
```
cbeer: Just checking if you mean to remove these methods?

app/models/embed/purl.rb
```diff
@@ -112,20 +112,20 @@ def purl_xml_url
...
+        conn = Faraday.new(url: purl_xml_url)
+        response = conn.get do |request|
+          request.options = {
+            timeout: 2,
+            open_timeout: 2
+          }
+        end
+        raise ResourceNotAvailable unless response.success?
+        response.body
+      rescue Faraday::Error::ConnectionFailed => error
```
cbeer: Not that it's new, but this could be:

```
rescue Faraday::Error::ConnectionFailed, Faraday::Error::TimeoutError => error
```

app/models/embed/purl.rb
mejackreed: Yes, they are noops

PR #285, opened 2015-07-10
[For comparison only] Issue 249

### LINKED ISSUE #249: Add link out toolbar button

aalsum:
Close #249


lib/embed/viewer/common_viewer.rb
```diff
@@ -54,7 +54,11 @@ def footer_html
               end
               if self.is_a?(Embed::Viewer::Image) && !@request.hide_download?
                 doc.button(class: 'sul-embed-footer-tool sul-embed-btn sul-embed-btn-xs sul-embed-btn-default fa fa-download', 'aria-expanded' => 'false', 'data-sul-embed-toggle' => 'sul-embed-download-panel')
-              end
+              end  
+              if external_url.present?              
+               doc.button(class: 'sul-embed-footer-tool sul-embed-btn sul-embed-btn-xs sul-embed-btn-default fa fa-external-link', 'aria-expanded' => 'false', 
+                  onclick:"location.href='#{external_url}'",)
```
mejackreed: Implementing an external with a button this way may an issue with accessibility. Could we try something like this?

```ruby
doc.a(class: 'sul-embed-footer-tool sul-embed-btn sul-embed-btn-xs sul-embed-btn-default fa fa-external-link', href='#{external_url}'", target='_top')
```

I'm not sure if the styling works here, without pulling it down.

aalsum: Unfortunately, the style doesn't work. It shows blank box for the default mode and shows the right behavior with hover mode.

mejackreed: Ok, i think we should try to proceed with that `<a>` approach and work on styling. I'll pull it down and mess around with it.

PR #293, opened 2015-07-13
Issue 290 - ImageX viewer stub

### LINKED ISSUE #290: Stub out new image viewer in Ruby
- New image viewer should register itself for `:image, :manuscript, :map, :book`
 - Should provide the manifest.json url as a data attribute
 - ~~Should also have the label~~ See https://github.com/sul-dlss/sul-embed/commit/c1f63a94408b6f91100e875dfa132c8b79de4b5d#commitcomment-12183814
 - Should also have top controls (two-page, single-page, gallery, fullscreen) placeholders

http://purl.stanford.edu/fw090jw3474/iiif/manifest.json

aalsum:
Close #290

### DIFF
--- a/spec/lib/embed/viewer/image_spec.rb
+++ b/spec/lib/embed/viewer/image_spec.rb
@@ -14,6 +14,7 @@ describe Embed::Viewer::Image do
   end
   describe 'self.supported_types' do
     it 'should return an array of supported types' do
+      pending(' is pending due to the migration to imageX viewer')
       expect(Embed::Viewer::Image.supported_types).to eq [:image, :manuscript, :map, :book]
     end
   end
--- a/app/assets/javascripts/image_x.js
+++ b/app/assets/javascripts/image_x.js
@@ -1,0 +1,12 @@
+//= require modules/css_injection
+//= require vendor/tooltip
+//= require modules/common_viewer_behavior
+//= require modules/popup_panels
+//= require modules/embed_this
+//= require modules/image_x_viewer
+
+CssInjection.injectFontAwesome();
+CssInjection.appendToHead();
+CommonViewerBehavior.initializeViewer(ImageXViewer.init);
+PopupPanels.init();
+EmbedThis.init();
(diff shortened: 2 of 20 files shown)


spec/lib/embed/viewer/image_spec.rb
```diff
@@ -13,9 +13,10 @@
     end
   end
   describe 'self.supported_types' do
-    it 'should return an array of supported types' do
-      expect(Embed::Viewer::Image.supported_types).to eq [:image, :manuscript, :map, :book]
-    end
+    pending(' is pending due to the migration to collection viewer')
```
mejackreed: Could we make the pending specs like this?

```ruby
describe 'self.supported_types' do
  pending 'should return an array of supported types' do
    expect(Embed::Viewer::Image.supported_types).to eq [:image, :manuscript, :map, :book    
  end
end

# or

describe 'self.supported_types' do
  it 'should return an array of supported types' do
    pending(' is pending due to the migration to collection viewer')
      expect(Embed::Viewer::Image.supported_types).to eq [:image, :manuscript, :map, :book]
   end
end
```

That way only 1 line is changing?

lib/embed/viewer/collection.rb
```diff
@@ -0,0 +1,25 @@
...
+  class Viewer
+    class Collection < CommonViewer
+
+      def initialize(*args)
+        super
+      end
+
+      def body_html
+        Nokogiri::HTML::Builder.new do |doc|
+          doc.div(class: 'sul-embed-body sul-embed-collection', 'style' => "max-height: #{body_height}px", 'data-sul-embed-theme' => "#{asset_url('geo.css')}") do
```
mejackreed: We are embedding the `geo.css` style here, should we just go ahead and stub out the new viewer style?

mejackreed:
A few minor style comments, and requesting adding the style stubs. I do have one concern about the naming here. Even though there are eventual plans to change it, `Collection` does not feel descriptive enough and is a loaded term in our library. Could we change the name here? Even something like  `ImageCollection` feels a bit more descriptive to someone not part of our team.

mejackreed:
Naming things :smile: 

https://xkcd.com/910/

mejackreed:
:+1: looks good to me. Do you mind squashing the commits down please?

PR #300, opened 2015-07-16
update app to use sul-styles gem

mejackreed:
Rather than maintaining the common sul colors in multiple places, use the sul-styles gem


aeschylus:
Some simple variable updates. Looks good to me.

PR #315, opened 2015-07-28
Read object title from dc:title instead of objectLabel

### LINKED ISSUE #307: Embed viewer should display title based on dc:title instead of objectLabel

aalsum:
Close #307

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -8,7 +8,7 @@ module Embed
     end
 
     def title
-      @title ||= ng_xml.xpath('//identityMetadata/objectLabel').try(:text)
+      @title ||= ng_xml.xpath('//dc:title','dc'=>'http://purl.org/dc/elements/1.1/').try(:text)
     end
 
     def type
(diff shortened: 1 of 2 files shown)


lib/embed/viewer/common_viewer.rb
```diff
@@ -238,7 +238,7 @@ def header_title_logic
 
       def header_title_html(doc)
         doc.span(class: 'sul-embed-header-title') do
-          doc.text @purl_object.title
+          doc.text @purl_object.ng_xml.xpath('//dc:title','dc'=>'http://purl.org/dc/elements/1.1/').text
```
jkeck: Perhaps the change to the definition of what the PURL object's title is should be in the PURL object model: https://github.com/sul-dlss/sul-embed/blob/master/app/models/embed/purl.rb#L10-L12

aalsum: Completely agree.

mejackreed: :+1: this suggestion

app/models/embed/purl.rb
```diff
@@ -8,7 +8,7 @@ def initialize(druid)
     end
 
     def title
-      @title ||= ng_xml.xpath('//identityMetadata/objectLabel').try(:text)
+      @title ||= ng_xml.xpath('//dc:title','dc'=>'http://purl.org/dc/elements/1.1/').text
```
mejackreed: Do we need to guard against the xpath not being there? I'm thinking we should if possible using `try(:text)` rather than getting an error.

mejackreed:
:+1:

PR #321, opened 2015-07-28
update application to use OSD v2.0.0

mejackreed:
Dev's may need to run:

`bundle exec rake tmp:clear`

to update the cache. I'm of course assuming we want to update OSD from v1.2.0 to 2.0.0


aeschylus:
Weird, looks like this didn't merge earlier. Was it re-opened?

PR #322, opened 2015-07-29
This commit does several things related to embargoed objects to display:

### LINKED ISSUE #302: Handle the case of an embargo'ed file
If a file is under embargo and is listed in the embed viewer, it should be clear to the user that file access is limited.

The item's public XML includes the embargo release date, so perhaps this data can be leveraged for the desired functionality.

mejackreed:
- Removes tooltips on files for embargoed content
 - Adds a message to the file viewer specific to a global embargo and those with Stanford view rights
 - Removes links to files that are "embargoed to the world (everybody)"
 - Adds a method to the `CommonViewer` class which parses and returns a standardized date format that should be used globably for dates

Tooltip behavior on links is deprecated, thus the removal of the feature spec for it. This behavior is still present on the legacy image viewer download panel.

Closes #302 

### mg662qq6088 - access none with embargo release date
screen shot 2015-07-29 at 10 42 34 am

### km993dd6050 - access stanford with embargo release date
screen shot 2015-07-29 at 10 43 02 am

### nd387jf5675 - no embargo
screen shot 2015-07-29 at 10 43 38 am

cc @rosy1280 and @hannahfrost


lib/embed/viewer/common_viewer.rb
```diff
@@ -199,6 +199,14 @@ def file_url(title)
       end
 
       ##
+      # Creates a pretty date in a standardized sul way
+      # @param [String] date_string
+      # @return [String]
+      def sul_pretty_date(date_string)
+        Date.parse(date_string).strftime('%d-%b-%Y')
```
cbeer: Can you use `localize` and i18n for the formatting?

lib/embed/viewer/file.rb
```diff
@@ -111,6 +107,80 @@ def self.supported_types
       end
 
       private
+      ## 
+      # Creates an embargo message to be displayed, customized for stanford
+      # only embargoed items
+      # @return [String]
+      def embargo_message
+        message = ['Access is restricted', pretty_embargo_date]
+        unless !@purl_object.stanford_only_unrestricted?
```
cbeer: `if`?

lib/embed/viewer/common_viewer.rb
mejackreed: Its possible but then we need to mixin ActionView::Helpers::TranslationHelper. Just one of those trade off things :/  @jkeck any opinion on that?

jkeck: I think adding `I18n` should be fine and I don't think we'll need to include the translation helper if we access it directly though `I18n` (e.g. `I18n.l(Date.parse(date_string), format: :something)`)

mejackreed: :+1: thanks for the suggestions, updated accordingly

app/models/embed/purl.rb
```diff
@@ -2,6 +2,7 @@ module Embed
   class PURL
     require 'dor/rights_auth'
     delegate :embargoed?, to: :rights
+    delegate :stanford_only_unrestricted?, to: :rights
```
jkeck: I think this could be in one delegate line.  Not sure if it reads better that way or this way, so totally up to you.

lib/embed/viewer/file.rb
```diff
@@ -111,6 +107,80 @@ def self.supported_types
...
 
       private
+      ## 
+      # Creates an embargo message to be displayed, customized for stanford
+      # only embargoed items
+      # @return [String]
+      def embargo_message
+        message = ['Access is restricted', pretty_embargo_date]
+        if @purl_object.stanford_only_unrestricted?
+          message[0] = message[0] + ' to Stanford-affiliated patrons'
```
jkeck: I think `message[0] << ' to Stanford-affiliated patrons'` may be a bit more idiomatic here.

app/models/embed/purl.rb
mejackreed: :+1:

lib/embed/viewer/file.rb
mejackreed: :+1:

rosy1280:
Looks good to me.

jkeck:
:shipit: ?

mejackreed:
:+1: @jkeck

rosy1280:
no :cake: for you @jkeck !

PR #323, opened 2015-07-30
fixes a bad test that hard coded a date

mejackreed:
This snuck in yesterday, which will cause the tests to fail on days that Are not July 29. :bow:

### DIFF
--- a/spec/lib/embed/viewer/common_viewer_spec.rb
+++ b/spec/lib/embed/viewer/common_viewer_spec.rb
@@ -134,7 +134,7 @@ describe Embed::Viewer::CommonViewer do
 
   describe '#sul_pretty_date' do
     it 'parses a date into a standardized format' do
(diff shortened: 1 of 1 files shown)


spec/lib/embed/viewer/common_viewer_spec.rb
```diff
@@ -134,7 +134,7 @@
 
   describe '#sul_pretty_date' do
     it 'parses a date into a standardized format' do
-      expect(file_viewer.sul_pretty_date((Time.now + 1.month).to_s)).to eq '29-Aug-2015'
+      expect(file_viewer.sul_pretty_date((Time.now + 1.month).to_s)).to match /^\d{2}-\w{3}-\d{4}$/
```
aalsum: Why do we have + 1.month?

mejackreed: No real reason, it's just what we are using in other places to do date checks.

PR #333, opened 2015-08-07
Show SLY sliders for was-seed-viewer

### LINKED ISSUE #288: Display the images list annotated by date in webarchiving-seed viewer
seed_viewer

aalsum:
Close #288

### DIFF
--- a/app/assets/javascripts/modules/was_seed_viewer.js
+++ b/app/assets/javascripts/modules/was_seed_viewer.js
@@ -1,9 +1,34 @@
+/*global Sly */
 
 (function( global ) {
   'use strict';
   var Module = (function() {
+
     return {
       init: function() {
+        var $wasThumbFrame = $('.sul-embed-was-seed');
+        var dataAttributes = $wasThumbFrame.data();
+
+        // Create the SLY scroll div on fly
+        var $wasThumbScroll = $('<div class="sul-embed-thumb-slider-scroll"><div class="sul-embed-thumb-slider-handle"/></div>');
+        $wasThumbScroll.insertAfter($wasThumbFrame);
+
+        var thumbSliderSly = new Sly($wasThumbFrame, {
+          horizontal: 1,
+          itemNav: 'forceCentered',
+          smart: 1,
+          activateMiddle: 1,
+          activateOn: 'click',
+          mouseDragging: 1,
+          touchDragging: 1,       
+          releaseSwing: 1,
+          scrollBar: $wasThumbScroll,
+          dragHandle: 1,
+          dynamicHandle: 1,
+          startAt: dataAttributes.sulThumbsListCount-2, //The focus should be on one before the last thumb
+          speed: 300,
+          elasticBounds: 1
+        }).init();
       }
     };
   })();
--- a/spec/features/was_seed_viewer_spec.rb
+++ b/spec/features/was_seed_viewer_spec.rb
@@ -1,11 +1,41 @@
 require 'rails_helper'
-include PURLFixtures
 
 describe 'was seed viewer public', js: true do
+  include PURLFixtures
+  include WasSeedThumbsFixtures
+
   before do
+    allow_any_instance_of(Embed::Viewer::WasSeed).to receive(:thumbs_list).and_return(get_thumbs_list_fixtures)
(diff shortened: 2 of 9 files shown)


aalsum:
Hi @mejackreed , the pull-request is ready for review

app/assets/javascripts/modules/was_seed_viewer.js
```diff
@@ -1,11 +1,36 @@
...
+
     return {
       init: function() {
-      }
-    };
+        var $wasThumbFrame = $('.sul-embed-was-seed');
+        var dataAttributes = $wasThumbFrame.data();
+
+        // Create the SLY scroll div on fly
+		var $wasThumbScroll = $('<div class="sul-embed-thumb-slider-scroll"><div class="sul-embed-thumb-slider-handle"/></div>');
```
mejackreed: There seems to be some spacing issues going on here.

spec/features/was_seed_viewer_spec.rb
```diff
@@ -2,10 +2,51 @@
...
+    it 'changes the active one with the select' do
+      expect(page.find('.active').text).to eq('12-Oct-2013')
+      page.first('.sul-embed-was-thumb-item').click
+      expect(page.find('.active').text).to eq('29-Nov-2012')
+    end
+  end
+
+  def get_thumbs_list_fixtures
+    thumbs_list = 
+    [
```
mejackreed: It seems like this data is repeated several times. Could you just include the `WasSeedThumbsFixtures` module and parse that json provided its `thumbs_list` method?

app/assets/javascripts/modules/was_seed_viewer.js
```diff
@@ -1,9 +1,34 @@
...
   'use strict';
   var Module = (function() {
+
     return {
       init: function() {
+        var $wasThumbFrame = $('.sul-embed-was-seed');
+        var dataAttributes = $wasThumbFrame.data();
+
+        // Create the SLY scroll div on fly
+        var $wasThumbScroll = $('<div class="sul-embed-thumb-slider-scroll"><div class="sul-embed-thumb-slider-handle"/></div>');
```
mejackreed: We don't seem to be testing for this scroll bar anywhere (if removed our tests would still pass). Should we add some js tests here?

aalsum: The scroll creation and display is tested here https://github.com/sul-dlss/sul-embed/blob/was_view_288/spec/features/was_seed_viewer_spec.rb#L22 
Do you think we need more complex scenarios for testing?

mejackreed: Ah gotcha thanks... missed that, thanks!

spec/features/was_seed_viewer_spec.rb
```diff
@@ -1,11 +1,40 @@
...
+
+  describe 'loading was-seed-viewer' do
+    it 'displays the viewer on the frame' do
+      expect(page).to have_css('.sul-embed-was-seed', count: 1, visible: true)
+      expect(page).to have_css('.sul-embed-was-thumb-item', count: 4, visible: true)
+      expect(page).to have_css('.sul-embed-was-thumb-item-div', count: 4, visible: true)
+    end
+    it 'loads the SLY slider and scroll' do
+      expect(page).to have_css('.sul-embed-thumb-slider-scroll', count: 1, visible: true)
+      sleep(5)
```
mejackreed: Is this line necessary? I'm less than thrilled about including this if we don't need to?

aalsum: The test was not stable without this delay.

app/assets/javascripts/modules/was_seed_viewer.js
mejackreed: Also to note, it may be a way to speed up the tests ... for each individual feature test, it creates a lot of overhead. So for each `it` block the test framework is doing quite a bit just to test for the scrollbar for example. Testing the scrollbar as a feature test works, but may make sense later as a js test... #Tradeoffs :) happy to discuss in office tomorrow, but I definitely don't think it should block this PR.

PR #352, opened 2015-08-18
handle file downloads for files without a filesize

### LINKED ISSUE #328: The file viewer should suppress the file size if no data is available
e.g. http://purl.stanford.edu/pf584py5361

mejackreed:
Closes #328

### DIFF
--- a/lib/embed/viewer/file.rb
+++ b/lib/embed/viewer/file.rb
@@ -167,7 +167,11 @@ module Embed
             doc.span(class: 'sul-embed-sr-only') do
               doc.text "Download item #{file_count}"
             end
-            doc.text pretty_filesize(file.size)
+            if file.size.blank?
+              doc.text 'Download'
+            else
+              doc.text pretty_filesize(file.size)
+            end
           end
         end
       end
--- a/spec/lib/embed/viewer/file_spec.rb
+++ b/spec/lib/embed/viewer/file_spec.rb
@@ -51,6 +51,15 @@ describe Embed::Viewer::File do
       html = Capybara.string(file_viewer.to_html)
       # visible false because we display:none the container until we've loaded the CSS.
       expect(html).to have_css '.sul-embed-file-list', visible: false
+      expect(html).to have_css 'a[download]', visible: false, text: '12.35 kB'
+    end
+    context 'without file size' do
+      it 'displays Download as the link text' do
+        stub_purl_response_and_request(file_purl_no_size, request)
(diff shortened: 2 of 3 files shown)


lib/embed/viewer/file.rb
```diff
@@ -167,7 +167,11 @@ def file_download_link(doc, file, file_count)
             doc.span(class: 'sul-embed-sr-only') do
               doc.text "Download item #{file_count}"
             end
-            doc.text pretty_filesize(file.size)
+            if file.size.blank? || file.size.empty?
```
cbeer: Doesn't `blank?` imply `empty?`?

mejackreed: Yeah.. your right. Thanks for catching that.

aalsum:
@mejackreed do you have an example of purl with this problem?

mejackreed:
See #328 http://purl.stanford.edu/pf584py5361

spec/lib/embed/viewer/file_spec.rb
```diff
@@ -51,6 +51,13 @@
       html = Capybara.string(file_viewer.to_html)
       # visible false because we display:none the container until we've loaded the CSS.
       expect(html).to have_css '.sul-embed-file-list', visible: false
+      expect(html).to have_css 'a[download]', visible: false, text: '12.35 kB'
+    end
+    it 'with no file size' do
```
aalsum: Could we make the description represents the test? For example, "Displays download instead of the file size for the files with no size"

mejackreed: I could, but part of me feels that rspec does this for us. For example with this test in a failure it reads:

```sh
Failures:

  1) Embed::Viewer::File body_html with no file size
     Failure/Error: expect(html).to have_css 'a[download]', visible: false, text: /^Download$/
       expected #has_css?("a[download]", {:visible=>false, :text=>/^Download$/}) to return true, got false
     # ./spec/lib/embed/viewer/file_spec.rb:61:in `block (3 levels) in <top (required)>'
```

And the hope is that the Rspec formatting should provide this to a developer?

What do you think?

aalsum: Sounds fine with me to make the test for the general behavior with no-file size.

cbeer: I think the idiomatic way is:

```
context 'without a file size' do
  it 'displays a download link' do
    ...
  end
end
```

mejackreed: Thanks @aalsum and @cbeer for your comments... updated!

PR #378, opened 2015-08-27
clicking overview button changes the perspective, also disables thumb…

### LINKED ISSUE #377: Hide bottom panel when entering the overview perspective.
This requires #371 to be completed.

mejackreed:
…sliderarea

I think this also may close #377

### DIFF
--- a/app/assets/javascripts/modules/canvas_store.js
+++ b/app/assets/javascripts/modules/canvas_store.js
@@ -31,16 +31,16 @@
         _this.selectCanvas(canvasId);
       });
       PubSub.subscribe('updateMode', function(_, newMode) {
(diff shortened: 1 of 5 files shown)


aeschylus:
Let me test it out. I'm a bit wary about the UI change being initiated directly from an event firing rather than from pulling down the current state. Merging, since it's probably robust enough for our purposes. With more things that can cause this behaviour, it would be error-prone to build the UI change into the event.

aeschylus:
Otherwise looks good. I'm moving on to the style fixes and dead code removal from that other pull request.

mejackreed:
Thanks for pointing that out. Should I add a ticket to clean up?

PR #391, opened 2015-08-28
add right click restriction for world restricted images

### LINKED ISSUE #387: Disable right click on canvas for images with zoom only restrictions
```xml
<group>Stanford</group>      <worldrule="no-download"/>
```
bb000kq3835

Revs use case, partly addresses #244

mejackreed:
Closes #387


aeschylus:
Cool. Looks good.

PR #399, opened 2015-09-01
adds fullscreen support in image viewer

### LINKED ISSUE #395: Add fullscreen for image viewer

mejackreed:
Closes #395

### DIFF
--- a/app/assets/javascripts/modules/image_x_viewer.js
+++ b/app/assets/javascripts/modules/image_x_viewer.js
@@ -64,6 +64,10 @@
         var perspective = $(this).data().sulViewPerspective;
         PubSub.publish('updatePerspective', perspective);
       });
+      embedHeader.find('[data-sul-view-fullscreen="fullscreen"]')
+        .on('click', function() {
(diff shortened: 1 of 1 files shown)


aeschylus:
Hm. Is this what we want? This will make the canvas fullscreen, but not include the title bar, controls, bottom panel or status bar. 

We can still use osd and achieve those other things, if they are desired. I can see arguments going both ways. For an existing implementation and background, see [this question](https://github.com/openseadragon/openseadragon/issues/669) from the osd issues, and [the lines](https://github.com/IIIF/mirador/blob/master/js/src/widgets/hud.js#L114-L124) in the Mirador source where we target the whole window.

mejackreed:
Good question, I'm not sure about it. I'll raise with @ggeisler and @jvine today.

mejackreed:
@aeschylus went down this path a bit. Unfortunately there are some styling complications that will cause issues. I think we should merge this and look at fullscreening everything later.

aeschylus:
Sounds good.

PR #429, opened 2015-09-15
Keyboard access

### LINKED ISSUE #143: Add keyboard navigation to image viewer views

### LINKED ISSUE #427: Keyboard control navigation

mejackreed:
Closes #427 , closes #143 

Adds support for basic left/right navigation in the following contexts:

context | key | action
---------- | ----- | -------
`bottomPanelOpen` | left | previous image
`bottomPanelOpen` | right | next image
`bottomPanelClosed` | left, right | no action
`overview perspective` | left | previous image
`overview perspective` | right | next image

These contexts are extendable and should be able to easily meet usability enhancement goals.

### DIFF
--- a/.jshintignore
+++ b/.jshintignore
@@ -1,3 +1,3 @@
-vendor
-app/assets/javascripts/vendor
-spec/javascripts/support
+vendor/*
+app/assets/javascripts/vendor/*
+spec/javascripts/support/*
--- a/app/assets/javascripts/image_x.js
+++ b/app/assets/javascripts/image_x.js
@@ -6,6 +6,7 @@
 //= require vendor/tooltip
 //= require vendor/sly
 //= require vendor/manifestor
+//= require vendor/keymaster
 //= require modules/common_viewer_behavior
 //= require modules/popup_panels
 //= require modules/embed_this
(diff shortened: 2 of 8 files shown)


mejackreed:
As far as I can tell, all of the hound violations are coming from the vendor javascript which should be ignored (but is not by hound :cry: ) :dog:

aeschylus:
Awesome. Why would the keyboard commands be different if the bottom panel is closed though? Is this when we would switch over to the osd canvas controls (panning instead of nexting)? Seems reasonable, but might confuse some people. I think we might eventually add another context for when the osd canvas is zoomed below its home bounds to use for that purpose.

jkeck:
Not sure if this helps but checkout 

https://github.com/sul-dlss/sul-requests/blob/master/.hound.yml#L11
https://github.com/sul-dlss/sul-requests/blob/master/.javascript_ignore

mejackreed:
Thanks @jkeck it worked!

mejackreed:
@aeschylus yep agree on the second context that is needed. I wrote it this way so we could easily extend the behavior to accommodate that or any other context. I'm figuring that there will be modifications that need to happen at some point.

PR #437, opened 2015-09-21
add left/right mouse buttons that fade out

### LINKED ISSUE #432: keyboard controls should work when film strip is hidden
keyboard controls only work when filmstrip is displayed.  the filmstrip conceals some of the image, so should not be the only mode in which keyboard controls work.

mejackreed:
Closes #432 , left/right are visible during mouse moves and then fadeout

screen shot 2015-09-21 at 2 22 04 pm


aeschylus:
Merging this. I will swoop back in and change out the next/previous for book view once I get that working in iiifManifestLayouts.

PR #460, opened 2015-10-19
Fix bottom image height

mejackreed:
screen shot 2015-10-19 at 8 17 40 am


aeschylus:
More ajax problems with the download panel. Is this still purl related?

mejackreed:
Yep... rerunning tests

PR #463, opened 2015-10-20
specify default_body_height for image_x viewer, which fixes embed pan…

mejackreed:
…el bugs

### DIFF
--- a/lib/embed/viewer/image_x.rb
+++ b/lib/embed/viewer/image_x.rb
@@ -122,6 +122,12 @@ module Embed
           str
         end
       end
+
+      ##
+      # Sets the default body height
+      def default_body_height
+        420
+      end
     end
   end
 end


aeschylus:
It clearly does, but how does this fix the width as well? Is the height is used to calculate the width somewhere?

mejackreed:
That happens here: https://github.com/sul-dlss/sul-embed/blob/master/lib/embed/viewer/common_viewer.rb#L203

mejackreed:
And by default width is `nil`: https://github.com/sul-dlss/sul-embed/blob/master/lib/embed/viewer/common_viewer.rb#L338

aeschylus:
Okay, I see that now. I didn't see where the height was coming from. So it just becomes a square by default. Seems reasonable.

PR #468, opened 2015-10-21
General cleanup

mejackreed:
- Removes pending specs that don't do anything.
 - Removes the legacy image viewer


aeschylus:
I don't think this coverage decrease is significant.

PR #492, opened 2015-12-14
update download links to match the new style expected by the stacks

peetucket:
anyone know why the three tests are failing?  i see them failing on other branches too, is there a known issue?

### DIFF
--- a/app/assets/javascripts/modules/download_panel.js
+++ b/app/assets/javascripts/modules/download_panel.js
@@ -29,7 +29,7 @@
           return;
         }
         var downloadUrl = imageData['@id'] + '/full/' + size.width + ',' +
-          size.height + '/0/default.jpg?action=download';
+          size.height + '/0/default.jpg?download=true';
         var list = $listMarkup.clone();
         var link = $('<a class="download-link" href="' + downloadUrl +
           '">Download (' + size.width + ' x ' + size.height + ')</a>');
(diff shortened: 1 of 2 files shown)


mejackreed:
Various reasons, but mostly having to do with last minute changes to the implemented IIIF-Auth.

peetucket:
so does that mean no changes (even simple ones like this) can be made before more time is devoted to this project?

mejackreed:
Yes. I need a few hours allocated to working on this.

app/assets/javascripts/modules/download_panel.js
```diff
@@ -22,14 +22,14 @@
         .append('<a class="download-link" href="' + imageData['@id'] +
         '/full/!' + disableDownloadWidthCutoff + ',' +
         disableDownloadWidthCutoff +
-        '/0/default.jpg" download>Download Thumbnail</a>'));
+        '/0/default.jpg?download=true" download>Download Thumbnail</a>'));
```
jkeck: I noticed that we missed this one.  I assume we want the `download=true` for thumbnails as well as the other downloadable derivatives?

mejackreed: Yep seems that way. @peetucket ?

peetucket: "no thumb for you!"
sorry, yes, I am sure this was just missed

PR #498, opened 2016-01-26
update rails to address vulnerabilities

mejackreed:


jkeck:
So, we should probably change all of our `*.css.scss` file to be just `*.scss`.

Is there any particular reason we're holding off on doing that?  It could probably be a separate commit in this PR or a separate PR all together.

I'm happy to do either, just wanted to make sure I asked before I plowed ahead.

mejackreed:
Go for it

PR #502, opened 2016-02-23
Add rubocop as a dependency and change the way it and hound are configured

jkeck:
Potentially controversial I guess. `¯\_(ツ)_/¯`


jkeck:
Just posting this to start the convo about the approach for rubocop integration, not necessarily trying to dictate how to do it.

mejackreed:
Yeah.. I'm really down for whatever. The previous attempt was more of a looking at what hound has as defaults and using that with a few minor changes.

I'm :ok_hand: with whatever as long as it's formalized.

PR #510, opened 2016-03-11
Add and configure Guard RSpec.

jkeck:


ndushay:
Do we have experience with guard?  Someone at the Sandi Metz class was having trouble with an editor or something because of guard ...

I'm willing to try it and see;  just wondering.

jkeck:
Well, the way this is implemented, you don't have to use it if it doesn't work for you.  This just adds the dependency and configuration so that if you want to run guard, you can.

I'm really imagining this being the kind of tool I turn-on when I'm refactoring a class or something like that.

PR #519, opened 2016-03-18
Add world restricted attribute to full-screen viewer and disable…

### LINKED ISSUE #488: Users can right click and download an image when in full screen
When viewing a no-download objects in full screen, a user can right click and save image as.  To re-create go to https://purl.stanford.edu/zz234zw6791, expand the image to full screen, right click on the image, choose save image as.  This works in Chrome and Firefox, but not Safari.

jkeck:
… right-clicks.

Fixes #488 

Paired w/ @ndushay on this

cc @mejackreed

### DIFF
--- a/app/assets/javascripts/modules/image_x_viewer.js
+++ b/app/assets/javascripts/modules/image_x_viewer.js
@@ -254,6 +254,7 @@
       $embedHeader.find('[data-sul-view-fullscreen="fullscreen"]')
         .on('click', function() {
           canvasStore.osd.setFullScreen(true);
+          _disableClicks();
       });
     };
 
@@ -392,11 +393,20 @@
       });
     };
 
+    var _defaultViewerOrFullpageViewer = function() {
+      if ($('.osd-container.fullpage').length > 0) {
+        return $('.osd-container.fullpage');
+      } else {
+        return $el;
+      }
+    };
+
     var _disableClicks = function() {
       if (dataAttributes.worldRestriction) {
-        $el.on('contextmenu', '.openseadragon-canvas', function(e) {
-          e.preventDefault();
-        });
+        _defaultViewerOrFullpageViewer()
+          .on('contextmenu', '.openseadragon-canvas', function(e) {
+            e.preventDefault();
+          });
       }
     };
 


app/assets/javascripts/modules/image_x_viewer.js
```diff
@@ -253,7 +253,10 @@
     var _setupButtonListeners = function() {
       $embedHeader.find('[data-sul-view-fullscreen="fullscreen"]')
         .on('click', function() {
+          var worldRestriction = $('[data-world-restriction]').data('world-restriction');
           canvasStore.osd.setFullScreen(true);
+          $('.osd-container').attr('data-world-restriction', worldRestriction);
```
jkeck: The `world-restriction` data attribute is not present on the full-screen OSD viewer so we need to add it there so `_disableClicks()` knows to disable the right-click.

app/assets/javascripts/modules/image_x_viewer.js
```diff
@@ -253,7 +253,10 @@
     var _setupButtonListeners = function() {
       $embedHeader.find('[data-sul-view-fullscreen="fullscreen"]')
         .on('click', function() {
+          var worldRestriction = $('[data-world-restriction]').data('world-restriction');
```
mejackreed: I think you can get away with removing this. This should already be stored in an object `dataAttributes.worldRestriction`.

app/assets/javascripts/modules/image_x_viewer.js
mejackreed: I think you can get away with removing this. This should already be stored in an object `dataAttributes.worldRestriction`.

app/assets/javascripts/modules/image_x_viewer.js
jkeck: :+1: Good call!

I think that was our first attempt at trying to figure out how to disable in full-screen, and when we figured out that we needed to disable the context menus scoped to a different element in full-screen, didn't go back to see if this was ever necessary.

mejackreed:
:+1:

ndushay:
@jkeck let's merge and deploy !

peetucket:
sweet

rosy1280:
Oh sure. Now that I'm gone you close my tickets. Punishment?

jkeck:
@rosy1280, I know!  We have been waiting for you to leave so we could fix all the problems you reported. 😜

peetucket:
Not only closed, but fixed too!

ndushay:
but @peetucket, we didn't deploy it yet.

PR #528, opened 2016-04-01
Install and configure DLSS Cops

### LINKED ISSUE #515: Install and configure DLSS Cops
The team should decide if there are things disabled in the baseline that we should re-enable (given that the code has already been operating under those rules previously).

jkeck:
There are commits that include running the autocorrector as well as some non-functional fixes that could not be addressed simply via autocorrecting.

This PR also changes the CI build to run rubocop.

Closes #515


jkeck:
Oh, that's right, hound doesn't honor the DLSS cops configuration.  _Wonderful._

Wonderful

jkeck:
Will re-open an identical PR to see if I can get Hound to respect DLSS Cops config in another way.

PR #532, opened 2016-04-06
Update media viewer with configurable streaming endpoint to generate...

### LINKED ISSUE #526: Connect video player to actual content (via public XML)
Pulling content from wowza server

This is related to #525

(was sul-dlss/media#8)

jkeck:
...`<source>` elements.

configured-media

Closes #526

### DIFF
--- a/spec/lib/embed/viewer/media_spec.rb
+++ b/spec/lib/embed/viewer/media_spec.rb
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 describe Embed::Viewer::Media do
+  include PURLFixtures
+
   let(:request) { Embed::Request.new(url: 'https://purl.stanford.edu/ignored') }
   let(:media_viewer) { Embed::Viewer::Media.new(request) }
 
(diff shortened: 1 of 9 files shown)


spec/lib/embed/viewer/media_spec.rb
```diff
@@ -17,6 +19,7 @@
   end
 
   describe '#body_html' do
+    before { stub_purl_response_with_fixture(video_purl) }
     it 'includes some dummy content' do
```
ndushay: would it be more appropriate to test for specific contents now that the fixture objects have them and the code exists to parse it?

ndushay:
just a question about tests.  o.w. looks good to me.

spec/lib/embed/viewer/media_spec.rb
jkeck: This test is ensuring that the `MediaTag` is being implemented correctly.  Since the `MediaTag` class is responsible for generating the markup w/i the `video`/`audio` tags, the tests are in that class: https://github.com/sul-dlss/sul-embed/blob/streaming-server-config/spec/lib/embed/media_tag_spec.rb

Perhaps what would be more useful is a change in the test description (although I didn't actually change the test functionally, just stubbed the PURL object since we're accessing it from the `MediaTag` class now.

jkeck:
@ndushay: Updated spec comment to match reality.

ndushay:
:+1:

jkeck:
@tingulfsen I updated this PR to address the style issue when a video was embedded with other content.

PR #537, opened 2016-04-11
Add download panel to the media viewer

### LINKED ISSUE #535: Add download panel to media viewer

jkeck:
Paired on a few of the commits w/ @tingulfsen.

Closes #535 

## bb582xs1304
bb582xs1304

### DIFF
--- a/lib/embed/viewer/image_x.rb
+++ b/lib/embed/viewer/image_x.rb
@@ -19,6 +19,10 @@ module Embed
         [:image, :manuscript, :map, :book]
       end
 
+      def self.show_download?
+        true
+      end
+
       def manifest_json_url
         "#{Settings.purl_url}/#{@purl_object.druid}/iiif/manifest.json"
       end
--- a/lib/embed/viewer/image_x.rb
+++ b/lib/embed/viewer/image_x.rb
@@ -64,21 +64,9 @@ module Embed
 
       def download_html
         return '' if @request.hide_download?
-        Nokogiri::HTML::Builder.new do |doc|
-          doc.div(class: 'sul-embed-panel-container') do
-            doc.div(class: 'sul-embed-panel sul-embed-download-panel', style: 'display:none;', 'aria-hidden': 'true') do
-              doc.div(class: 'sul-embed-panel-header') do
-                doc.button(class: 'sul-embed-close', 'data-sul-embed-toggle' => 'sul-embed-download-panel') do
-                  doc.span('aria-hidden' => true) do
-                    doc.cdata '&times;'
-                  end
-                  doc.span(class: 'sul-embed-sr-only') { doc.text 'Close' }
-                end
-                doc.div(class: 'sul-embed-panel-title') do
-                  doc.text 'Download image'
-                  doc.span(class: 'sul-embed-panel-item-label')
-                end
-              end
+        Embed::DownloadPanel.new(title: 'Download image') do
+          Nokogiri::HTML::Builder.new do |doc|
+            doc.div(class: 'sul-embed-panel-body') do
               doc.ul(class: 'sul-embed-download-list') {}
               @purl_object.contents.each do |resource|
                 next unless resource.type == 'object'
@@ -101,7 +89,7 @@ module Embed
                 end
               end
             end
-          end
+          end.to_html
         end.to_html
       end
 
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -48,6 +48,11 @@ def visit_sandbox
   visit page_path(id: 'sandbox')
 end
 
+def toggle_download_panel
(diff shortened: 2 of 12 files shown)


aeschylus:
This is an excellent forehead. Now downloadable.

lib/embed/viewer/image_x.rb
```diff
@@ -1,3 +1,5 @@
+require 'embed/download_panel'
```
ndushay: why does this viewer have this require statement, but geo viewer doesn't ?

spec/rails_helper.rb
```diff
@@ -48,6 +48,13 @@ def visit_sandbox
   visit page_path(id: 'sandbox')
 end
 
+def click_download_panel
+  expect(page).to have_css('.sul-embed-download-panel', visible: false)
+  page.find('button[data-sul-embed-toggle="sul-embed-download-panel"]', visible: true)
+  page.find('[data-sul-embed-toggle="sul-embed-download-panel"]', match: :first).click
+  expect(page).to have_css('.sul-embed-download-panel', visible: true)
+end
```
ndushay: I'm not so fond of embedding the expectations in this helper method;  I would be more in favor of having a test in a spec file with the two expect statements bracketing a call to a 2-line click-download-panel.   It would surface the test as a test, and it would avoid running the test for every click.

If this is something that the calling class needs to test ... then rename to click_toggable_download_panel   

Not a huge thing I feel strongly about ... just a thought.

spec/lib/embed/viewer/geo_spec.rb
```diff
@@ -50,13 +50,13 @@
     end
     it 'generates a file list when file has resources' do
       stub_purl_response_and_request(geo_purl_public, request)
-      html = Capybara.string(geo_viewer.download_html)
+      html = Capybara.string(geo_viewer.download_html.to_html)
```
ndushay: there's something uncomfortable to me about needing to call "to_html"  on a method called download_html.   Should the object be named download_panel?   Should "to.html" be called inside   download_html method?    Or is my ignorance showing?

ndushay:
I really like this approach.  It's a genius refactoring, IMO.   My line notes are all quibbles.  The PR as it stands is an improvement;   the only one that feels remotely important to me is about whether the require statement is needed.

spec/rails_helper.rb
jkeck: Totally good point.  My thinking here is, if I write a test that uses this helper (and am presumably making assertions about the content inside the download panel) I want my test to fail because of the expectations I make on the content.  If the test fails because the download panel failed to open/is not present, then the failure raised would point to the expectation in the helper instead of the expectation in my test.

I don't feel strongly about it either, though.  So totally willing to kill it.

spec/lib/embed/viewer/geo_spec.rb
jkeck: Yeah, I struggled with this interface a bit.  Here is what I was trying to accomplish with this, and let's see if we can come up with a more appropriate alternative.

The `download_html` method **used** to return a string (which happened to be the result of the 'to_html' method on the `Nokogiri::HTML::Builder` object).  This method now returns an instance of `Embed::DownloadPanel`, which is why we are calling `to_html` on the resulting object.

`Embed::DownloadPanel`'s responsibility is to return HTML which is why it has a `to_html` method.  The **actual** implementation is more of a string, so the class delegates `to_str` to `to_html` (which allows the `... << download_html << ...` to work in [`CommonViewer`](https://github.com/sul-dlss/sul-embed/blob/master/lib/embed/viewer/common_viewer.rb#L41)).

I can think of a few immediate options.

* Require the viewers implementing `Embed::DownloadPanel` to explicitly call `to_html` (which would mimic how we're using `Nokogiri::HTML::Builder`)
* Move everything into string methods like `to_s` (I feel like I went down this route at one point, but am willing to explore again).

lib/embed/viewer/image_x.rb
jkeck: :+1: Good :eyes:

I think this was just leftover from the `DownloadPanel` refactor.  I'll check that it all works w/o it and remove it.

spec/lib/embed/viewer/geo_spec.rb
ndushay: sounds like option 1 is a better option?

spec/rails_helper.rb
ndushay: "kill it" as in ... move on and forget about it?  I'm ok w that.  Slight pref for rename, but it's ok w me as is.

jkeck: Sorry, "Kill it" as in remove the expectation.

jkeck: And I will rename, agreed that it's not descriptive.

ndushay:
:+1:

PR #540, opened 2016-04-14
lib/embed.rb: use mattr_accessor instead of class variable

ndushay:
... for registered_viewers


ndushay:
closed in favor of #546

PR #541, opened 2016-04-14
Fix reader method names

ndushay:


ndushay:
closed in favor of #546

PR #542, opened 2016-04-14
Avoid is_ and has_ method names

ndushay:


ndushay:
closed in favor of #546

PR #543, opened 2016-04-14
Rails time zone fix in spec file

ndushay:


ndushay:
closed in favor of #546

PR #546, opened 2016-04-15
refactorings inspired by rubocop violations

ndushay:
- use mattr_accessor instead of class var in lib/embed
- fix reader method names (no get_  prefix)
- avoid is_ and has_  method names
- Rails time zone fix in spec file


spec/lib/embed/viewer/file_spec.rb
```diff
@@ -89,14 +89,14 @@
...
-      expect(html).to have_css('.sul-embed-embargo-message', text: "Access is restricted until #{(Time.now + 1.month).strftime('%d-%b-%Y')}")
+      expect(html).to have_css('.sul-embed-embargo-message', text: "Access is restricted until #{(Time.current + 1.month).strftime('%d-%b-%Y')}")
       expect(html).not_to have_css('a')
     end
     it 'adds a generalized embargo message and no links are present for an edge case ordering issue' do
       stub_purl_response_and_request(embargoed_edge_purl, request)
       expect(file_viewer).to receive(:asset_host).at_least(:twice).and_return('http://example.com')
       html = Capybara.string(file_viewer.body_html)
-      expect(html).to have_css('.sul-embed-embargo-message', text: "Access is restricted until #{(Time.now + 1.month).strftime('%d-%b-%Y')}")
+      expect(html).to have_css('.sul-embed-embargo-message', text: "Access is restricted until #{(Time.current + 1.month).strftime('%d-%b-%Y')}")
```
jkeck: Oddly, I'm getting test failures locally in this file.  It appears like we have changed `Time.now` to `Time.current` here but in the fixtures we're still using `Time.now` (https://github.com/sul-dlss/sul-embed/blob/master/spec/fixtures/purl_fixtures.rb#L287)

Should we change that to `Time.current` as well?

jkeck:
One comment about some failing tests I'm getting locally.

Otherwise :+1:

spec/lib/embed/viewer/file_spec.rb
ndushay: addressing this now.  Missed it b/c it didn't fail on my box, and because spec/fixtures  is excepted in rubocop.yml

ndushay:
Made the suggested change.  Merge at will (or find more opportunities for improvement!)

PR #547, opened 2016-04-15
Refactoring: extract EmbedThisPanel class from CommonViewer

### LINKED ISSUE #544: Refactor Embed-This Panel out of common viewer
use download-panel as exemplar of how to do this.

ndushay:
EmbedThisPanel class handles the embed panel markup.   It is modeled on the DownloadPanel class.

Closes #544


lib/embed/embed_this_panel.rb
```diff
@@ -0,0 +1,58 @@
+module Embed
+  class EmbedThisPanel
+    def initialize(druid:, height:, width:, purl_object_title:, &block)
```
ndushay: I couldn't think of a better way to get all this information to this class.  5 args, and 5 instance variable feels like sort of a lot.

jkeck:
@ndushay this needs a rebase after merging the rubocop fixes.

ndushay:
rebased.

lib/embed/embed_this_panel.rb
```diff
@@ -0,0 +1,58 @@
...
+                  <input type='checkbox' id='sul-embed-embed' data-embed-attr='hide_embed' checked=true />
+                  <label for='sul-embed-embed'>embed</label>
+                </div>
+                <div>
+                  <div class='sul-embed-options-label'>
+                    <label for='sul-embed-iframe-code'>Embed code:</label>
+                  </div>
+                  <div class='sul-embed-section'>
+                    <textarea id='sul-embed-iframe-code' data-behavior='iframe-code' rows=4>
+                      <iframe src='#{Settings.embed_iframe_url}?url=#{Settings.purl_url}/#{druid}' height='#{height}px' width='#{width || height}px' frameborder='0' marginwidth='0' marginheight='0' scrolling='no' allowfullscreen></iframe>
```
jkeck: Unfortunately it looks like the heredoc spacing in the `<textarea>` is making its way into the actual form.

textarea

lib/embed/embed_this_panel.rb
```diff
@@ -0,0 +1,58 @@
...
+                <span aria-hidden='true'>&times;</span>
+                <span class='sul-embed-sr-only'>Close</span>
+              </button>
+              <div class='sul-embed-panel-title'>Embed</div>
+            </div>
+            <div class='sul-embed-panel-body'>
+              <div class='sul-embed-embed-this-form'>
+                <span class='sul-embed-options-label'>Select options:</span>
+                <div class='sul-embed-section sul-embed-embed-title-section'>
+                  <input type='checkbox' id='sul-embed-embed-title' 'data-embed-attr'='hide_title' checked=true />
```
jkeck: The `data-embed-attr` should not have the single quotes around it.  This is currently causing the title option to add the parameter `undefined=true` instead of `hide_title=true` (there should have been a test for this, looks like that one got away from us).

undefined-title

jkeck:
@ndushay noticed a few things.  Some inline comments, but here are two screenshots that illustrate the issues.

### Expected
expected-embed

### Actual
undefined-title

lib/embed/embed_this_panel.rb
ndushay: ok, will fix.

jkeck: More than happy to to take a look at this with you on Monday!

ndushay:
@jkeck   I figured out how to see what you were looking at in my browser.  I made the corrections you indicated, and added a test for the hide_title checkbox.   I also rebased for good measure.   Let me know if you have any other suggestions.   I still haven't come up with a way to avoid 5 args and 5 instance vars, but since I think this PR is an improvement on the existing code base, maybe that's a problem for another PR.

ndushay:
tweaked it a bit more per F2F with @jkeck.

PR #554, opened 2016-04-22
Refactor MediaTag to use MediaStacks URLs instead of…

jkeck:
…Wowza URLs directly.


ndushay:
I paired with @jkeck on this work, and I approve this PR.  :+1:

PR #558, opened 2016-04-30
Add Dash.js reference player as MPEG-DASH prototype

### LINKED ISSUE #545: Implement prototype dash player to allow playback in all browsers
This allows video display support for browsers other than safari and ios.

jkeck:
Closes #545 

### Viewer examples in Chrome, Firefox, and Safari
multi-browser-video


lib/embed/media_tag.rb
```diff
@@ -19,22 +19,20 @@ def initialize(ng_doc, viewer)
...
-        label = resource.description
         next unless SUPPORTED_MEDIA_TYPES.include?(resource.type.to_sym)
+        label = resource.description
         resource.files.each do |file|
           label = file.title if label.blank?
           ng_document.send(
             resource.type.to_sym,
             'data-file-label': label,
             'data-slider-object': file_index,
+            'data-src': streaming_url_for(file, :dash),
```
ndushay: So :dash is hardcoded here, and ENABLED_STREAMING_TYPES array above is single element of :hls -- feels a little hinky to me.  It's not clear to me from the code how we determine which format to play.

ndushay:
My main question is how does embed know which streaming format to use?  It appears to be hardcoded for dash now; whereas it was hardcoded for mls before.   It's possible I simply don't understand how it works.

My line comment still stands, tho:  there are 3 places where the expected formats are entered:  Settings, ENABLED_STREAMING_TYPES in media_tag  and  in the new method in media_tag.  Maybe this could be knocked down to 1 or 2 places for the info.

lib/embed/media_tag.rb
jkeck: The reason there is an `ENABLED_STREAMING_TYPES` is because it wasn't immediately obvious (and it still not solidified) that the MPEG-DASH url will be used in child `<source>` element  or as the `src` attribute of the `<video>` element itself.  Having this as an array that can be configured/modified is an attempt to make this code open to new formats or changing how we handle existing formats.

I agree that it should be in fewer places though. This has mainly come out of a few different features coming together where we now have a more cohesive direction for configuration.

Would you view it as an improvement in this regard if we had a single `streaming` configuration in `Settings.yml` that included the pertinent information about the enabled types, suffixes, and mimetypes?

ndushay: It makes sense that it hasn't all settled down yet.   I was mostly aware that I personally wasn't getting why different info was in different places and how the code was choosing hls vs dash.

Putting all the info in one place could be an improvement.   Dunno if leaving it as is with comments would be viable as well.

jkeck:
@ndushay there are two possible paths here WRT how a browser will know which version to use.

Both HLS and DASH exist in the markup.  The Dash manifest is in the `data-src` attribute of the `<video>` element and the HLS playlist is in a `<source>` element.  When a browser does not support HLS then the Dash Player is initialized w/ the `data-src` attribute, and when the browser does not support Dash, then the `data-src` element is ignored and the HLS playlist provided in the `<source>` element is picked up.

I may have partially answered this in the previous comment, but thought I would clarify just in case.

ndushay:
What you said in the previous PR comment was perfect -- that was exactly the part I didn't understand.  Perhaps the PR comment can become a class or method comment for the might-as-well-be-inebriated coder.

jkeck:
I added a class level comment as well as moved the relevant configurations into settings.yml.

I did this in a separate commit, but will rebase it into 60b8d154301b2b64d24a6204c27e9a8da8fd9b48 once I get the 👍

lib/embed/media_tag.rb
```diff
@@ -1,11 +1,9 @@
 module Embed
-  # Utility class to handle generating HTML media tags
+  # Utility class to handle generating HTML <video> and <audio> tags
+  # Currently, MPEG-DASH is used at the <video> element level (in a data attribute to be picked up by javascript)
+  # and HLS is used as a <source> within the <video> or <audio> tag.
```
ndushay: here 'tis!  Thanks for this!

ndushay:
yes, I am happier now.  Thank you.  👍 on the rebase and merge.

PR #560, opened 2016-05-13
Add Webauth login link to media viewer

jkeck:
Related to sul-dlss/media#42

media-auth-link

### DIFF
--- a/app/assets/javascripts/modules/media_viewer.js
+++ b/app/assets/javascripts/modules/media_viewer.js
@@ -7,8 +7,9 @@
   var Module = (function() {
     function thumbsForSlider() {
       var thumbs = [];
-      $('.sul-embed-media video, .sul-embed-media audio').each(function(index, mediaObject) {
-        mediaObject = $(mediaObject);
+      var sliderSelector = '.sul-embed-media [data-slider-object]';
+      jQuery(sliderSelector).each(function(index, mediaDiv) {
+        var mediaObject = $(mediaDiv).find('audio, video');
         var cssClass;
         if(mediaObject.prop('tagName') === 'AUDIO') {
           cssClass = 'sul-i-file-music-1';
@@ -25,7 +26,7 @@
           '<li class="sul-embed-slider-thumb ' + activeClass + '">' +
             '<i class="' + cssClass + '"></i>' +
             '<div class="sul-embed-thumb-label">' +
-              mediaObject.data('file-label') +
+              $(mediaDiv).data('file-label') +
             '</div>' +
           '</li>'
         );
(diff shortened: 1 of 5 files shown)


ndushay:
insofar as I can tell, it looks good to me.  Dunno if you can shorten the long lines hound found;  the code looks readable and I feel kinda unqualified to determine if the tests are checking the right things, etc.

jkeck:
I would be willing to pair with somebody to try and find solutions for how to test the auth link in the media viewer.

It seems particularly complicated because we would write an integration test in ruby but would need to stub a javascript ajax call to stacks.  We could potentially allow the test to make a real call to stacks, but then the related stacks code would need to be released and deployed.

ndushay:
Maybe we can get the stacks code deployed soon to allow for the test?  Maybe make a separate ticket for writing those tests?

jkeck:
I have added a technical debt ticket about testing the auth-link (#567).

lib/embed/media_tag.rb
```diff
@@ -5,45 +5,59 @@ module Embed
...
+
+    attr_reader :purl_document, :request, :viewer
+
+    def media_element(label, file, type)
+      media_wrapper(label) do
+        <<-HTML.strip_heredoc
+          <#{type}
+            data-src="#{streaming_url_for(file, :dash)}"
+            controls='controls'
+            height="#{viewer.body_height.to_i - 24}px">
```
tingulfsen: Why do we subtract 24 here?

jkeck:
@tingulfsen the `24px` height reduction is a temporary workaround to make room for the multi-file index until we work on #565.

FWIW, that was not introduced in this PR but has been there ever since we added the multi-file index.

lib/embed/media_tag.rb
```diff
@@ -5,45 +5,59 @@ module Embed
...
-        )
+    private
+
+    attr_reader :purl_document, :request, :viewer
+
+    def media_element(label, file, type)
+      media_wrapper(label) do
+        <<-HTML.strip_heredoc
+          <#{type}
+            data-src="#{streaming_url_for(file, :dash)}"
```
tingulfsen: Why hardcode :dash here?

app/assets/javascripts/modules/media_viewer.js
```diff
@@ -47,6 +48,55 @@
...
+              mediaObject[0].load();
+              mediaObject
+                .parent('[data-slider-object]')
+                .find('[data-auth-link]')
+                .hide();
+              return;
+            }, 500);
+        });
+    }
+
```
tingulfsen: There some odd indentation going on here? See line 69-80. Could we also have some comments to explain what's happening here?

lib/embed/media_tag.rb
jkeck: It's not particularly `hardcoded`.  This is required by how the DASH player works.

HLS can be defined in a `<source>` element, however the DASH player works a bit differently since we initialize the javascript client ourselves.  Part of this implementation is to store the url to the DASH manifest in the `data-src` tag to get picked up, and served up to the DASH client code via javascript as you can see [here](https://github.com/sul-dlss/sul-embed/blob/master/app/assets/javascripts/modules/media_viewer.js#L81-L83).

So in this case, `:dash` is all we're really interested in for this attribute.

app/assets/javascripts/modules/media_viewer.js
jkeck: Good 👀.  I think it was just lined up w/ the line above, which it should not have been.  I have addressed the indentation issue.

app/assets/javascripts/modules/media_viewer.js
```diff
@@ -88,6 +137,7 @@
     return {
       init: function() {
         setupThumbSlider();
+        setupAuthLinks();
         this.initialzeDashPlayer();
```
tingulfsen: Create a future ticket to rename this to initializeDashPlayer()? 😺

tingulfsen:
The Ruby code appears to be well tested, but the JavaScript code is a little more lightly tested. Is this anything we should be concerned about? I don't know what the DLSS ambition level is for Ajax requests etc.

jmartin-sul:
@tingulfsen and i got a walk-through from @jkeck this afternoon and everything made sense.  the concerns raised seemed addressed.

agree w/ @tingulfsen on the `initialzeDashPlayer` to `initializeDashPlayer` ticket.

jmartin-sul:
created issue #569

PR #566, opened 2016-05-18
added drag and drop icon to metadata panel. Also there are some space fixes.

aeschylus:
connected to #622


aeschylus:
This is a PR for the purposes of design review. See the gifs below:

drag and drop sul-embed

aeschylus:
There are probably some more styling updates we want to make. The underline shouldn't appear underneath the icon itself, and we should perhaps adjust the whitespace and text alignment of the instructions with the icon.

aeschylus:
see also: #567

snydman:
@mejackreed @jvine @ggeisler can you please review this and give feedback? I'd like the media team to get this in if possible during this wave of work on sul-embed.

ggeisler:
* Should there be an underline on any part of this? The text is not a link, correct?

* Can the icon be a tiny bit smaller? It feels a little out of proportion to me (although it is a bit tricky to judge, having to watch the gif repeat over and over; the gif is nice for showing the interaction but a still image might make it easier to review this for design purposes).

* I wonder if we can think of better wording for the action? It's a little odd because it is under a heading called "Available Online" but it's not really an availability link like the PURL link. Also, the instruction text doesn't tell the user what it is for (why are you telling me to drag the icon?), just how to do the action.

Maybe something like:

> View image in a IIIF viewer (drag icon to another viewer)

Or maybe just:

 > View image in a IIIF viewer (drag icon)

?

jvine:
Huh, I'm wondering where the "Available Online" header came from - I hadn't noticed it before, and it doesn't seem appropriate here at all. I think we were referring to the PURL as the "Citation URL", but that doesn't provide a good label for the IIIF logo. 

I don't think I mind the size of the logo - a larger mouse target for a drag operation doesn't seem inappropriate - though it's hard to tell with the small screenshot. 

I agree that the current text doesn't seem like it should be a link. But at one point we were talking about linking to an external page that explains about IIIF and viewers. Does such a page exist? And if so, should we link to it?

ggeisler:
If there is a linkable reference page, maybe a variation of my previous suggestion:

> Drag icon to view image in a IIIF viewer (more information)

Where "more information" is the link? (and is underlined as a link)

snydman:
Alas there is not such an informational page to link to, unless we go to http://iiif.io, but that probably wouldn't be too helpful.  I guess we should create a page...

aeschylus:
screen shot 2016-05-19 at 11 03 55
screen shot 2016-05-19 at 11 08 34

aeschylus:
@ggeisler Good point about the underline. I'm not sure what I was thinking. The item is an anchor with an href in the markup, and that forces it to have an underline, but I never overrode it it seems. I agree it should be removed, and that the icon should be roughly the same size as the text, with the vertical alignment made less janky.

I added some tweaks above. Let me know if any of those seem right and I will update the PR.

jvine:
Why is it an anchor with an href? The icon itself needs to be focusable, but the text doesn't. 

I prefer the original text with the action closest to the thing it's acting on:  Drag icon to view...  @ggeisler?

And any other thoughts for the heading? Or is that out of scope of this PR?

ggeisler:
Thanks for the updates @aeschylus. I like the third version, but @jvine has done more work on this so I'd defer to her. I will say I much prefer the text in the third version rather than the original text, though. What percentage of users will even know what IIIF is? Telling them to drag the icon without explaining what the purpose of dragging it is seems odd to me.  

If we really want to start the text with the action, maybe something like:

> Drag this icon to view image in a IIIF viewer

jvine:
> Drag this icon to view image in a IIIF viewer

👍 

I agree with the "IIIF viewer" part - I just don't like the "drag icon" to be far away from what it refers to.

mejackreed:
Maybe a `cursor: move` would help?

aeschylus:
I agree we should add some cursor styling, but I would cast a vote for the open grabby hand (`cursor:grab`), which also has a "dragging" ("grabbing") style.

https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

@jvine: Should we change the title/sectioning?

I agree there should be some kind of informational link. The closest information page we have is zimeon's personal [page on the topic](http://zimeon.github.io/iiif-dragndrop/), which contains a number of outdated links. There has been [talk](http://zimeon.github.io/iiif-dragndrop/) of updating this and making it a permanent part of iiif.io.

snydman:
@aeschylus @mejackreed we should produce a new version of a page and do pull request to iiif.io. But it should be different from zimeon's page and oriented towards informing users of what it is and how they can use it.  Maybe with a little animated gif or something.  We should NOT try to solve the "explaining IIIF" to the masses problem with this, but put something out there that is reasonable and helpful.  If the IIIF eds and coco don't accept the page then we can add it to library.stanford.edu for now and change the link later.

jvine:
screen shot 2016-06-01 at 4 26 43 pm

- IIIF feature and PURL are in separate dt/dd pairs 
- dt for PURL changes to "Citation URL"
- dt for icon is "International Image Interoperability Framework"
- icon is about 25px wide w 5px right margin
- icon has "grab" cursor
- instruction text is "Drag icon to open this image/set in a IIIF viewer."
 - generic text is okay, but if possible, it would be better to check to see if there is 1 image, or multiple images, and adjust the text accordingly. 
 - Drag icon to open **this image** in a IIIF viewer.
 - Drag icon to open **these images** in a IIIF viewer.
- text colour for instruction is dimgrey 
- link on "IIIF viewer" points to http://library.stanford.edu/projects/international-image-interoperability-framework

app/assets/stylesheets/image_x.scss
```diff
@@ -147,6 +147,39 @@ $image-thumb-slider-background-color: $sul-background;
...
+    @include stanford-only;
+  }
+}
+
+.#{$namespace}-image-x-iiif-drag-and-drop-link {
+  border-bottom: none !important;
+  cursor: move; // For non-webkit browsers
+  cursor: -webkit-grab;
+  &:active {
+    cursor: -webkit-grabbing;
```
jkeck: Using `move` above to get that behavior on browsers that don't support `-webkit-grab`.

app/assets/stylesheets/image_x.scss
```diff
@@ -147,6 +147,39 @@ $image-thumb-slider-background-color: $sul-background;
...
+
+  a {
+    @include stanford-only;
+  }
+}
+
+.#{$namespace}-image-x-iiif-drag-and-drop-link {
+  border-bottom: none !important;
+  cursor: move; // For non-webkit browsers
+  cursor: -webkit-grab;
```
jkeck: Using move above to get that behavior on browsers that don't support -webkit-grab.

jkeck:
New progress for review:

## zw433vr9951 (single image)
zw433vr9951

## fw090jw3474 (multiple images)
fw090jw3474

jmartin-sul:
👍 from me, but i sort of paired on it with @jkeck yesterday, and so will refrain from being the one to merge

ggeisler:
👍

app/assets/stylesheets/image_x.scss
```diff
@@ -147,6 +147,39 @@ $image-thumb-slider-background-color: $sul-background;
...
+    position: absolute;
+  }
+
+  a {
+    @include stanford-only;
+  }
+}
+
+.#{$namespace}-image-x-iiif-drag-and-drop-link {
+  border-bottom: none !important;
```
ndushay: do we care about hound calling out `!important` ?

lib/embed/metadata_panel.rb
```diff
@@ -3,16 +3,17 @@ module Embed
...
+      @panel_content = yield block if block_given?
     end
 
     def to_html
       <<-HTML.strip_heredoc
         <div class='sul-embed-panel-container'>
           <div class='sul-embed-panel sul-embed-metadata-panel' style='display:none' aria-hidden='true'>
             <div class='sul-embed-panel-header'>
-              <button class='sul-embed-close' data-sul-embed-toggle='sul-embed-download-panel'>
+              <button class='sul-embed-close' data-sul-embed-toggle='sul-embed-metadata-panel'>
```
ndushay: was this an old typo that wasn't caught?

ndushay:
LGTM aside from a couple of line comments.  I only eyeballed it.

lib/embed/metadata_panel.rb
jkeck: Yeah, copy/paste error I believe.

app/assets/stylesheets/image_x.scss
jkeck: In this case, I think I can live with letting it slide.  Typically, yes, we should care about it as it indicates that the specificity in the css is too complicated, but the amount of change that would require at this point would be to high to consider it viable.

snydman:
excited to see this. I do notice that the metadata panel does not appear in PURL versions of the embed.  I believe the rationale for this was that it would be duplicative of metadata in the PURL page itself.  I suppose this means we ought to add the IIIF logo to the PURL page for images, unless @LynnMcRae and others have a different idea.

jkeck:
This has actually come up elsewhere as well.  I believe PURL is now the only app that hides the metadata panel.  Do we want to retain this behavior for PURL?  (FWIW, we have a version of purl on sul-purl-dev that shows the metadata panel w/ IIIF icon).

jvine:
@ggeisler do you have any opinion about no longer suppressing the info icon, regardless of context? so that the IIIF drag-and-drop is always available. otherwise it needs to be added to specific apps. i'm okay with not suppressing it.

LynnMcRae:
@jkeck seems like SW and Purls are equals here...both show the metadata independently and no harm in SW that it shows some data also in the info panel.  I'm fine with that behavior in Purl. It just emphasizes the portability of embed.

ggeisler:
This sounds fine to me, too. I don't see a strong argument for suppressing it.

PR #573, opened 2016-06-02
Only apply index slider padding to the media player when there are …

### LINKED ISSUE #565: Ensure margins around video are set correctly if there a single video
Single video should have no bottom margin - i.e. no space left for an index if there is no index.

Single:
single-video

Multi with (closed) index:
multi-video-index-closed

jkeck:
…multiple media files.

Closes #565 

## Single Video
single-video

## Multiple Videos
multiple-videos

### DIFF
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -4,6 +4,7 @@ module Embed
   # and HLS is used as a <source> within the <video> or <audio> tag.
   class MediaTag
     SUPPORTED_MEDIA_TYPES = [:audio, :video].freeze
+    MEDIA_INDEX_CONTROL_HEIGHT = 24
 
     def initialize(viewer)
       @viewer = viewer
@@ -35,7 +36,8 @@ module Embed
             data-src="#{streaming_url_for(file, :dash)}"
             data-auth-url="#{authentication_url(file)}"
             controls='controls'
-            height="#{viewer.body_height.to_i - 24}px">
+            class="#{'sul-embed-many-media' if many_media?}"
+            height="#{media_element_height}">
             #{enabled_streaming_sources(file)}
           </#{type}>
         HTML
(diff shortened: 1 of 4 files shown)


spec/lib/embed/media_tag_spec.rb
```diff
@@ -39,8 +39,17 @@
...
+    context 'single video' do
+      let(:purl) { single_video_purl }
+      it 'includes a height attribute equal to the body height' do
+        expect(subject).to have_css("video[height='#{viewer.body_height}px']")
+      end
+    end
+
+    context 'multiple videos' do
+      it 'includes a height attribute equal to the body height minus some px to make way for the thumb slider' do
+        expect(subject).to have_css('video[height="276px"]')
```
ndushay: maybe a comment indicating the actual video height is 300px  or whatever?

lib/embed/media_tag.rb
```diff
@@ -4,6 +4,7 @@ module Embed
   # and HLS is used as a <source> within the <video> or <audio> tag.
   class MediaTag
     SUPPORTED_MEDIA_TYPES = [:audio, :video].freeze
+    MEDIA_INDEX_CONTROL_HEIGHT = 24
```
ndushay: naive question:   is it a coincidence that the scss thing is 24px and this constant is 24?  I'm assuming  "no" but there is no good way to share a constant between ruby code and scss.

ndushay:
two line comments;   o.w. 👍 to merge

spec/lib/embed/media_tag_spec.rb
jkeck: That's kind of indicated in the test above.  `viewer.body_height` is set above and there is a test indicating that in the context of a single video purl, the height of the video should equal the viewer body height.

In this case, I prefer to not include implementation details in the comments since they are less likely to change than the test itself.

lib/embed/media_tag.rb
jkeck: The only things that I can think of that would allow us to pull this value into the place that it would be needed seems worse that this (add the value to the markup, then pick up the value with javascript to add to the media index markup as it is being built).

ndushay: ja, i figured it was not a good idea.  Just wanted to be sure.

spec/lib/embed/media_tag_spec.rb
ndushay: I understood the description, but I didn't see how to understand that 276 was the right number, esp. with data elsewhere.

I was thinking something as simple as this:
```
it 'includes a height attribute equal to the body height (300) minus some px to make way for the thumb slider
```
but not worth another PR.

PR #579, opened 2016-06-07
Adding number of downloadable files in footer.

### LINKED ISSUE #561: Show number of downloadable files next to download icon
Only images and media files are listed in the index (filmstrip) below the player. Extra files (pdf, zip, etc.) are listed only in the download slider. 

Add a badge next to the download icon showing the total number of downloadable items (i.e. downloadable media and images, plus extra files).

download-count

List all media files and all supplemental files in the download panel (and count) regardless of which file is currently selected in the player.

Deprecating this next bit for now, not sure we want to do it.
> Possible future requirement:  Follow the image model, and list in the download panel (and count):
> 1. the currently-selected media file, 
> 2. any dependent files (e.g. a linked transcript),
> 3. any object-level supplemental files (e.g. images of the container, zip file...)
> ... so the list and (possibly) count will change with each selected video.

tingulfsen:
Closes #561 and supersedes #572 

### Before (https://purl-test.stanford.edu/cc340qm0511):
579 before

### After:
579 after


tingulfsen:
I think that the Travis CI build fails because there are a few too many lines of code.

This spurred me to refactor to use the strip_heredoc method instead of Nokogiri. See #580

ndushay:
@tingulfsen I am so sad that this is missing before and after pictures.  how can I ever trust you again?

jmartin-sul:
👍 as far as i can tell, though i am not the most qualified person to evaluate this.  and i can't quite understand the usage of `sul-i-download-3`:  it appears to already be in use as a css class name, but also appears to be added by this PR.  is this just my front-end naivete showing?

spec/lib/embed/viewer/common_viewer_spec.rb
```diff
@@ -72,10 +72,10 @@
   end
   describe 'footer_html' do
     it 'should return the objects footer' do
-      stub_request(request)
+      stub_purl_response_and_request(file_purl, request)
       html = Capybara.string(file_viewer.footer_html)
       expect(html).to have_css 'div.sul-embed-footer'
-      expect(html).to have_css '[aria-label]', count: 2
+      expect(html).to have_css '[aria-label]', count: 3
```
ndushay: a better test would look for the values in the aria-label;  this is pretty generic.

Also, it isn't testing for the VALUE displaying, only the label.

ndushay:
I'm not happy with the testing yet.  So sadly, 👎

ndushay:
I improved the tests, so someone should look again and then merge if they're ok with this.

PR #594, opened 2016-06-10
Add Stanford S icon to Stanford only files in multiple-media filmstrip

### LINKED ISSUE #588: add S icon to media  thumb slider for stanford restricted content
from #577:

- [ ] if there are multiple files in the index, any stanford-restricted files have a red "s" after the label. 

* The logic to determine if a file is stanford-restricted is already in sul-embed (see download panel file);  we should add a data-attribute (like there is for file title) in the filmstrip.

* Note the login label position is relative to the index, so it's not hidden when the index is open
 
login-to-play-index

jkeck:
Closes #588 

stanford-only-media-filmstrip

### DIFF
--- a/app/assets/javascripts/modules/media_viewer.js
+++ b/app/assets/javascripts/modules/media_viewer.js
@@ -23,10 +23,16 @@
         }
 
         var thumbClass = 'sul-embed-slider-thumb sul-embed-media-slider-thumb ';
+        var labelClass = 'sul-embed-thumb-label';
+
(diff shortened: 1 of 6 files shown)


jkeck:
@jvine we wondered if putting the Stanford S before the text would be a reasonable alternative.

It allows us to avoid the text overflow issues and seems to play pretty nicely.

Are you okay with that?

jvine:
👍  fine w me.

Issue #589, opened 2016-06-09
add overlay message to media viewer when rights are location OR stanford


ndushay:
(do this before #587)

from #577 

- [ ] show the player, with keyframe if there is one (Note:  we have no keyframe support in contentMetadata at this time)
- [ ] overlay message to non-stanford users
- [x] show login link
- [x] 1 item available for download; panel will have a red s next to the filename

![location-or-login](https://cloud.githubusercontent.com/assets/6945691/15796077/e950993c-29ac-11e6-90e4-a1522c3d5dd2.png)

jkeck:
@jvine should the `Limited access for non-Stanford guests.` be a heading?

jvine:
Not a heading. I'm pretty sure it doesn't need to be anything other than text content.

ndushay:
@jvine:   we clumped and got to here.  We need more info on colors and styling desired.  

![wip - 587](https://cloud.githubusercontent.com/assets/96775/15980976/a642336c-2f24-11e6-85f1-9a6f94c61538.png)

jvine:
Here's what I did:
```
.sul-embed-media-location-restricted {
    background-color: white;
    border-radius: 10px;
    padding: 20px 25px;
    opacity: .9;
    width: 80%;
}
}
.sul-embed-media-location-restricted.line1 {
    display: block;
    font-size: 35px;
    font-weight: 100;
}
.sul-embed-media-location-restricted.line2 {
    display: block;
    font-size: 24px;
    font-weight: 100;
}
```

![screen shot 2016-06-10 at 5 16 47 pm](https://cloud.githubusercontent.com/assets/6945691/15981769/30e30208-2f2f-11e6-8c9f-1e7d776266eb.png)

jmartin-sul:
WIP branch:  https://github.com/sul-dlss/sul-embed/compare/589-overlay-message

added a couple small commits (one takes a shot at applying @jvine's css suggestions, the other is a minor rubocop fix).  did no rebasing (to make it easy to revert in case i did the wrong thing).

Issue #587, opened 2016-06-08
add (Restricted) text to media for location only restricted content


ndushay:
(do this after #589)

When media content is location restricted (but NOT also stanford restricted) 

- [ ] need to add red "(Restricted)" text to thumb slider for file(s)
- [ ] need to show "Restricted media cannot be played in your location." in media player area.
- [ ] ensure location restricted files do not show up in the download panel 
  - [ ] adjust download count to 0

Somewhat dependent on sul-dlss/digital_stacks_rails#54

original from #577:

### 3) Must come to reading room, no matter who you are.
- show icon, no keyframe
- overlay message to all users
- no login link
- download panel has a count of 0, assuming there are no supplemental files

![nothing-to-see-here](https://cloud.githubusercontent.com/assets/6945691/15796082/01a33af8-29ad-11e6-8259-1a7aaed53f98.png)

- if there are multiple files in the index, any location-restricted files have the word "(Restricted)" after the label
- download count ... I'm guessing if the item is location-restricted, it's not downloadable at all? so the count matches the number of actually downloadable files (including stanford-authed downloads)

![nothing-to-see-index](https://cloud.githubusercontent.com/assets/6945691/15796286/3ef96726-29b0-11e6-8afb-88ca39a497b8.png)

PR #598, opened 2016-06-27
access restriction messaging

jmartin-sul:
resolves issue #587
resolves issue #589

this branch should implement the messaging overlays for access restrictions (location based and stanford user based).

this depends on sul-dlss/digital_stacks_rails#61.

clumped with @ndushay, @jkeck, and @jvine to implement this.

could we get @tingulfsen to do code review?  and @jvine to review styling?

code has been deployed to embed-dev and stacks-test.
screen shot 2016-06-27 at 4 18 31 pm
screen shot 2016-06-27 at 4 24 38 pm

connected to #587
connected to #589

### DIFF
--- a/app/assets/stylesheets/media.scss
+++ b/app/assets/stylesheets/media.scss
@@ -16,15 +16,44 @@
 
 .#{$namespace}-media-wrapper {
   display: inline-block;
+  position: relative;
+  text-align: center;
 }
 
-.#{$namespace}-media-location-restricted {
-  @include rounded($border-radius-base);
+.#{$namespace}-media-access-restricted-container {
+  left: 0;
   position: absolute;
-  top: 50px;
-  font-size: 25px;
-  padding: 4px;
-  background-color: $gray-54-percent;
+  top: 15%;
+  width: 100%;
+}
+
+.#{$namespace}-media-access-restricted {
+  @include rounded($border-radius-base);
+  background-color: $white-color;
+  margin: 0 auto;
+  opacity: 0.9;
+  padding: 20px 25px;
+  width: 80%;
+
+  .line1 {
+    display: block;
+    font-size: 35px;
+    font-weight: 100;
+  }
+
+  .line2 {
+    display: block;
+    font-size: 24px;
+    font-weight: 100;
+  }
+}
+
+.#{$namespace}-media-location-only-restricted-overlay {
+  color: $color-beige-30;
+  font-size: 20em;
+  position: absolute;
+  top: -10%;
+  width: 100%;
 }
 
 .#{$namespace}-media-slider-thumb {
--- a/app/assets/stylesheets/media.scss
+++ b/app/assets/stylesheets/media.scss
@@ -14,6 +14,19 @@
   }
 }
 
+.#{$namespace}-media-wrapper {
+  display: inline-block;
+}
+
+.#{$namespace}-media-location-restricted {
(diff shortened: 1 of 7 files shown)


jmartin-sul:
i don't know my CSS best-practices well enough to know whether @houndci-bot is complaining about something we should change.  IIRC, @jvine had said to keep the properties in the order she had them, though i'm not 100% sure about that.

jvine:
@jmartin-sul Nope, as long as the properties are attached to the right elements, it doesn't matter (to the display) how they're ordered.

app/assets/stylesheets/media.scss
```diff
@@ -14,6 +14,48 @@
...
+.#{$namespace}-media-access-restricted-container {
+  left: 0;
+  position: absolute;
+  top: 15%;
+  width: 100%;
+}
+
+.#{$namespace}-media-access-restricted {
+  @include rounded($border-radius-base);
+  background-color: #ffffff;
```
jmartin-sul: i couldn't find anything in sul_styles for white.  should i declare it somewhere in here, or ignore hound's complaint?

jmartin-sul:
addressed all of the hound complaints that were easy to address, left a question about the remaining one

jvine:
"White" is a browser given - so you can do "color: white" without declaring it anywhere. Maybe hound will like that better?

jmartin-sul:
well, hound's prior complaint was about that specifically:
> Color `white` should be written in hexadecimal form as `#ffffff`<br>Color literals like `white` should only be used in variable declarations; they should be referred to via variable everywhere else.

so i guess it just doesn't want inline non-variable color references?

app/assets/stylesheets/media.scss
jkeck: @jmartin-sul use `$white-color` ([code](https://github.com/sul-dlss/sul-embed/blob/master/app/assets/stylesheets/sul_variables.scss.erb#L10))

tingulfsen:
I was just testing this on my localhost. After logging in per the red message, that overlay message is still there. Is this the intended behavior? Tested on Safari.

affiliatedafterlogin

jmartin-sul:
no, i don't think that's intended.  but maybe expected?  did the login link disappear before?  @jkeck might have more insight since he's more familiar w/ embed, but it looks to me like the code that removes the access restriction message and login label only gets run on page load.

i tested just now, and on both purl-test and my localhost, i saw the same behavior, but the login link and message both disappeared when i reloaded the page.

jkeck:
Certainly not intended behavior.  We should fix that in this PR, because I believe it must have been a regression caused by this change.

tingulfsen:
I wonder if this is related to another issue I've seen in the past: Sometimes when you log in to watch a video, you have to reload the page in order for the video to appear or start playing. @jkeck @jvine are we ok with this behavior?

jmartin-sul:
i would agree that we should try to fix it, and in this PR if possible... it's a pretty rough edge.  can we run auth check whenever the window gets focus or something like that?

jkeck:
Hi all,
I believe I have fixed both the placement of the login link as well as ensuring it (and the restriction message) is removed on login.  @tingulfsen you should be able to give it another look.


### gm746xg1831 (before)
gm746xg1831-before

### gm746xg1831 (after)
gm746xg1831-after

jmartin-sul:
worked for me when i tested the latest code on my laptop

PR #599, opened 2016-07-05
Allow image thumbnails to be viewed in the media viewer

### LINKED ISSUE #576: Add non-media content to the multi-file thumb image slider
When an object has a combination of media and "other" (images of the media, PDF transcript, etc.) the goal is to include in the index all the items that can be viewed in place; items without a viewer are available only in the download list. 

Object has media + image of media + PDF - index looks like this:
3-files-index

download list looks like this:
3-files-download

Clicking the image in the index displays the thumbnail in the viewer (no OSD zoom/pan behaviour)
showing-thumbnail

which is equivalent functionality to clicking "Preview" on the image in the current file list
screen shot 2016-06-03 at 4 15 06 pm

jkeck:
Closes #576 

### bd786fy6312

bd786fy6312-first

bd786fy6312-second


lib/embed/stacks_image.rb
```diff
@@ -0,0 +1,24 @@
+module Embed
+  ###
+  # A module to be mixed into classes to get generate
+  # URLs to stacks images using the IIIF image API
```
ndushay: typo in comment "to get generate URLS"

ndushay:
LGTM but I didn't test it locally or anything.

lib/embed/stacks_image.rb
jkeck: 👍 Good 👀 

I'll fix this as soon as I'm at a good stashing point in my current working branch.

PR #601, opened 2016-07-06
Increase the bottom position of WebAuth link for media viewer.

jkeck:
## xv428cw8400 (before)
xv428cw8400-before

## xv428cw8400 (after)
xv428cw8400-after

### DIFF
--- a/app/assets/stylesheets/modules/auth_link.scss
+++ b/app/assets/stylesheets/modules/auth_link.scss
@@ -4,7 +4,7 @@
 .#{$namespace}-auth-link {
   @include stanford-only-red;
   background-color: $sul-auth-link-background-color;
-  bottom: 15px;
+  bottom: $sul-media-index-height;
   padding: 5px;
   position: absolute;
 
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/modules/auth_link.scss
```diff
@@ -4,7 +4,7 @@
 .#{$namespace}-auth-link {
   @include stanford-only-red;
   background-color: $sul-auth-link-background-color;
-  bottom: 15px;
+  bottom: 24px;
```
jkeck: I could be convinced to change this to a variable (since I think it may exist elsewhere in the css).

ndushay: i would be in favor of a variable.

ndushay:
make it a variable, use variable in all approp places -- then it will LGTM

app/assets/stylesheets/modules/auth_link.scss
jkeck: 👌

jmartin-sul:
tested in chrome and safari from my local instance, seems to have taken @ndushay's variable suggestion, 👍 from me

Issue #604, opened 2016-07-08
Remove drop-shadow from embed and add border


jkeck:
This should be completed after #601 is merged.

Due to the fact that we're moving to an iframe only solution the drop shadow in not visible.  This is caused because the embed takes up 100% of the frame and the drop-shadow is not part of the calculated width of the element.  If we add a margin around the element to make the drop shadow visible it will make the alignment of the embed not match up with where it did (effectively pushing the embed left by 10px for every existing consumer).

@jvine suggested that we remove the drop-shadow altogether and instead add an explicit border which would match the top border of the panels (EmbedThis, Download, etc).

tingulfsen:
This picture illustrates the issue:

![shadowbox](https://cloud.githubusercontent.com/assets/3740294/16887382/93b1550e-4a8e-11e6-9702-16594800a8a7.png)

The drop-shadow in question is clearly visible at the very bottom of the image, below the three bottom icons. If we just remove it, we'll also lose the left and right borders that delimit the embed. Therefore we need to add left and right margins, probably just like the horizontal margin just above the "Download item" text.

jvine:
Unless I'm missing a later development, this is answered above:
> remove the drop-shadow altogether and instead add an explicit border which would match the top border of the panels (EmbedThis, Download, etc).

On .sul-embed-container, remove the box-shadow, and replace it with border: 1px solid #5f574f. 

Will the iframe then clip the 1px border? I dunno. But we don't want to add margin within the iframe, or the embed will be misaligned with other things on the page.

PR #609, opened 2016-07-15
Return an iframe as the oEmbed API's rich HTML payload

### LINKED ISSUE #600: Change HTML payload of oEmbed service to be the iframe code itself
This addresses a few issues for sites that embed the content directly via the oEmbed service:

1. CSS clashes between embedding site (we have noticed this w/ the `line-height` in PURL).
1. Javascript clashes when embedding multiple objects in the same page.
1. Having a predictable `Origin` for the `Access-Control-Allow-Origin` header for media content (see https://github.com/sul-dlss/sul-embed/issues/581#issuecomment-226237112 and https://github.com/sul-dlss/digital_stacks_rails/pull/63#issuecomment-230074196).

jkeck:
Closes #600

Note: The drop-shadow and border style issue that you can see in the after image below will be addressed by #604

## SearchWorks: zb732fz3904 (before)
zb732fz3904-before

## SearchWorks: zb732fz3904 (after)
zb732fz3904-after


lib/embed/viewer/common_viewer.rb
```diff
@@ -228,7 +228,12 @@ def height_style
       end
 
       def width_style
-        "max-width:#{width}px;" if width
+        "max-width:#{with_style_attribute};"
+      end
+
+      def with_style_attribute
+        return '100%' unless width
+        "#{width}px"
```
jmartin-sul: should this be `width_style_attribute`?  seems like a plausible intentional name, but also seems like it could be a typo?

jmartin-sul:
other than the one question left inline, LGTM.  i redeployed this branch to embed-dev earlier this afternoon (embed-dev had this branch, but a different commit, so i assumed it'd been rebased and maybe edited since its last deployment).  did a bit of cursory testing on purl-test, and things looked fine there.

lib/embed/viewer/common_viewer.rb
jkeck: I'll fix that.

spec/rails_helper.rb
```diff
@@ -58,9 +58,19 @@ def fill_in_default_sandbox_form(druid = 'ab123cd4567')
...
 
+def visit_iframe_response(druid = 'ignored', min_files_to_search: nil, hide_search: nil, hide_title: nil, hide_download: nil)
+  visit iframe_path(
+    url: "#{Settings.purl_url}/#{druid}",
+    min_files_to_search: min_files_to_search,
+    hide_search: hide_search,
+    hide_title: hide_title,
+    hide_download: hide_download
+  )
+  expect(page).to have_css('.sul-embed-container')
```
ndushay: I'm still not fond of these helper methods hiding expectations but it's not a show-stopper.

ndushay:
LGTM AFAICT.   Probably need a deployment plan to hook this up with purl, searchworks, exhibits and ... ?

jkeck:
Actually, if things go according to the plan it should be completely transparent to all consumers (which is the promise of oEmbed, if done right!).

ndushay:
Purl and SearchWorks and Exhibits all are not currently putting it in an iframe?  Or a nested iframe doesn't matter?  Or am I being dense?

jkeck:
PURL and SearchWorks blindly inject the markup provided in the `html` value in the oEmbed API response.  That is currently the markup of the viewer itself, now this PR returns markup for the iframe.  So, once this PR is in place, SearchWorks and PURL will simply begin to inject the iframe into the page instead of the raw viewer markup.

PR #610, opened 2016-07-15
Replace drop shadow around .sul-embed-container with a simple border.

### LINKED ISSUE #604: Remove drop-shadow from embed and add border

tingulfsen:
Fixes #604 

Before:

before

After:

after

I've just verified that this seems to work fine with #609

### DIFF
--- a/app/assets/stylesheets/common.scss.erb
+++ b/app/assets/stylesheets/common.scss.erb
@@ -13,7 +13,7 @@
   font-family: 'Source Sans Pro', 'Arial Unicode MS', Helvetica, sans-serif;
   font-size: $font-size-base;
   overflow-y: hidden;
-  @include shadow(3);
+  border: 1px solid #5f574F;
 
   * {
     padding: 0;
--- a/app/assets/stylesheets/common.scss.erb
+++ b/app/assets/stylesheets/common.scss.erb
@@ -13,7 +13,7 @@
   font-family: 'Source Sans Pro', 'Arial Unicode MS', Helvetica, sans-serif;
   font-size: $font-size-base;
   overflow-y: hidden;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/common.scss.erb
```diff
@@ -13,7 +13,7 @@
   font-family: 'Source Sans Pro', 'Arial Unicode MS', Helvetica, sans-serif;
   font-size: $font-size-base;
   overflow-y: hidden;
-  @include shadow(3);
+  border: 1px solid #5f574F;
```
jkeck: We shouldn't reference colors directly in our css declarations.  This is a color that is defined in SUL Styles as `$color-pantone-405` (which we have defined as `$sul-border-color` in SUL Embed).  Can we use that variable here instead if the color itself?

jkeck: FWIW, I do see that this was used below (although I don't think in a border context).  We should probably fix that too, but don't feel obligated to include that in this PR.

jmartin-sul:
worked for me on my laptop.

since tommy's on vacation and jessie's suggestion seemed both reasonable and straight-forward, i pushed a commit fixing both of the explicit color references.  still works on my laptop.

PR #614, opened 2016-07-25
Add a max-height to images in the media-viewer that ensures they are …

### LINKED ISSUE #607: Make sure images with 400px long-edge on the y-axis are not clipped in media viewer
Currently the `body-height` in embed does not give enough room for images that are 400px tall.

We need to allow the embed to be large enough to show the full image or in the case where a maxheight option is set, resize the height of the image (if necessary).

jkeck:
…not larger than the body-height of the viewer itself.

Closes #607 

## xv428cw8400 (before)
xv428cw8400-before

## xv428cw8400 (after)
xv428cw8400-after

### DIFF
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -52,7 +52,11 @@ module Embed
 
     def previewable_element(label, file)
       media_wrapper(label: label, thumbnail: stacks_square_url(@purl_document.druid, file.title, size: '75')) do
-        "<img src='#{stacks_thumb_url(@purl_document.druid, file.title)}' class='sul-embed-media-thumb' />"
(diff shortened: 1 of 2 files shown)


ndushay:
LGTM.  I can't imagine why it's bad for images to get cropped at an "arbitrary" place based on viewer height - what could possibly be bad about that?  ;-)   I mean, do we really need to see that your t-shirt says "CHUG" ?

jmartin-sul:
LGTM too, tested on laptop, merged

jkeck:
@ndushay It's an onomatopoeia for the sound a train makes.  Hence the train-tracks beneath the word. ;)

PR #619, opened 2016-07-27
Correct fixture names (merge before 618!)

ndushay:
and clean up a duplicated fixture/test.  (yes, I looked through git history trying to see if it was ever anything else ... and ... nope.)

### DIFF
--- a/spec/fixtures/purl_fixtures.rb
+++ b/spec/fixtures/purl_fixtures.rb
@@ -108,11 +108,11 @@ module PURLFixtures
       </publicObject>
     XML
   end
-  def file_purl_no_size
+  def image_no_size_purl
     <<-XML
       <publicObject>
         <identityMetadata>
-          <objectLabel>Book as Images</objectLabel>
+          <objectLabel>image without size</objectLabel>
         </identityMetadata>
         <contentMetadata type="image">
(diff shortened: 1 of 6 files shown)


spec/fixtures/purl_fixtures.rb
```diff
@@ -261,21 +224,13 @@ def multi_media_purl
...
-            </machine>
-          </access>
+          #{access_discover_world}
+          #{access_read_world}
         </rightsMetadata>
       </publicObject>
     XML
   end
-  def multi_resource_multi_media_purl
+  def multi_contentMd_multi_media_purl
```
jkeck: Hmm, I wonder if we actually intended to have multiple `<contentMetadata>` elements here.  @LynnMcRae that's not legitimate is it?

ndushay: it may not be legit, but I'll wager we have data like this.  :-P

ndushay: @LynnMcRae says there can only be one contentMetadata block because there can only be one contentMd datastream.  I will change the fixtures accordingly.

ndushay:
updated to get rid of duplicate contentMetadata sections - thanks @jkeck !

PR #621, opened 2016-07-27
timeouts for getting purls and was thumbs now in Settings file

ndushay:
Seems more appropriate than having them hardcoded ...

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,4 +1,7 @@
 purl_url: 'https://purl.stanford.edu'
+# timeouts specified in seconds
+purl_read_timeout: 2
+purl_conn_timeout: 2
 valid_purl_url: <%= /https?:\/\/purl\.stanford\.edu\/*/ %>
 stacks_url: 'https://stacks.stanford.edu'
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -10,6 +13,9 @@ squash_disable: <%= (Rails.env.development? || Rails.env.test?) %>
 geo_external_url: 'https://earthworks.stanford.edu/catalog/stanford-'
 geo_wms_url: 'https://geowebservices.stanford.edu/geoserver/wms/'
 was_thumbs_url: 'https://thumbnail-service-example'
+# timeouts specified in seconds
+was_thumb_read_timeout:
+was_thumb_conn_timeout:
```
jmartin-sul: accidentally omitted values for these two?  or yml magic i'm not understanding?

jmartin-sul:
left a note inline, but 👍 on the general idea (or 👍 overall if my comment is off-base)

config/settings.yml
ndushay: whoops!   thanks for catching this.

PR #624, opened 2016-07-29
add more urls to is_it_working

ndushay:


jmartin-sul:
tests repeatedly fail with different errors for the CI build that hasn't passed above.  but tests passed in the other build, and pass on my laptop, and the change seems unlikely to cause the failures seen here, so chalking it up to poltergeist flakiness and merging...

PR #628, opened 2016-08-03
[WIP] set video element height and width from PURL XML videoData info (if possible)

### LINKED ISSUE #626: Add height/width attributes to the video element
This information is available (in our test objects) in attributes on the `<videoData>` element.

jmartin-sul:
if there's no videoData in the PURL XML, the height defaults to what it would've been before this PR (viewer body height, minus thumb slider height if there's a slider), and the width attribute is omitted.

before (note lack of width, old default height):
screen shot 2016-08-03 at 3 01 52 pm

after (note presence of width from object's PURL XML videoData; and height, which is the videoData height of 288, minus the 24 pixel height of the thumb slider):
screen shot 2016-08-03 at 3 00 32 pm

closes #626 
connect to #626

### DIFF
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -86,13 +86,22 @@ module Embed
       end.join
     end
 
+    def usable_body_height
+      many_primary_files? ? viewer.body_height.to_i - MEDIA_INDEX_CONTROL_HEIGHT : viewer.body_height.to_i
+    end
+
     def media_element_height(file)
-      elt_base_height = (file.video_height || viewer.body_height).to_i
-      many_primary_files? ? elt_base_height - MEDIA_INDEX_CONTROL_HEIGHT : elt_base_height
+      return usable_body_height unless file.video_height && (file.video_height.to_i < usable_body_height)
+      file.video_height.to_i
     end
 
     def media_element_width(file)
-      file.video_width.to_i if file.video_width
+      return unless file.video_width
+
+      # if there's a width specified, scale it by the same factor as the actual
+      # display height vs the specified height, so correct aspect ratio is maintained
+      height_scale_factor = media_element_height(file) / file.video_height.to_d
+      (file.video_width.to_i * height_scale_factor).to_i
     end
 
     def media_element_height_attr(file)
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -44,7 +44,8 @@ module Embed
             data-auth-url="#{authentication_url(file)}"
             controls='controls'
             class="#{'sul-embed-many-media' if many_primary_files?}"
-            height="#{media_element_height}">
+            #{media_element_height_attr(file)}
+            #{media_element_width_attr(file)}>
             #{enabled_streaming_sources(file)}
           </#{type}>
         HTML
@@ -56,7 +57,7 @@ module Embed
         "<img
           src='#{stacks_thumb_url(@purl_document.druid, file.title)}'
           class='sul-embed-media-thumb #{'sul-embed-many-media' if many_primary_files?}'
-          style='max-height: #{media_element_height}'
+          style='max-height: #{media_element_height(file)}'
         />"
       end
     end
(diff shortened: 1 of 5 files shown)


spec/models/embed/purl_spec.rb
```diff
@@ -230,6 +230,14 @@
           expect(file.location).to eq 'http://stacks.stanford.edu/file/druid:abc123/Title_of_the_PDF.pdf'
         end
       end
+      describe 'video_data' do
+        before { stub_purl_response_with_fixture(multi_media_purl) }
+        let(:video) { Embed::PURL.new('JessieSaysNo.mp4').contents.first.files.first }
```
jkeck: Can we make the argument passed into the initializer either `ignored` or an arbitrary druid?  `Embed::Purl#initialize` takes a string argument that is a druid.

spec/lib/embed/media_tag_spec.rb
```diff
@@ -147,8 +156,12 @@
     end
 
     describe '#previewable_element' do
-      before { stub_purl_response_with_fixture(purl) }
+      before do
+        stub_purl_response_with_fixture(purl)
+        allow(file).to receive(:video_height).with(no_args)
```
jkeck: Hmm, is `.with(no_args)` required if we're not trying to assert that the method is invoked explicitly w/o args?

lib/embed/media_tag.rb
```diff
@@ -84,9 +85,21 @@ def enabled_streaming_sources(file)
       end.join
     end
 
-    def media_element_height
-      return "#{viewer.body_height}px" unless many_primary_files?
-      "#{viewer.body_height.to_i - MEDIA_INDEX_CONTROL_HEIGHT}px"
+    def media_element_height(file)
+      elt_base_height = (file.video_height ? file.video_height : viewer.body_height).to_i
```
jkeck: Could this be more sussinctly written as `elt_base_height = (file.video_height || viewer.body_height).to_i`?

jmartin-sul: indeed

spec/lib/embed/media_tag_spec.rb
jmartin-sul: fair point, that would make more sense on an `expect` than this `allow`.

spec/models/embed/purl_spec.rb
jmartin-sul: d'oh, yeah, was thinking it took a file name for some reason when i was writing this.  will do the same as the other examples in this file.

jmartin-sul:
ok, made @jkeck's suggested changes, should be ready for review again once tests finish...

jkeck:
I'm noticing on larger videos that the bottom controls end up being obscured below the bottom of the embed panel.

The `max-height` of the `.sul-embed-container` is `400px` so we need to make sure we don't set the media height to something larger than the max height of the container.  More than happy to take a look at this with you this afternoon.

### gm746xg1831
gm746xg1831

jmartin-sul:
would it be reasonable to have the height max out at the sul-embed container max height, and to preserve the aspect ratio (i.e. scale the width accordingly if need be)?

jmartin-sul:
discussed with @jkeck on friday and he agreed scaling was reasonable.  i implemented that, but... it looks a bit odd to me for narrow videos, e.g.:
screen shot 2016-08-07 at 6 26 44 pm

for example, it seems like it'd be harder to seek on narrow videos, which seems like an undesirable side-effect.

video not playing there is unrelated to this, i was testing in firefox and this branch doesn't yet incorporate jessie's work to get video to play in non-safari browsers.  though, looking at that PR, even if we decide to keep this work, we should probably wait, merge the other one, and then rebase this one on updated master, since both branches touch some of the same parts of `media_tag.rb`.

jmartin-sul:
closing as per the [note](https://github.com/sul-dlss/sul-embed/issues/626#issuecomment-238999600) on issue #626 (TL;DR: turns out video JS already does what we wanted by default).

PR #633, opened 2016-08-05
Add video.js to play HLS streams in non-safari browsers

jkeck:
This PR also migrates the logic of how to construct URLs to the streaming server from stacks to the SUL Embed code.

_FWIW, this code is currently deployed to the embed service being served up via purl-test_


app/assets/javascripts/modules/media_viewer.js
```diff
@@ -180,9 +220,11 @@
       init: function() {
         setupThumbSlider();
         authCheck();
-        this.initializeDashPlayer();
       },
 
+      // This is currently unused under video-js.
+      // We should remove it (as well as all related code)
+      // if we go w/ video-js as our media client
```
jmartin-sul: if it is totally unused, i think i'd be in favor of removing it now, unless you think we'd need it back soon anyway (esp if it'd be a pain to add back).

app/assets/javascripts/modules/media_viewer.js
```diff
@@ -98,42 +157,23 @@
...
-            // As best I can tell, .touch() does nothing (and does not exist in jQuery),
-            // but without calling some function after .load() the media fails to load.
-            mediaObject[0].load().touch();
+            authCheckForMediaObject(mediaObject, function(_, data) {
+              if(data.status === 'success') {
+                clearInterval(checkWindow);
+                mediaObject
+                  .parents(sliderObjectSelector)
+                  .find('[data-auth-link], ' + restrictedOverlaySelector + ', ' + restrictedMessageSelector)
+                  .hide();
```
jmartin-sul: sorry for the JS/JQuery naivete, but is what's being found and hidden here in this callback different than the elements that are being found and hidden in lines 128-134?  or is it happening at a different point in the process (e.g. before vs after login)?  couldn't tell whether this was duplicating behavior or if i was just misunderstanding things.

jmartin-sul:
left a couple comments inline, but otherwise LGTM

app/assets/javascripts/modules/media_viewer.js
jkeck: The things being hidden in 128-134 are overlays and divs with message text.  This particular code is hiding the Login link itself.

Now that I look at this though, we should have a direct reference to the link (since this is in an on-click event) and shouldn't have to re-find it.  I'll see how easy that is to change.

jkeck: Ah, actually, now that I read that selector correctly, I see that it is doing both (I thought it was constructing a single selector, nut multiple).

This was required before since the act of clicking the link wouldn't trigger another call to the check the authentication status of the user (which now it does).  I'll see if I can change this so that we don't attempt to do that twice (although it is harmless).

ndushay:
LGTM per eyeball review.

PR #635, opened 2016-08-10
Adding css to sul-embed-container to help mobile responsive behavior.

### LINKED ISSUE #634: Look into mobile UI issues for media viewer
- [ ] Login link not being hidden after authenticating
- [ ] 100% width iframe potentially causing issue (may need CSS solution)
 * _This may be a symptom of moving to an all-iframe solution._

jkeck:
Closes #634 (fwiw, it looks like the login link issue was resolved when we stopped redirecting through stacks to wowza).

See http://stackoverflow.com/questions/23083462/how-to-get-an-iframe-to-be-responsive-in-ios-safari for explanation.

_Note: The video element itself is not responsive, but @jmartin-sul and I have been looking at ways to handle this natively with videojs._

## bd786fy6312 (before)
bd786fy6312-before

## bd786fy6312 (after)
bd786fy6312-after

### DIFF
--- a/app/assets/stylesheets/common.scss.erb
+++ b/app/assets/stylesheets/common.scss.erb
@@ -9,11 +9,14 @@
 
 .#{$namespace}-container {
   background: $white-color;
+  border: 1px solid $color-pantone-405;
   color: $sul-font-color;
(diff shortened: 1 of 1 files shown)


ndushay:
LGTM if this has also been tested against browser for no change or for positive change ...

Issue #636, opened 2016-08-10
retrieve info from videoData element of PURL XML


jmartin-sul:
`videoData` should have attributes indicating duration (in ISO 8601 format), height, and width for the video file.  e.g.:
```XML
<videoData duration="PT10S" height="960" width="540"/>
```

`Embed::PURL::Resource::ResourceFile` should have methods making this info available to callers.

Issue #574, opened 2016-06-03
Add running time media index thumbnails


jkeck:
This is currently blocked by sul-dlss/media#106

<img src="https://cloud.githubusercontent.com/assets/6945691/15342894/8459bad4-1c4d-11e6-9981-972203195730.png" alt="running time" />

_**Note:** We should ensure that the lack of this data is handled properly as we should not expect all media objects to be remediated with this data when we launch._

PR #637, opened 2016-08-11
use iso8601 duration info from videoData to display video duration in parens after the title

jmartin-sul:
this adds a new [dependency](https://github.com/arnau/ISO8601) because:
* ruby's time libraries don't parse ISO 8601 durations
* ActiveSupport's Duration class does, but not until rails 5
* arnau's library seems well tested, and a better fit than the other options i looked at:
  * https://github.com/peleteiro/ruby-duration
  * https://github.com/janko-m/as-duration (just reimplements that AS4 duration)

AFAICT, nothing off the shelf provides for nice human readable printing of a duration object in the way you'd expect a media timer to do it.  as the mockup implies, this will print durations such that:
* all fields after the first non-zero field are included
* all fields after the first included field are zero padded

e.g.:
`P0DT1H2M3S` => `1:02:03` (i.e. 0 days, 1 hour, 2 minutes, 3 seconds)
`PT10S` => `0:10` (i.e. 10 seconds)
`P9D` => `9:00:00:00` (i.e. 9 days)

before:
screen shot 2016-08-11 at 2 32 44 pm

after (title on last object pushes duration out of view; @jkeck and i were inclined to say we can live with that, but also wanted input from @jvine to see if there was a better solution or a placement that'd keep that from happening):
screen shot 2016-08-11 at 2 33 24 pm

see also: https://en.wikipedia.org/wiki/ISO_8601#Durations (it turns out the actual ISO doc costs 138 swiss francs)

closes issue #636 
closes issue #574 
connect to sul-dlss/sul-embed#636 
connect to sul-dlss/sul-embed#574

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -127,6 +127,7 @@ module Embed
         nil
       end
     end
+
     class ResourceNotAvailable < StandardError
       def initialize(msg = 'The requested PURL resource was not available')
         super
@@ -137,6 +138,43 @@ module Embed
         super
       end
     end
+
+    class Duration < ISO8601::Duration
+      def to_s
+        return nil unless supported_duration?
+
+        # build up a list of the fields we want to use for output string
+        field_accumulator = []
+        [:years, :months, :days, :hours, :minutes, :seconds].each do |atom_type|
+          atom_val = atoms[atom_type]
+          # if we've already started accumulating fields, or this field is non-zero, add this field to the list.
+          # if we've hit the minutes field, start accumulating no matter what, since we always want to show minutes
+          # and seconds.
+          field_accumulator << atom_val if !field_accumulator.empty? || atom_val != 0 || atom_type == :minutes
+        end
+
+        # zero pad any field after the first, join with colons (e.g., returns '1:02:03' for 'P0DT1H2M3S',
+        # '2:03' for 'PT2M3S').
+        field_accumulator.map.with_index do |atom_val, idx|
+          (idx > 0) ? atom_val.to_i.to_s.rjust(2, '0') : atom_val.to_i.to_s
+        end.join ':'
+      end
+
+      private
+
+      # though they're valid ISO8601 durations, we don't support negative durations or durations specified in
+      # weeks, because those flavors of duration don't make much sense for the running time of a piece of media.
+      def supported_duration?
+        errors = []
+        errors << "#{self.class} does not support specifying durations in weeks" if atoms[:weeks] != 0
+        errors << "#{self.class} does not support specifying negative durations" if atoms.values.any? { |val| val < 0 }
+        return true if errors.empty?
+
+        errors.each { |e| Rails.logger.warn e }
+        false
+      end
+    end
+
     class Resource
       def initialize(resource, rights)
         @resource = resource
@@ -164,6 +202,7 @@ module Embed
           ResourceFile.new(file, @rights)
         end
       end
+
       class ResourceFile
         def initialize(file, rights)
           @file = file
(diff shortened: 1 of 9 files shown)


jmartin-sul:
attempted to break up into a reasonable set of logical commits.  can squash further if people see reason to do so.

app/models/embed/purl.rb
```diff
@@ -137,6 +138,30 @@ def initialize(msg = 'The requested PURL resource was not embeddable.')
         super
       end
     end
+
+    class Duration < ISO8601::Duration
+      def to_s
+        raise 'Duration#to_s does not support specifying durations in weeks' if atoms[:weeks] != 0
+        raise 'Duration#to_s does not support specifying negative durations' if atoms.values.detect { |val| val < 0 }
```
jkeck: Do we really want to raise an error in this case?  That would, AFAICT, cause the embed viewer to error out and not display anything for these objects.  Should we consider a more graceful approach for when the data is not what we want/expect?

jkeck: If we do keep these checks, I think `atoms.values.any? { |val| val < 0 }` might be more idiomatic than `atoms.values.detect { |val| val < 0 }`.

jmartin-sul: ah, yeah, fair point about the erroring out of the whole viewer, hadn't thought about that.

would you keep the check and do something milder, like logging?  or just ditch it entirely?

and yeah, i like `any?`, were we to keep the check, though i'm happy to get rid of it.

jmartin-sul: i guess another option would be some sort of text indicator of the error, like returning `--`, though that seems like it might just confuse users.

jvine:
> title on last object pushes duration out of view

Hard to do gracefully without some kind of calculation to add ellipsis to the title. Pin the duration to the 4th row, so it always appears in the same place; and if the title is longer than 3 rows (or 2 if it's restricted) it would be truncated. 

ellipsis

(And maybe expose the full title on hover/focus...?)

app/models/embed/purl.rb
jkeck: Yeah, I think logging something.  I'm not sure if there is a way in honeybadger (which we aren't even setup for yet, but no sense in looking into squash) to tell it to log/track an issue that doesn't necessarily raise an error through the actual stack.

But in the meantime, I think logging it would be good and just printing nothing for the duration.

jmartin-sul:
ok, updated to address feedback.  squashed fixes for @jkeck's suggestions into existing commits.

added a new commit for something like what @jvine suggested, but instead of limiting the text to 3 lines, it truncates it with an ellipsis at 45 chars.  this seems to display well in my manual testing, and as seen in the screenshot, there are at least some examples where the implemented solution shows more of the title than would otherwise be shown. but if that solution is unacceptable, it's easy enough to remove the commit.

screen shot 2016-08-11 at 7 58 47 pm

i didn't implement the full text on hover/focus suggestion, but would be happy to look into it (though maybe we should do so under a separate ticket?).

jmartin-sul:
also:  github is showing commits ba705f0 and 0569a11 in the wrong order:  the refactor in ba705f0 comes first, and should show up that way if you check out the branch and look at the log on the command line (ba705f0 still has an earlier timestamp, which i imagine is the cause of github's display confusion).

lib/embed/media_tag.rb
```diff
@@ -36,7 +36,7 @@ def to_html
     attr_reader :purl_document, :request, :viewer
 
     def media_element(label, file, type)
-      media_wrapper(label: label, stanford_only: file.stanford_only?, location_restricted: file.location_restricted?) do
+      media_wrapper(label: label, file: file) do
```
jkeck: 👍

app/models/embed/purl.rb
```diff
@@ -137,6 +138,35 @@ def initialize(msg = 'The requested PURL resource was not embeddable.')
...
+
+    class Duration < ISO8601::Duration
+      def to_s
+        errors = []
+        errors << 'Duration#to_s does not support specifying durations in weeks' if atoms[:weeks] != 0
+        errors << 'Duration#to_s does not support specifying negative durations' if atoms.values.any? { |val| val < 0 }
+        unless errors.empty?
+          errors.each { |e| Rails.logger.warn e }
+          return nil
+        end
```
jkeck: It feels like this method may have an extra responsibility that could be abstracted out.

What would you think about pulling this logic into private validation methods so we could do something along the lines of `return nil unless valid_duration?` at the top of `#to_s`?

Totally up to you.

jkeck:
Just one minor comment.  LGTM 👍

jmartin-sul:
ok, pushed up a new version with @jkeck's refactoring suggestion

ndushay:
LGTM

PR #638, opened 2016-08-11
Improve handling of embed customization parameters

jkeck:
This PR passes the `Embed::Request` context into the `EmbedThisPanel` so that the iframe code has access to the original request's customization parameters (such as whether to hide the title, download panel, etc).

This allows us to do two things:

1. The `embed/iframe` url that is in the `src` of the iframe returned in the oembed response now properly includes the incoming customization parameters.
1. The `EmbedThisPanel` now defaults to the customizations that were in the original request (i.e. if the embed the user is viewing was customized to hide the title, the embed code that the user gets from that object will default to having its title hidden as well).

## Before
iframe-options-before

## After
iframe-options

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ Viewers can customize the embed panel.  To do this, the viewer should override #
 
     def embed_this_html
       return '' if @request.hide_embed_this?
-      Embed::EmbedThisPanel.new(druid: @purl_object.druid, height: height, width: width, purl_object_title: @purl_object.title) do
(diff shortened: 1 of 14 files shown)


ndushay:
LGTM.  👍

tingulfsen:
I like your animated gifs(?), except that they're about 18 times faster than my brain can process...

jkeck:
Yeah, the ways I get the file down to an upload-able size is by removing every X frame as well as scaling the size.  😢

PR #644, opened 2016-08-16
Add ability to construct rtmp links for flash fallbacks

### LINKED ISSUE #631: Enable flash mp3 playback
Fall back to Flash for mp3 files (dependent on updates to wowza)

jkeck:
Also removes HLS playlist URLs for mp3 on browsers that don't support HLS natively.

Closes #631


ndushay:
This looks great to me.  👍   (I eyeballed it;  i didn't run it.)

PR #647, opened 2016-08-16
Remove the (hidden) hide metadata panel feature.

### LINKED ISSUE #641: Remove ability to hide the metadata panel
PURL is the only app that makes use of this option and with the addition of the IIIF drag and drop icon it might make sense to remove the ability to hide it (which will turn the feature on in PURL).

_See [related conversation](https://github.com/sul-dlss/sul-embed/pull/566#issuecomment-239874495)_

jkeck:
Closes #641

### DIFF
--- a/spec/lib/embed/embed_this_panel_spec.rb
+++ b/spec/lib/embed/embed_this_panel_spec.rb
@@ -88,8 +88,7 @@ describe Embed::EmbedThisPanel do
           maxheight: '555',
           maxwidth: '666',
           hide_title: 'true',
-          hide_embed: 'true',
-          hide_metadata: 'true'
+          hide_embed: 'true'
         )
       end
 
@@ -99,11 +98,6 @@ describe Embed::EmbedThisPanel do
         expect(src).to match(/&maxwidth=666/)
         expect(src).to match(/&hide_embed=true/)
         expect(src).to match(/&hide_title=true/)
-      end
-
-      it 'does not include parameters that are not intended to be passed' do
-        src = subject.find('iframe')['src']
-        expect(src).not_to match(/hide_metadata/)
       end
     end
   end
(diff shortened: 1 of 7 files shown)


spec/lib/embed/embed_this_panel_spec.rb
```diff
@@ -100,11 +99,6 @@
         expect(src).to match(/&hide_embed=true/)
         expect(src).to match(/&hide_title=true/)
       end
-
-      it 'does not include parameters that are not intended to be passed' do
```
ndushay: is this check no longer useful, or is "hide_metadata" no longer a good test string?

spec/lib/embed/request_spec.rb
```diff
@@ -45,17 +37,13 @@
...
-      Embed::Request.new(url: purl, hide_title: 'true', hide_metadata: 'true').as_url_params
+      Embed::Request.new(url: purl, hide_title: 'true').as_url_params
     end
 
     it 'is a hash of parameters to be turned into url optinos' do
       expect(url_params).to be_a Hash
       expect(url_params[:hide_title]).to eq('true')
     end
-
-    it 'does not include parameters that should not be passed along in a url (such as hide_metadata)' do
```
ndushay: as above, is this check no longer useful, or is "hide_metadata" no longer a good test string?

ndushay:
question about 2 tests;  o.w. LGTM.

spec/lib/embed/request_spec.rb
jkeck: 👍 I'll add a test back in.  This is whitelisting so testing that other things aren't making their way in is a good idea.  Good catch.

jkeck:
Thanks for the feedback @ndushay. I believe I have resolved your concern.

ndushay:
👍 for merging

PR #650, opened 2016-08-16
Add a target to links intended for downloading files from stacks.

### LINKED ISSUE #648: Give download links a target so they work inside an iframe
Some browsers will treat the link click as navigational and replace the embed viewer with blank content.  Adding `target="_blank"` to the download links in the download panel as well as the file viewer will ensure that the viewer's content is not replaced.

jkeck:
Closes #648 

This prevents certain browsers from viewing the link as navigational
and replacing the embed content with empty HTML.


ndushay:
LGTM.  👍

Issue #623, opened 2016-07-29
Handle Thumbnails (Object and File level) in Media Player


jkeck:
Currently, SUL Embed deals with all `<file>` elements w/i a `<resource>` as first-class files.  However; there are certain cases where they should not be treated as first-class files (particularly in the case of the media player).

- [ ] Do not treat files within `<resource type='thumb'>` or `<resource thumb='yes'>` as first-class files (i.e. they should not end up in the media filmstrip, and not be shown in the viewer like a video/audio file).
- [ ] Do not treat `<file type='thumb'>` or `<file thumb='yes'>` as first-class files.
- [ ] Replace icon in media filmstrip with thumb for objects that have a companion thumbnail `<file>` within the same `<resource>`.

jkeck:
@jvine a question came up today when discussing this ticket around object level thumbs and behavior in the file and media viewers.

FWIW, I _believe_ these are currently displaying in the file viewer (as we don't have any code parsing  thumb attributes).  It seems like we should not remove this behavior (as it my be odd to not have access to that anywhere on the detail/record view that uses the embed).

How should we handle these in the media viewer?  If we follow the rules above the object level thumb would only be accessible via the download panel.

jkeck:
dv738kn2168 (purl-test) is a good example object that has both an object-level thumb as well as a file-level thumb.

jmartin-sul:
discussed w/ @LynnMcRae, @jkeck, and @jvine...

what we came up with was the following set of rules, which in practice, mostly line up with the behavior in the description, though this seems like a simpler/clearer formulation:
* the media thumb slider will make available:
  * resources of type `audio` of the appropriate file type (e.g. `mp3`)
  * resources of type `video` of the appropriate file type (e.g. `mp4`)
  * resources of type `image` of the appropriate file type (e.g. `jp2`)
* an image file in a resource will be considered a resource level thumb.  there should only be one image file in a given resource, but if there are multiple, the first will be used.
  * the resource level thumb will show up in place of the generic media icon on the resource entry in the slider, and as the poster (keyframe) for the media player instance when the object is selected.
* other resource types (e.g. resource type `thumb` or resource type `file`) will be ignored by the media slider, which should automatically keep object level thumbs out of the media slider.

![thumbs for media resources 20160816](https://cloud.githubusercontent.com/assets/7741604/17718531/5929e1c2-63c9-11e6-957d-a83b7e6b8734.jpg)

[...]

PR #658, opened 2016-08-18
[WIP] resource level thumbnails

jmartin-sul:
before (chrome):
screen shot 2016-08-18 at 2 56 46 pm

after (chrome):
screen shot 2016-08-18 at 3 07 06 pm

note the extra copy of "The relaxing song.mp3" in the before version.  that was due to a bug where all files belonging to supported resources were listed, instead of just an entry for each resource (with the supported media file from the given resource).  additionally, there was a bug in determining the label in such cases, where the label from the first file in the resource would be used for all files under the resource.  both bugs were fixed as a natural consequence of the refactoring needed to implement this enhancement.  the old offending code was: https://github.com/sul-dlss/sul-embed/blob/8bc11d43f8058ddf98c10263bfaf986b38ec8c49/lib/embed/media_tag.rb#L16-L32

the WIP label on this is because there's a display issue in FF, where the div with the label text for a slider entry appears below the slider entry:
screen shot 2016-08-18 at 3 01 56 pm

advice from @jkeck or @jvine on the display issue?

this is currently deployed to embed-dev (and so should be used by purl-test).  ~~note that audio won't play for the test object for which the screenshots were taken, likely because there's a space in each of the file names, which we suspect embed is not properly encoding.  however, this should work for testing that audio still plays properly:  https://purl-test.stanford.edu/bb051hp9404~~ _note: audio should play for files with spaces now, including "Pizza Time!.mp3" and "The Relaxing Song.mp3" from the dv738kn2168 test object in the screenshots.  this got fixed with the merge of PR #660.  bb051hp9404 should still work also, of course._

closes sul-dlss/sul-embed#623
connects to sul-dlss/sul-embed#623

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -203,6 +203,16 @@ module Embed
         end
       end
 
+      # get first file of appropriate mime type
+      def media_file
+        files.find { |file| file.mimetype.match "^#{type}\/" }
+      end
+
+      # get first image file (if any) for the resource
+      def media_thumb
+        files.find(&:image?)
+      end
+
       class ResourceFile
         def initialize(file, rights)
           @file = file
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -15,34 +15,34 @@ module Embed
 
     def to_html
       output = ''
-      @file_index = 0 # TODO: this file index thing is set in a messy way
+      file_index = 0
       purl_document.contents.select { |resource| primary_file?(resource) }.each do |resource|
-        output << output_for_resource(resource)
-        @file_index += 1
+        output << output_for_resource(resource, file_index)
+        file_index += 1
       end
       output
     end
 
     private
 
-    attr_reader :file_index, :purl_document, :request, :viewer
+    attr_reader :purl_document, :request, :viewer
 
-    def output_for_resource(resource)
+    def output_for_resource(resource, file_index)
       label = resource.description
       if resource.media_file
         label = resource.media_file.title if label.blank?
-        media_element(label, resource)
+        media_element(label, resource, file_index)
       elsif resource.media_thumb
         label = resource.media_thumb.title if label.blank?
-        previewable_element(label, resource.media_thumb)
+        previewable_element(label, resource.media_thumb, file_index)
       else
         ''
       end
     end
 
-    def media_element(label, resource)
+    def media_element(label, resource, file_index)
       thumbnail = stacks_square_url(purl_document.druid, resource.media_thumb.title, size: '75') if resource.media_thumb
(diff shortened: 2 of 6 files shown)


app/models/embed/purl.rb
```diff
@@ -203,6 +203,16 @@ def files
         end
       end
 
+      # get first file of appropriate mime type
+      def media_file
+        files.find { |file| file.mimetype.match "^#{type}\/" }
```
jmartin-sul: IIRC i was once told to prefer `find` over `detect`.  whatever people want.  rubocop didn't complain, FWIW.  does not matter to me at all.

jkeck:
@jmartin-sul let's look @ that CSS issue together.

tingulfsen:
I think the CSS issue has to do with the value of the CSS white-space property set here:

https://github.com/sul-dlss/sul-embed/blob/master/app/assets/stylesheets/media.scss#L88-L91

Firefox seems to parse/compute the result differently than the other browsers.

jkeck:
@jmartin-sul try adding a `float: left` to the image element in the thumb.   There is a `float: left` on the `<i>` and if it exists on both then things seems to line up correctly.

On a side-note, it may be a good idea to add a class to both the `<i>` and the `` that contains the  common styling necessary for the display to be correct (which might just be the `float: left`).

lib/embed/media_tag.rb
```diff
@@ -149,10 +160,6 @@ def enabled_streaming_types
       Settings.streaming[:source_types]
     end
 
-    def file_index
```
jkeck: This is an interesting change.  I wasn't completely happy with its implementation, but it seemed to work.

Was there something not working with this under our new approach that requires us to pass this index around as an argument to 4 methods in the class?

jmartin-sul: discussed in person w/ @jkeck and he was 👌 w/ this change.

(i'd refactored this because i found the original version, where the counter was read inside a method but incremented in another place by the method's calling loop, to be a bit hard to read/follow, and while the parameter passing is a bit more verbose than i'd like, it did also feel a bit cleaner and easier to follow)

jmartin-sul:
@jkeck's float left suggestion appears to fix the problem in FF (and it still looked fine to me in chrome and safari):
screen shot 2016-08-25 at 12 30 40 am

jmartin-sul:
since the CSS issue appears to be fixed and the remaining question addressed, taking this out of WIP and putting back up for review.  @jkeck @tingulfsen @ndushay

jkeck:
I'm putting `[WIP]` back on this as we've been seeing oddities with the deployed version that makes @LynnMcRae think this is not necessarily meeting the spec.  I think we need to evaluate this closer before we merge.

jmartin-sul:
superseded by PR #696

PR #660, opened 2016-08-18
Treat the video/audio tag like cdata so Nokogiri doesn't try to URL…

jkeck:
…escape the streaming URL.


jmartin-sul:
LGTM, deployed to embed-dev and it works, merging...

PR #662, opened 2016-08-19
Move some blank targets to parent targets and add some explanatory`…

jkeck:
…text about which targets to use in the README.

Also added a `rel` attribute to `_blank` targets for good measure (even though we control _almost_ every endpoint we link to using this method, it is still good practice to include this attribute).

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,9 @@ The file that the class is defined in (or your preferred method) should register
 
 ### Linking in viewers
 
-The rich HTML payload that is supplied via the oEmbed API is an iframe. This means that all consumers will be embedding an iframe into their page. Given this fact, generating links will require explicit targets if they are not intended to internally replace embed content.  Given this, there are two patterns that can be used.  For links intended to download files, a `target="_blank"` can be used (effectively opening a new tab for the download which is immediately closed). For links that are intended to navigate the users browser away from the current page (e.g. the links to Memento/GeoBlacklight/etc.) then `target="_parent"` should be used to give the link the default browser behavior. [More about link targets](http://www.w3schools.com/tags/att_a_target.asp).
+The rich HTML payload that is supplied via the oEmbed API is an iframe. This means that all consumers will be embedding an iframe into their page. Given this fact, generating links will require explicit targets if they are not intended to internally replace embed content.  Given this, there are two patterns that can be used.  For links intended to download files, a `target="_blank"` can be used (effectively opening a new tab for the download which is immediately closed).  When using `target="_blank"` add `rel="noopener noreferrer"` **particularly** when linking externally (although this should be reserved for linking to internal resources when possible). See [this blog post](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/) for an explanation. *Note: This does not apply to WebAuth links.*
+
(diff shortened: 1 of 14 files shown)


tingulfsen:
👍

README.md
```diff
@@ -104,6 +104,10 @@ The file that the class is defined in (or your preferred method) should register
     Embed.register_viewer(Embed::Viewer::DemoViewer) if Embed.respond_to?(:register_viewer)
 
 
+### Linking in viewers
+
+The rich HTML payload that is supplied via the oEmbed API is an iframe. This means that all consumers will be embedding an iframe into their page. Given this fact, generating links will require explicit targets if they are not intended to internally replace embed content.  Given this, there are two patterns that can be used.  For links intended to download files, a `target="_blank"` can be used (effectively opening a new tab for the download which is immediately closed). For links that are intended to navigate the users browser away from the current page (e.g. the links to Memento/GeoBlacklight/etc.) then `target="_parent"` should be used to give the link the default browser behavior.
+
```
ndushay: what do you think about having examples of each link here?

jkeck: What do you mean by example?

ndushay:
This looks good, except I think link examples would really help if someone like me had to understand the diff between _blank and _parent.

README.md
ndushay: The paragraph above speaks to two patterns, and seeing an actual example (does `target="_blank"`  go in an `<a>` tag?) with the explanation attached makes it easier for me to understand.

- `<a href=aasdfasdfsad blah blah blah>`  would give xxx behavior
- `<a href=asdfdafasdf yadda yadda>`  would give yyy behavior

jkeck: Ah, okay.  This is standard HTML behavior, so I've instead added a link to a page explaining more about the target attribute of `<a>` tags.

README.md
```diff
@@ -104,6 +104,12 @@ The file that the class is defined in (or your preferred method) should register
...
 
 
+### Linking in viewers
+
+The rich HTML payload that is supplied via the oEmbed API is an iframe. This means that all consumers will be embedding an iframe into their page. Given this fact, generating links will require explicit targets if they are not intended to internally replace embed content.  Given this, there are two patterns that can be used.  For links intended to download files, a `target="_blank"` can be used (effectively opening a new tab for the download which is immediately closed).  When using `target="_blank"` add `rel="noopener noreferrer"` **particularly** when linking externally (although this should be reserved for linking to internal resources when possible). See [this blog post](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/) for an explanation. *Note: This does not apply to WebAuth links.*
+
+For links that are intended to navigate the users browser away from the current page (e.g. the links to Memento/GeoBlacklight/etc.) then `target="_parent"` should be used to give the link the default browser behavior. [More about link targets](http://www.w3schools.com/tags/att_a_target.asp).
+
```
ndushay: thank you.   this is much better for me.

lib/embed/viewer/media.rb
```diff
@@ -54,7 +54,7 @@ def download_list_items
             file_size = "(#{pretty_filesize(file.size)})" if file.size
             <<-HTML.strip_heredoc
               <li class='#{('sul-embed-stanford-only' if file.stanford_only?)}'>
-                <a href='#{file_url(file.title)}' title='#{file.title}' target='_blank'>
+                <a href='#{file_url(file.title)}' title='#{file.title}' target='_blank' rel='noopener noreferrer'>
```
ndushay: one more nitpick - should a test be looking for  `rel='noopener noreferrer'` ?

jkeck: Added.

PR #663, opened 2016-08-23
Update default configurations to use https all the time.

### LINKED ISSUE #629: Go all https all the time in embed
Currently we allow connections over http since we use protocol-relative URLs (e.g. `//embed.stanford.edu` instead of `https://embed.stanford.edu`).

We should evaluate going all https all the time (which may be better for streaming media).  This should be able to be handled almost completely in configuration on the embed side, but we may want to also change the apache configuration to redirect http traffic to https.

jkeck:
Closes #629 

This PR turns the sandbox into a protocol relative URL since it is intended to work under http locally and under https remotely.

All the other configs have been turned from protocol relative to be https only.  We'll also need to update remote configs when we deploy.

### DIFF
--- a/app/views/pages/sandbox.html.erb
+++ b/app/views/pages/sandbox.html.erb
@@ -83,7 +83,7 @@
 
 <body>
   <form action="" class="form-params">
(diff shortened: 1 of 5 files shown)


ndushay:
👍

PR #673, opened 2016-08-25
Add access restriction message for screen readers.

### LINKED ISSUE #668: Location-restricted message isn't perceivable by screen reader
The "restricted" message doesn't get read by the screenreader when focus is on the video. 

Add an id to the message div, and add aria-labelledby="id" to the video element.

`<div class="sul-embed-media-access-restricted" id="message1">`

`<video id="sul-embed-media-1" aria-labelledby="message1"...`

tingulfsen:
Closes #668

### DIFF
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -43,6 +43,7 @@ module Embed
             data-src="#{streaming_url_for(file, :dash)}"
             data-auth-url="#{authentication_url(file)}"
             controls='controls'
+            aria-labelledby="access-restricted-message-div-#{file_index}"
             class="#{'sul-embed-many-media' if many_primary_files?}"
             style="height: #{media_element_height}; display:none;"
             height="#{media_element_height}">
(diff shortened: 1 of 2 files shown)


jvine:
If you use the same ID for every message in an object that has more than one restricted item, the screen reader may grab the first message with that ID in the DOM, regardless of which one is actually displayed on the page. ChromeVox does this. I think each pair (item + corresponding message) needs a unique ID. There might also be something you could do with `aria-hidden` but I haven't been able to make it work.

lib/embed/media_tag.rb
```diff
@@ -127,7 +128,7 @@ def access_restricted_overlay(stanford_only, location_restricted)
       <<-HTML.strip_heredoc
         #{location_only_overlay(stanford_only, location_restricted)}
         <div class='sul-embed-media-access-restricted-container' data-access-restricted-message>
-          <div class='sul-embed-media-access-restricted'>
+          <div class='sul-embed-media-access-restricted' id='access-restricted-message-div'>
```
jkeck: You should be able to use `file_index` here to (and above in the `aria-labelledby` attribute) to give a unique id to these elements.

jkeck:
@tingulfsen a branch that you have rebased this on top of (I think) makes the video/audio element `display: none` by default.  If you have capybara assertions that deal w/ the raw HTML (and does not trigger the entire JS stack like a full integration test) then you need to add `visible: false` to the `expect`.

ndushay:
LGTM.  👍

jkeck:
LGTM.  Should we squash any of these commits before 🚢ing?

PR #686, opened 2016-08-31
Update the height of the file viewer based on the number of resources…

### LINKED ISSUE #677: Gracefully display file lists that are much shorter than default 400 pixels tall
The height of the iframe needs to be specified or else defaults to 400px tall. In the case where the viewer doesn’t take that space (i.e. when there is only 1 or 2 files), a lot of extra whitespace appears around the viewer.

Example: https://purl.stanford.edu/bb054mz1203

jkeck:
… being displayed.

Closes #677 

## bb054mz1203 (before)
bb054mz1203-before

## bb054mz1203 (after)
bb054mz1203-after

## bd786fy6312 (before)
bd786fy6312-before

## bd786fy6312 (after)
bd786fy6312-after

## zw177zm0384 (before)
zw177zm0384-before

## zw177zm0384 (after)
zw177zm0384-after

## tn629pk3948 (before)
tn629pk3948-before

## tn629pk3948 (after)
tn629pk3948-after

### DIFF
--- a/lib/embed/viewer/file.rb
+++ b/lib/embed/viewer/file.rb
@@ -65,7 +65,23 @@ module Embed
       end
 
       def default_body_height
-        400 - (header_height + footer_height)
+        file_specific_body_height - (header_height + footer_height)
+      end
+
+      # This is neccessary because the file viewer's height is meant to be dynamic,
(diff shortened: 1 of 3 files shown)


ndushay:
LGTM  👍

Issue #685, opened 2016-08-31
image viewer thumbnail index keyboard/screen reader accessibility


ndushay:
See #669, which indicated what the issues were for the media_viewer, which uses thumb_slider.js, which is HTML5.    The image_x_viewer has its own thumb_slider code and altering it was out of scope for #669.

mejackreed:
No longer relevant

Issue #669, opened 2016-08-24
Thumbnail index is not keyboard/screen-reader accessible (media viewer)


jvine:
(used to be fine; got broken somewhere; media didn't break it as it affects images, too)

Currently the "..." icon and the list items don't get keyboard focus. And once the index is open, the list items don't get keyboard focus either.

Here's a model that is working:
https://searchworks.stanford.edu/?utf8=%E2%9C%93&search_field=search&q=5762702
(the green Online button opens a list of links)

- the div that contains the "..." icon should be a `<button>` element
- the aria-expanded attribute is not set correctly:  it says aria-expanded="true" all the time. it should toggle true/false as the div is opened/closed
- Bootstrap (and SearchWorks) has the aria-expanded attribute on both the parent button and the child div - it's not clear to me that both are necessary, but it doesn't hurt anything - so I'm guessing we should add it to the thumb-slider div and toggle it to match the "..." button
- when the thumb-slider is open, the list items within it don't get focus. The divs should be (or contain)  `<a>` elements with an href or tabindex="0".

ndushay:
hrmm - we need to look at https://jirasul.stanford.edu/jira/browse/MEDIAINFRA-44  and determine what work remains.   @hannahfrost @jkeck

ndushay:
@jvine   help!  I'm trying to allow each list element in the thumb-slider to get focus.  I've tried a bunch of things, but only the active element gets focus in Firefox.  

![list element](https://cloud.githubusercontent.com/assets/96775/18108419/f7ee95b2-6ec0-11e6-9290-5637c643c64a.png)

That works in Chrome, but not FF.  In addition to `href="javascript:;"` I've tried   `href="0"`, `tabindex="0"`, `tabindex="n"` (where n is the index to the list) ...   but nothing seems to work in FF.

ndushay:
- [x]  the div that contains the "..." icon should be a `<button>` element - done.
- [x] add aria-expanded attribute to the thumb-slider div 
- [ ] when the thumb-slider is open, the list items within it don't get focus. The divs should be (or contain)  `<a>` elements with an href or tabindex="0".  

probably will be a separate ticket
- [ ] the aria-expanded attribute is not set correctly:  it says aria-expanded="true" all the time. it should toggle true/false as the div is opened/closed;  same for thumb-slider div aria-expanded attr.

jvine:
[...]

PR #687, opened 2016-08-31
Thumbindex accessible

ndushay:
Addresses 3 out of 4 parts of #669:

- the div that contains the "..." icon should be a `<button>` element
- add aria-expanded attribute to the thumb-slider div 
- when the thumb-slider is open, the list items within it don't get focus. The `<div>` elements should be (or contain)  `<a>` elements with an `href="#"`

Note that toggling aria-expanded will be done in separate PR.  Note further that we are scoping #669 to just the media_viewer.   #685 is the issue for the image viewer, and is out of scope.   media_viewer is the only consumer of thumb_slider.js, so it is the only calling code that needs the list items changed to anchor tags (from div tags).

#### the div that contains the "..." icon should be a `<button>` element

This is so it can get the focus without the mouse.  Note that it has the focus in the picture.

thumbindex as button

#### add aria-expanded attribute to the thumb-slider div 

Per the ticket.  Note that toggling aria-expanded will be done in separate PR.

thumbindex div aria-expanded

#### when the thumb-slider is open, the list items within it don't get focus. The `<div>` elements should be (or contain)  `<a>` elements with an `href="#"`

This is so it can get the focus without the mouse.  Note that it has the focus in the picture.

thumbindex list elements as anchors

Connects to #669

### DIFF
--- a/app/assets/javascripts/modules/thumb_slider.js
+++ b/app/assets/javascripts/modules/thumb_slider.js
@@ -17,7 +17,8 @@
 
     var thumbList =
       jQuery(
-        '<div class="sul-embed-thumb-slider" style="display:none">' +
+        '<div class="sul-embed-thumb-slider" style="display:none"' +
+          ' aria-expanded="false">' +
           '<ul></ul>' +
         '</div>'
       );
@@ -84,6 +85,7 @@
           });
           thumbList.find('ul').append(thumbnail);
         });
+
       },
 
       scrollFrame: function() {
--- a/app/assets/javascripts/modules/thumb_slider.js
+++ b/app/assets/javascripts/modules/thumb_slider.js
@@ -10,9 +10,9 @@
 
     var toggleControl =
       jQuery(
-        '<div class="sul-i-navigation-show-more-1 sul-embed-thumb-slider-open-close"' +
+        '<button class="sul-i-navigation-show-more-1 sul-embed-thumb-slider-open-close"' +
           ' aria-expanded="true" aria-label="toggle thumbnail viewer">' +
-        '</div>'
+        '</button>'
       );
 
     var thumbList =
--- a/.jshintrc
+++ b/.jshintrc
@@ -35,7 +35,7 @@
   "latedef": true,
 
   // Enforce line length to 80 characters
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/modules/thumb_slider.js
```diff
@@ -10,14 +10,15 @@
...
           ' aria-expanded="true" aria-label="toggle thumbnail viewer">' +
-        '</div>'
+        '</button>'
       );
 
     var thumbList =
       jQuery(
-        '<div class="sul-embed-thumb-slider" style="display:none">' +
+        '<div class="sul-embed-thumb-slider" style="display:none"' +
+          ' aria-expanded="true">' +
```
jkeck: Since the default state of the slider is closed should the `aria-expanded` default to `false` (and then we'll toggle it to `true` once it is opened)?

jkeck:
Is the change in background color of the open/close toggle intentional?

## Before
button-background-old

## After
button-background-new

ndushay:
I didn't notice a change in color.  Could it be the diff between selected and not selected?  I didn't change any css or ids or anything ...

app/assets/javascripts/modules/thumb_slider.js
ndushay: good idea.  will fix.

jkeck:
Hmm, it appears to have that color regardless of if it is selected or not (notice the gray color in your screenshots vs. what is currently deployed).

I think it may be the button's default styling coming through. (where the native element that it is replacing is not natively styled, as much, by the browser)

Typically with buttons you'll need to "unstyle" the background color and borders.

app/assets/javascripts/modules/thumb_slider.js
```diff
@@ -37,12 +38,44 @@
       }
     }
 
+    function toggleAriaAttributes() {
```
jkeck: Should we remove this until the next PR that will add this behavior (particularly given the log statements)?

ndushay:
fixed the css with help of @jkeck  so button styling isn't applied and background color is ok.

PR #688, opened 2016-08-31
thumb_slider: toggle aria-expanded appropriate when clicking '...' button

### LINKED ISSUE #669: Thumbnail index is not keyboard/screen-reader accessible (media viewer)

ndushay:
closes #669 by completing the fourth of four fixes listed in that ticket.

`aria-expanded` attribute for for `...` button and for the div containing the thumbslider list now are toggled appropriately to true and false when the button is clicked.

#### before clicking `...` button
aria before click

#### after clicking `...` button
aria after click 1

#### after clicking `...` button again
aria after click 2

### DIFF
--- a/app/assets/javascripts/modules/thumb_slider.js
+++ b/app/assets/javascripts/modules/thumb_slider.js
@@ -11,7 +11,7 @@
     var toggleControl =
       jQuery(
         '<button class="sul-i-navigation-show-more-1 sul-embed-thumb-slider-open-close"' +
-          ' aria-expanded="true" aria-label="toggle thumbnail viewer">' +
+          ' aria-expanded="false" aria-label="toggle thumbnail viewer">' +
         '</button>'
       );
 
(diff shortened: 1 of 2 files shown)


jkeck:
@tingulfsen I paired on this w/ @ndushay so maybe we can get your :eyes: on it.

PR #691, opened 2016-09-02
Add (Restricted) text to links in the download panel as well as the file viewer

### LINKED ISSUE #683: Add (Restricted) text to file viewer and download links
Currently there is no way for a user to know a particular file is location restricted from the file viewer or in the download panel.

We need to add the `(Restricted)` text to both of these places to indicate that the content is location restricted.

This is one of the pieces of #667

_**Note:**  I looked into the stacks code, and unfortunately the [403 page is static HTML](https://github.com/sul-dlss/digital_stacks_rails/blob/master/public/403.html), so doing anything intelligent with that page is going to prove to be a bit more difficult._

jkeck:
Closes #683 

## gm746xg1831 (via file viewer)
gm746xg1831-as-file

## gm746xg1831 (via media viewer)
gm746xg1831-as-media-index

## gm746xg1831 (via media viewer)
gm746xg1831-as-media-download

### DIFF
--- a/app/assets/stylesheets/common.scss.erb
+++ b/app/assets/stylesheets/common.scss.erb
@@ -108,9 +108,16 @@
   }
 }
 
-.#{$namespace}-stanford-only a{
-  @include stanford-only;
-  padding-right: 20px;
+.#{$namespace}-stanford-only a {
+  @include stanford-only(right, 20px);
+}
+
+.#{$namespace}-stanford-only-text {
+  @include stanford-only(left, 15px);
+}
+
+.#{$namespace}-location-restricted-text {
+  color: $sul-restricted-text-color;
 }
 
 .#{$namespace}-header {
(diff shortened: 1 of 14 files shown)


app/assets/stylesheets/common.scss.erb
```diff
@@ -108,9 +108,16 @@
   }
 }
 
-.#{$namespace}-stanford-only a{
-  @include stanford-only;
-  padding-right: 20px;
+.#{$namespace}-stanford-only a {
```
ndushay: so is the "a" standing alone here implying an anchor tag?

ndushay:
LGTM  👍

app/assets/stylesheets/common.scss.erb
jkeck: Yes.  This is basically saying, style _an anchor tag within an element classed `.sul-embed-stanford-only`_.

PR #692, opened 2016-09-03
add running time to audio resource thumbnails in media viewer thumbnails index 

### LINKED ISSUE #665: add duration to audio resources in media player
currently, we only parse duration info from the `videoData` tag.  there's an analogous tag for audio resources called `audioData`.  shouldn't be too difficult to generalize the video duration machinery to handle audio duration also (the two appear to be structurally identical in terms of both the PURL XML and the presentation logic).

see:
https://github.com/sul-dlss/sul-embed/blob/9057b962992ef760377d6d062550dfeb16aa3729/app/models/embed/purl.rb#L279  (`PURL::Resource::Resource#video_duration_attr_str`)
https://github.com/sul-dlss/sul-embed/blob/9057b962992ef760377d6d062550dfeb16aa3729/app/models/embed/purl.rb#L248  (`PURL::Resource::Resource#video_duration`)
https://github.com/sul-dlss/sul-embed/blob/9057b962992ef760377d6d062550dfeb16aa3729/lib/embed/media_tag.rb#L71  (sets the `data-duration` attr on the media element to the result of `video_duration`, if there is a non-nil result)

one possible solution off the top of my head:
* add an `audio_duration_attr_str` method as a sibling to `Resource#video_duration_attr_str`
* likewise, we could and an `audio_duration` as a sibling to `Resource#video_duration`
* and maybe a general `duration` method as a sibling to `audio_duration` and `video_duration`, which would call the appropriate method (`audio_duration` or `video_duration`) based on the resource type
* then media_tag could just set the `data-duration` attribute to the result of `Resource#duration` instead of `Resource#video_duration`.

ndushay:
fixes #665.

audio duration


spec/models/embed/purl_spec.rb
```diff
@@ -234,24 +234,37 @@
...
+            expect(video.duration('video').to_s).to eq '1:02:03'
           end
         end
         context 'duration in an invalid format' do
           before { stub_purl_response_with_fixture(invalid_video_duration_purl) }
           let(:video) { Embed::PURL.new('12345').contents.first.files.first }
           it 'should return nil and log an error' do
             expect(Honeybadger).to receive(:notify).with("ResourceFile\#video_duration ISO8601::Errors::UnknownPattern: 'PDDTMMS'")
-            expect(video.video_duration).to be_nil
+            expect(video.send(:video_duration)).to be_nil
```
tingulfsen: Is there any particular reason to use .send() here?

tingulfsen:
One very minor question, looks good to me though. 👍

app/models/embed/purl.rb
```diff
@@ -280,6 +277,32 @@ def video_duration_attr_str
...
+        def audio_data?
+          @file.xpath('./audioData').present?
+        end
+
+        def audio_duration_attr_str
+          @file.xpath('./audioData').first.attributes['duration'].try(:text) if audio_data?
+        end
+
+        def audio_duration
+          Duration.new(audio_duration_attr_str) if audio_duration_attr_str
```
jkeck: What would you think about having a `duration` method that calls `Duration.new` and passes in either the `audioData` or `videoData` value into it.  We might be able to reduce the number of methods that this class needs, which could reduce some complexity.

I think we're running into a lot of places where a lack of proper abstraction is making things clumsy, so totally understand if you don't want to do that as part of this work.  My hope is that we can eventually find the proper abstraction between `audioData`, `videoData`, and `imageData` (or at least start coalesce the approaches between video/audio).

lib/embed/media_duration.rb
```diff
@@ -0,0 +1,65 @@
...
+          # if we've already started accumulating fields, or this field is non-zero, add this field to the list.
+          # if we've hit the minutes field, start accumulating no matter what, since we always want to show minutes
+          # and seconds.
+          field_accumulator << atom_val if !field_accumulator.empty? || atom_val != 0 || atom_type == :minutes
+        end
+
+        # zero pad any field after the first, join with colons (e.g., returns '1:02:03' for 'P0DT1H2M3S',
+        # '2:03' for 'PT2M3S').
+        field_accumulator.map.with_index do |atom_val, idx|
+          (idx > 0) ? atom_val.to_i.to_s.rjust(2, '0') : atom_val.to_i.to_s
```
ndushay: I think the parens are helpful, personally.

PR #694, opened 2016-09-07
Remove use of includes in javascript, since it don't work.

jkeck:


tingulfsen:
👍  from me, this is what we want (although we paired). Just FYI, this is the second part of the fix for sul-dlss/media#122 and is necessary because the Microsoft browsers don't implement .include().

PR #696, opened 2016-09-07
Handle file and object level thumbnails in the media viewer

### LINKED ISSUE #623: Handle Thumbnails (Object and File level) in Media Player

jkeck:
Closes #623 

## dv738kn2168 (before)
dv738kn2168-before

## dv738kn2168 (after)
dv738kn2168

### DIFF
--- a/app/assets/stylesheets/media.scss
+++ b/app/assets/stylesheets/media.scss
@@ -60,6 +60,7 @@
   }
 
   .#{$namespace}-media-square-icon {
+    float: left;
     margin-right: 5px;
     width: auto;
   }
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -172,6 +172,10 @@ module Embed
         end
       end
 
+      def non_thumbnail_files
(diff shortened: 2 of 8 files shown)


ndushay:
LGTM  👍 .   Dunno if hound has a point or not.

jkeck:
I think Hound's complaint is an odd Hound vs. Rubocop inconsistency.  Rubocop does not complain about it, and I'm inclined to go with that.

PR #697, opened 2016-09-08
Purl::Resource::ResourceFile - refactoring

ndushay:
Given how height and width are used, is there any reason to have the methods for image height and width be distinct from the ones for video?   It's looking for the same attribute names, just on different elements.  

Either it has a height, or it doesn't.
Similarly for width.
Similarly for duration, in PR #692.


jmartin-sul:
LGTM, will merge when tests pass...

PR #698, opened 2016-09-08
Ensure videoJS audio and video elements are 100% height and width in …

### LINKED ISSUE #693: Address full screen scaling of media
There is a full screen class and we can remove any height/width restrictions we've added to the media.

jkeck:
…fullscreen.

Closes #693 

## Before
before

## After
after

### DIFF
--- a/app/assets/stylesheets/media.scss
+++ b/app/assets/stylesheets/media.scss
@@ -101,8 +101,13 @@
     visibility: hidden;
   }
 
+  // Ensure fullscreen media is as large as possible.
   &.vjs-fullscreen {
-    height: auto;
+    audio,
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/media.scss
```diff
@@ -101,8 +101,13 @@
     visibility: hidden;
   }
 
+  // Ensure fullscreen media is as large as possible.
   &.vjs-fullscreen {
-    height: auto;
+    audio,
+    video {
+      height: 100% !important;
```
jkeck: `!important` is being used here and above because we're overriding styles applied directly to the element.

Issue #630, opened 2016-08-04
Multi-item object film-strip conceals body of content


snydman:
@jvine @ggeisler - I notice in the wonderful media sprint demo the capability of displaying multiple first-class items in the embed view for an object (eg, audio file plus thumb).  The UX convention used to reveal the multi-items is similar to the convention used for a multi-image object filmstrip (such as a book or manuscript). 

In both cases this filmstrip, when revealed conceals part of the body of the object (the video, the image, etc.)  The image viewer is being reworked so that the film strip is below the image always and does not conceal the main image in the viewport.  An alternative approach would be to move the film strip to the left of the embed widget and position it vertically. 

Is adjusting the layout of the embed widget so that the filmstrip does not block the content something we can consider?

snydman:
see: 
![image](https://cloud.githubusercontent.com/assets/3279382/17421115/2348e600-5a5c-11e6-84da-4f5042ecd573.png)

aeschylus:
Previously when we've talked about this, the interim solution was to have the filmstrip closed by default. I'll point to where that can be done, but a deeper fix for this would be something similar to Mirador's sidepanel/new bottom panel. The content area should be in the negative space left by the bottom panel when it's open. When the bottom panel closes, the content area should expand as it shrinks, sort of "see-sawing" it into position so that whatever is currently framed in the viewport is simply framed in a smaller/larger space that it was before.

To change the default, have a look at: 
https://github.com/sul-dlss/sul-embed/blob/master/app/assets/javascripts/modules/image_x_viewer.js#L43-L50

That is where the toggle is performed. The toggle reads from the layout state store. To understand this language see: https://scotch.io/tutorials/getting-to-know-flux-the-react-js-architecture

To change the default see the layout store itself: 
https://github.com/sul-dlss/sul-embed/blob/master/app/assets/javascripts/modules/layout_store.js#L19-L30

At least for the image viewer, the long-term content-area will be handled through the manifestLayouts API and reshaping will likely go here: https://github.com/sul-dlss/sul-embed/blob/master/app/assets/javascripts/modules/image_x_viewer.js#L40-L42

[...]

PR #702, opened 2016-09-08
image viewer thumb filmstrip now closed by default, …

ndushay:
… like media viewer thumbstrip.

#### before fw090jw3474

embed first state for this purl

fw090jw3474 before

#### after fw090jw3474

embed first state for this purl

fw090jw3474 after

connects to #630


jmartin-sul:
code changes and manual testing both LGTM

PR #703, opened 2016-09-09
Change how the thumb slider hides objects to be more flash friendly

### LINKED ISSUE #670: Flash player does not resume playing media after toggling to different media
Steps to reproduce:

1. Navigate to an object with multiple top-level files that will initiate a flash player.
  * (e.g. dv738kn2168 on test because it has a multiple MP3s)
1. Play the first object.
1. Open the thumbnail index and select a different top-level file (any, it doesn't matter).
1. Navigate back to the original media that was paused and try to resume.

I believe this may be related to hiding/showing the flash player.

Potentially pertinent information:
* https://github.com/videojs/video.js/issues/2755
* https://github.com/videojs/video.js/issues/2752
* https://github.com/videojs/video.js/blob/master/docs/guides/removing-players.md

jkeck:
Closes #670 

This also includes a commit that pauses the media via the VideoJS API (as I noticed once I got flash working w/ the toggle that it was not pausing).

### DIFF
--- a/app/assets/javascripts/modules/media_viewer.js
+++ b/app/assets/javascripts/modules/media_viewer.js
@@ -99,9 +99,9 @@
     }
 
     function pauseAllMedia() {
-      var mediaObjects = $('.sul-embed-media audio, .sul-embed-media video');
+      var mediaObjects = jQuery('.sul-embed-media').find('.video-js');
       mediaObjects.each(function() {
-        this.pause();
+        videojs(jQuery(this).attr('id')).pause();
       });
     }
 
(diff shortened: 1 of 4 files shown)


ndushay:
this looks like a very slick solution.  👍    

FYI:  I tried to run the branch at home on FF, but video doesn't seem to like me doing FF from home.  If I don't VPN, I can't get to the stacks;  if I do VPN, then I can't play the video (true also for split-tunnel VPN).  Odd, but that is consistent for me at home.    I ran into similar probs on Chrome.

jkeck:
Unless you're changing the code to force flash (or your FF is loading Flash instead of HTML5), I wouldn't expect you to be able to play the video in any non-safari browser running locally (from home or on campus).

If you want I can show you how to force your local env. to use flash for testing.

jmartin-sul:
code changes LGTM.  tested locally and appears to be working as expected.  merging...

PR #704, opened 2016-09-10
Force the use of Flash when the browser is IE or Edge on Windows 10.

### LINKED ISSUE #701: Force Flash on Windows IE/Edge
With Windows 10 64-bit, the latest versions of Internet Explorer (version 11.0.10586.0) and Edge (version 25.10586.0.0) exhibit some strange behavior with our VideoJS player:

* Internet Explorer will play the first 11 seconds of a video, then stop and the video becomes unresponsive.
* Edge plays the videos fine.
* Internet Explorer plays audio fine.
* Edge refuses to play audio.

This was tested with our current Sul-Embed, which uses the VideoJS HLS plugin to play video and plays audio (MP3) with Flash.

Forcing the use of Flash with the Microsoft browsers for all content appears to work fine.

Note that Chrome and Firefox both seem to work as usual, both with and without Flash.

tingulfsen:
Fixes #701 

This PR pulls in the Client.js browser fingerprinting library, a small JavaScript library that I think does a fairly decent job at figuring out what browser is being used. It mainly uses the user agent string for what we're doing here, I believe, but is probably more robust than rolling our own UA parsing.

This PR updates the media viewer JS code to force the use of Flash on Windows 10 with the Microsoft browsers (Edge and IE). This is due to a known issue with the VideoJS HLS plugin:

https://github.com/videojs/videojs-contrib-hls#known-issues

https://github.com/videojs/videojs-contrib-hls/issues/833

Hopefully this problem will be resolved either by the VideoJS HLS plugin authors or by Microsoft in the not too distant future. Until then, the only recommended solution is to force Flash. Note that due to the nature of the bug, feature detection won't help us solve this one - hence browser detection/fingerprinting.


app/assets/javascripts/modules/media_viewer.js
```diff
@@ -146,11 +147,22 @@
       return hlsSource.prop('src').indexOf('/mp3:') >= 0;
     }
 
+    function ms10Browser() {
+      var fingerprinter = new ClientJS();
```
jkeck: You can avoid this by adding `/*global ClientJS */` at the top of the file.

tingulfsen: I'm pulling in Client.js, so it *is* in fact defined. Let me know if there's a way to make Hound quiet.

app/assets/javascripts/modules/media_viewer.js
```diff
@@ -146,11 +147,22 @@
...
+               fingerprinter.getOS().toLowerCase() === 'windows' &&
+               fingerprinter.getOSVersion().toLowerCase() === '10');
+    }
+
     // We must use flash for MP3s on browsers that do not support HLS natively.
     // This is a simple boolean function that can easily tell us if the current
-    // browser does not support HLS and the media is an MP3
+    // browser does not support HLS and the media is an MP3. We also use flash
+    // on Windows 10 Microsoft browsers (IE and Edge) because of a known issue
+    // with the VideoJS HLS plugin.
```
jkeck: What would you think about including the link to the known-issues section (that was in the PR comment) here in the comment?

That way another developer can easily check that page from the code to see if the known-issue still exists (and they don't have to track the PR down for context).

app/assets/javascripts/modules/media_viewer.js
jkeck: Adding `/*global ClientJS */` to the top of the file is how you will make Hound quiet (which is what we're doing there w/ the `ThumbSlider` and `videojs` globals).

tingulfsen:
I think that except for the weird anomaly with Nicholas' Firefox setup, this branch has been shown to work across the various operating systems and browsers that we want to support, so deleting the WIP designation.

app/assets/javascripts/modules/media_viewer.js
ndushay: I keep wanting to thumbsup this PR, but I don't know if y'all agreed on punting `/*global ClientJS */` or not.

jkeck: We should add the comment I describe to let hound know that a global is declared elsewhere that we will be using in this file (as is the established pattern).

PR #705, opened 2016-09-12
Add a simple rack app to stub the Stacks auth endpoint...

### LINKED ISSUE #567: Write integration test for Media Auth Link
We do not have a good way of stubbing an external URL request made via JS in our integration tests.

Since the relevant Stacks code is not in production yet, we cannot write an integration test around the log-in links for media.  Once it has been deployed, we can do that much easier.

jkeck:
...and include some integration tests for the WebAuth login link being shown or hidden.

Closes #567

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -1,3 +1,4 @@
+Dir[Rails.root.join('spec/support/stub_apps/*.rb')].each { |f| require f }
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,8 @@ Rails.application.routes.draw do
   unless Rails.env.production?
     get '/pages/*id' => 'pages#show', as: :page, format: false
   end
+
+  mount(StubAuthEndpoint.new => :test_auth_endpoint) if Rails.env.test?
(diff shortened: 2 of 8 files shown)


ndushay:
So what would other alternatives be?  Here's what I can think of:

* mocking the stacks calls and what is returned using rspec mocks / webmock 
* VCR cassettes for the stacks calls

other options?

jkeck:
Yeah, that's pretty much the alternative, and the dependencies this is attempting to avoid introducing.

ndushay:
I'm wondering if we can do mocks reasonably well with less complexity.  Maybe i'll give it a shot in a separate branch.

ndushay:
hahaha - ok, my short experiment was a colossal failure, as @jkeck no doubt expected.

jmartin-sul:
i like this.

is this better than than webmock?  neither approach seems clearly better than the other to me at first glance.  i'd be happy to merge this, if only to see how its maintainability compares to webmock's as development continues.

but this approach and webmock are both greatly preferable to VCR, IMO.  definite 👍 for a non-vcr solution.

ndushay:
@jmartin-sul webmock won't work, as I empirically discovered, because the request is coming from javascript, not from ruby.

jmartin-sul:
ah, yeah, of course.

PR #708, opened 2016-09-15
Remove reference to .map file that does not exist in videoJS.

jkeck:
Talked with @mejackreed about this and he has done something similar w/ the OSD source as well.


ndushay:
LGTM 👍

PR #709, opened 2016-09-16
Remove reference to .map file that does not exist in videoJS.

jkeck:


tingulfsen:
👍  from me - this seems like a no-brainer.

PR #710, opened 2016-09-16
Omit access restriction overlay for regular (non-media) thumbnails.

### LINKED ISSUE #707: Thumbnails still show access restriction message even after logging in
Consider https://purl-test.stanford.edu/bb582xs1304 - this object is access restricted to Stanford. Even after logging in, the thumbnails still show the access restriction message (although the actual image is also shown):

screen shot 2016-09-15 at 10 56 30 am

This is not a total dealbreaker, since you still get to see the thumbnail itself.

tingulfsen:
Fixes #707 

Currently, we display access restriction messages for all thumbnails. I think we only need to display them for actual videos and audio files. I believe the regular thumbnails that are mostly pictures of the media itself (cassettes, disks etc.) are in the open domain and can be viewed by anyone at any time regardless of restrictions.

Before:

screen shot 2016-09-16 at 12 30 07 pm

After:

screen shot 2016-09-16 at 12 27 09 pm

The access restriction message is still the same for playable media.

### DIFF
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -40,6 +40,7 @@ module Embed
       file_thumb = stacks_square_url(@purl_document.druid, file.thumbnail, size: '75') if file.thumbnail
       media_wrapper(thumbnail: file_thumb, file: file) do
         <<-HTML.strip_heredoc
+          #{access_restricted_overlay(file.try(:stanford_only?), file.try(:location_restricted?))}
           <#{type}
             id="sul-embed-media-#{file_index}"
             data-src="#{streaming_url_for(file, :dash)}"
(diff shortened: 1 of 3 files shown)


lib/embed/media_tag.rb
```diff
@@ -75,7 +75,7 @@ def media_wrapper(thumbnail: '', file: nil, &block)
              data-thumbnail-url="#{thumbnail}"
              data-duration="#{file.try(:duration)}">
           <div class='sul-embed-media-wrapper'>
-            #{access_restricted_overlay(file.try(:stanford_only?), file.try(:location_restricted?))}
+            #{access_restricted_overlay(file.try(:stanford_only?), file.try(:location_restricted?)) unless previewable_only}
```
jkeck: What if, instead of passing `preview_only` around, we only make the call to `access_restricted_overlay` above the `audio/video` markup in `media_element` (between lines 42 and 43 of this file)?

jmartin-sul: i like that suggestion

Issue #706, opened 2016-09-13
embed: okcomputer to replace is_it_working


ndushay:
Exemplar:  purl-fetcher

possibly:
- `/status` endpoint for load-balancer needs and nagios errors
- `/status/[some-dependency]` for nagios warnings;  nagios routes them to approp parties
- `/status/all-dependencies` for full dependency documentation (?)

after this
- [x] PR merged to master (#711)
- [x] deploy
  - [x] dev
  - [x] stage
  - [x] uat
  - [x] prod-a, prod-b
- [x] update Load Balancer (SAR-3261)
- [x] update nagios 
  - [x] basic checks
    - [x] dev (sul-dlss/puppet/pull/1009)
    - [x] stage (sul-dlss/puppet/pull/1008)
    - [x] uat (sul-dlss/puppet#1046)
    - [x] prod-a, prod-b (sul-dlss/puppet/pull/1065)
    - [x] embed-lb (sul-dlss/puppet/pull/1065)
  - [ ] any individual checks need separate nagios handling? (media? geo? image? web-archive-seed?)
- [ ] remove is_it_working
  - [ ] deploy
- [x] update DevOpsDocs (new nagios, load balancer ...)  (sul-dlss/DevOpsDocs#67)

PR #711, opened 2016-09-20
okcomputer app monitoring at /status

ndushay:
connected to #706

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -1,0 +1,38 @@
+require 'okcomputer'
+
+# /status for 'upness' (Rails app is responding), e.g. for load balancer
+# /status/all to show all dependencies
+# /status/<name-of-check> for a specific check (e.g. for nagios warning)
+OkComputer.mount_at = 'status'
+OkComputer.check_in_parallel = true
+OkComputer::Registry.deregister 'database' # don't check (unused) ActiveRecord database conn
+
+# REQUIRED checks, required to pass for /status/all
+#  individual checks also avail at /status/<name-of-check>
+OkComputer::Registry.register 'ruby_version', OkComputer::RubyVersionCheck.new
+# TODO: add app version check when okcomputer works with cap 3 (see http://github.com/sportngin/okcomputer#112)
+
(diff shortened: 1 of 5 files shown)


config/settings.yml
```diff
@@ -5,7 +5,7 @@ purl_conn_timeout: 2
 purl_feedback_email: ''
 valid_purl_url: <%= /https?:\/\/purl\.stanford\.edu\/*/ %>
 stacks_url: 'https://stacks.stanford.edu'
-iiif_stacks_url: 'https://stacks.stanford.edu'
+iiif_stacks_url: 'https://stacks.stanford.edu' # unused???
```
tingulfsen: I agree that it's good to call out unused code, but maybe we can work together to figure out if we can delete this one?

jkeck: Yeah, it doesn't look like it's used anymore (I'm guessing it did at one point before we swapped out the image viewer).

I say kill it.

ndushay: done.   This also made me think:  we're starting to get better at documenting why we include gems in a Gemfile.  Should we not be better about documenting un-obvious things in a settings file?   Like "valid_purl_url" and "embed_iframe_url" -- those aren't obvious to me.  (might the latter be better named "iframe_url_prefix" or something?)  If you agree, maybe we should 
* make a ticket for this?
* watch for this on PRs in the future

config/initializers/okcomputer.rb
```diff
@@ -0,0 +1,38 @@
+require 'okcomputer'
+
+# /status for 'upness' (Rails app is responding), e.g. for load balancer
+# /status/all to show all dependencies
+# /status/<name-of-check> for a specific check (e.g. for nagios warning)
+OkComputer.mount_at = 'status'
+OkComputer.check_in_parallel = true
+OkComputer::Registry.deregister "database" # don't check (unused) ActiveRecord database conn
```
jkeck: Are you dead-set against having a consistent style for quotes in this file?

config/settings.yml
jkeck: 👍 I'm not sure we need a ticket officially for this.  But I do like being more careful to document these things in the future (similarly to how we do for the Gemfile).

tingulfsen:
I'm 👍  on this now.

PR #726, opened 2016-09-30
Add a fullheight flag as an embed option used only by the...

### LINKED ISSUE #718: related to #716 getting embed to fill screen
Hi, related to  issue #716, I didn't initially notice it (wrong shape image) but when you use the embed with an undefined maxwidth and  maxheight, it no longer fills the screen:

screen shot 2016-09-22 at 7 19 30 pm

The old behavior, when you could specify a maxheight=100% and maxwidth=100% would make the image in the embed almost as wide as the screen and would extend the height to the screen height, vertically centering (if I remember) the image.

thanks

jkeck:
...image viewer to support a legacy use-case.

Closes #718 

## Desktop
desktop

## iOS
ios


app/assets/stylesheets/image_x.scss
```diff
@@ -14,6 +14,23 @@ $image-thumb-slider-background-color: $sul-background;
...
+// This is used for the fullheight option
+// This gives the image viewer enough
+// space given the header/footer
+.#{$namespace}-fullheight {
+  .#{$namespace}-image-x {
+    height: calc(100vh - 120px);
+  }
+
+  // When in the fullheight mode we
+  // want to hide the full-screen button
```
tingulfsen: What you're doing here seems fair enough to me, but is this tested anywhere? I assume it would be a fairly simple spec to check that the button is indeed hidden from view.

spec/lib/embed/viewer/common_viewer_spec.rb
```diff
@@ -104,6 +104,22 @@
       html = Capybara.string(file_viewer.to_html)
       expect(html).to have_css '.sul-embed-container[style="display:none; max-height:200px; max-width:200px;"]', visible: false
     end
+
+    context 'fullheight' do
+      it 'includes the sul-embed-fullheight class if the fullheight option is passed' do
+        stub_purl_response_and_request(file_purl, request)
+        expect(request).to receive(:fullheight?).at_least(:once).and_return(true)
+        html = Capybara.string(file_viewer.to_html)
+        expect(html).to have_css('.sul-embed-fullheight', visible: false)
```
tingulfsen: Can you remind me again why visible: false is necessary here?

jkeck: The viewer is `display:none` by default and is displayed via JS once the JS and CSS is loaded.

Since we're using `Capybara.string` above we're really just using the capybara DSL to make assertions against the HTML (not executing any javascript).  That means that the element we're checking is `display:none` so we have to say `visible: false` otherwise Capybara will say the element does not exist (which it does).

app/assets/stylesheets/image_x.scss
jkeck: I'll see if there is a good way to test this.

ndushay:
LGTM - can you deploy branch to uat to see if Rumsey room use case is satisfied?

jkeck:
I've deployed this to all envs. besides production.

spec/rails_helper.rb
```diff
@@ -63,15 +63,16 @@ def fill_in_default_sandbox_form(druid = 'ab123cd4567')
...
     url: "#{Settings.purl_url}/#{druid}",
     min_files_to_search: min_files_to_search,
     hide_search: hide_search,
     hide_title: hide_title,
-    hide_download: hide_download
+    hide_download: hide_download,
+    fullheight: fullheight
   )
-  expect(page).to have_css('.sul-embed-container')
+  # expect(page).to have_css('.sul-embed-container')
```
tingulfsen: Is there any reason for leaving this commented out?

jkeck: Good :eye:! I forgot to uncomment that from some debugging I was doing.

tingulfsen:
After the recent changes I am 👍 on this.

PR #727, opened 2016-10-04
Recognize image/jpeg as a mimetype for images.

### LINKED ISSUE #723: Issue with handling of file-level thumbs.
Example: xy943wj0513

There should only be two thumbnails in the index but the second file in the `Tape 1, Side A` resource is an `.jpg` and not a `.jp2`.  (do we get IIIF responses for non JP2 images?)

tingulfsen:
Fixes #723

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -210,9 +210,8 @@ module Embed
           preview_types.include?(mimetype)
         end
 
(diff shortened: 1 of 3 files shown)


tingulfsen:
It turns out that this is a blind alley, and not really the solution that is required for this problem. I'm closing the PR without merging.

PR #729, opened 2016-10-06
Add methods to properly identify what the primary file is for a…

### LINKED ISSUE #723: Issue with handling of file-level thumbs.

jkeck:
…resource.
Closes #723 

This will allow us to single out the primary file for a resource, the thumbnail for a resource,
and by extension any secondary files for a resource like transcripts or captioning.

## Before (xy943wj0513)
xy943wj0513-before

## After (xy943wj0513)
xy943wj0513-after

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -172,8 +172,12 @@ module Embed
         end
       end
 
-      def non_thumbnail_files
-        files.reject(&:thumbnail?)
+      def primary_file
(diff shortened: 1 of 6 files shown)


ndushay:
LGTM;  👍

PR #732, opened 2016-10-27
media viewer: only show downloadable files in download panel

### LINKED ISSUE #728: Unlink no-download files in the download panel
Example: https://purl.stanford.edu/zr649jg5679

ndushay:
closes #728 

Assumptions: 
- if a file is location restricted for read access, it is NOT downloadable.
- this is for the media viewer only

### Before (zr649jg5679)

zg649jg5679 download before
### After (zr649jg5679)

zg649jg5679 download after

### Before (nm201qr8213)

nm201qr8213 before

### After (nm201qr8213)

nm201qr8213 after

### DIFF
--- a/lib/embed/viewer/media.rb
+++ b/lib/embed/viewer/media.rb
@@ -16,7 +16,7 @@ module Embed
       end
 
       def download_html
-        return '' if @request.hide_download?
+        return '' unless show_download?
         Embed::DownloadPanel.new do
           <<-HTML.strip_heredoc
             <div class='sul-embed-panel-body'>
@@ -39,6 +39,11 @@ module Embed
         [:media]
       end
 
+      # override CommonViewer instance method to ensure we do not show download panel when no downloadable files
+      def show_download?
+        super && @purl_object.downloadable_files.present?
+      end
+
       def self.show_download?
         true
       end
@@ -50,18 +55,16 @@ module Embed
       end
 
       def download_list_items
-        @purl_object.contents.map do |resource|
-          resource.files.map do |file|
-            file_size = "(#{pretty_filesize(file.size)})" if file.size
-            <<-HTML.strip_heredoc
-              <li>
-                <a href='#{file_url(file.title)}' title='#{file.title}' target='_blank' rel='noopener noreferrer'>Download #{file.label}</a>
-                #{restrictions_text_for_file(file)}
-                #{file_size}
-              </li>
-            HTML
-          end
-        end.flatten.join
+        @purl_object.downloadable_files.map do |file|
+          file_size = "(#{pretty_filesize(file.size)})" if file.size
+          <<-HTML.strip_heredoc
+            <li>
+              <a href='#{file_url(file.title)}' title='#{file.title}' target='_blank' rel='noopener noreferrer'>Download #{file.label}</a>
+              #{restrictions_text_for_file(file)}
+              #{file_size}
+            </li>
+          HTML
+        end.join
       end
 
       def media_accessibility_note
--- a/spec/features/media_viewer_spec.rb
+++ b/spec/features/media_viewer_spec.rb
@@ -71,7 +71,7 @@ describe 'media viewer', js: true do
       expect(page).to have_css('[data-slider-object="1"] video', visible: false)
 
       within('.sul-embed-thumb-slider') do
-        expect(page).to have_css('.sul-embed-slider-thumb', count: 2)
+        expect(page).to have_css('.sul-embed-slider-thumb', count: 3)
       end
     end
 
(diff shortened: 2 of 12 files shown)


jvine:
Wait, can the authed user in fact no longer download the thumbnail that is available to the non-authed user?

In prod, for qb158yg5063, non-authed - displays thumbnail, which user can save from browser, but download list is empty. 

screen shot 2016-10-27 at 11 36 49 am

Authed - image is in zoom mode, so no longer saveable as a thumbnail; download list includes "Download Thumbnail". 

screen shot 2016-10-27 at 11 37 23 am

In the **after** state shown in the PR, there is no download thumbnail option. Is this intentional? I don't understand.

ndushay:
@jvine yeah I noticed that for this example that the download thumbnail went away.  That wasn't my intention.  I have to investigate if this is an oddity of the image viewer or what.  

So we want to keep allowing download of the thumbnail, regardless of rights? 
```xml
<rightsMetadata objectId="druid:cg616xd9084">
  <access type="discover">
    <machine>
      <world/>
    </machine>
  </access>
  <access type="read">
    <machine>
      <group rule="no-download">stanford</group>
    </machine>
  </access>
```

jvine:
If they can see the thumbnail, they can de facto download the thumbnail, so we should be explicit about that, IMO.

ndushay:
Note that had F2F conversations w @jvine and she is okay with
1.  not changing any viewers except media viewer for this
2.  thumbnail images of non-image data NOT being explicitly included in the download panel.  (e.g. nm201qr8213 above).

I have checked that image, file and geo viewers do not have any changes to their download panel behavior with this PR.

lib/embed/viewer/media.rb
```diff
@@ -39,6 +39,13 @@ def self.supported_types
         [:media]
       end
 
+      # override CommonViewer instance method to ensure we do not show download panel when no downloadable files
+      def show_download?
+        show_it = method(:show_download?).super_method.call
+        return @purl_object.downloadable_files.present? if @purl_object && show_it
+        show_it
```
jkeck: So, I admit I don't quite understand what we're trying to accomplish here.

Can line 44 just be `show_it = super`?

On line 45, I'm not sure we will ever have the context of the `@purl_object` not existing.  That gets set in the initializer for this class, so I don't know if we need to check that here.

spec/features/media_viewer_spec.rb
```diff
@@ -71,7 +71,7 @@
       expect(page).to have_css('[data-slider-object="1"] video', visible: false)
 
       within('.sul-embed-thumb-slider') do
-        expect(page).to have_css('.sul-embed-slider-thumb', count: 2)
+        expect(page).to have_css('.sul-embed-slider-thumb', count: 3)
```
jkeck: Just out of curiosity, why is this number dropping here?  What change is removing items from the thumb-slider?

spec/lib/embed/media_tag_spec.rb
```diff
@@ -30,11 +30,11 @@
     end
 
     it 'includes a data-src attribute for the dash player' do
-      expect(subject).to have_css('[data-src]', count: 2, visible: false)
+      expect(subject).to have_css('[data-src]', count: 3, visible: false)
     end
 
     it 'includes a data attribute that includes the url to check the users auth status' do
-      expect(subject).to have_css('video[data-auth-url]', count: 2, visible: false)
+      expect(subject).to have_css('video[data-auth-url]', count: 3, visible: false)
```
jkeck: Similar question about these dropping, even if an object was no-download, it shouldn't remove the number of video objects we're getting. (although perhaps I'm miss-understanding).

ndushay: One of the video objects in the original fixture was location restricted, so effectively no-download.  That resulted in some other tests breaking.  I felt the easiest way to make the fixture work for the various tests was to add another file to the fixture that was NOT download restricted.  Hence this change.

jkeck: Ah.  I was miss-reading the diff.  I see now that it's increasing the number which is explained by the fixture change.

Apologies.

spec/features/media_viewer_spec.rb
ndushay: ditto to comment below - this was a side effect of changing that fixture object.  Note that the number is increased by 1, not dropping, so nothing is being removed from thumb slider -- the additional file in the fixture is being added to the thumb slider.

ndushay:
@jkeck thanks comments - was able to simplify the method considerably!   W00t!   Removing WIP so this is ready to merge if you want.

lib/embed/viewer/media.rb
ndushay: whew - thanks for this - was able to simplify method and make it clearer.

jkeck:
Okay, one final comment.  I swear. 😁

ndushay:
@jkeck updated for your last comment - thanks for spotting that simplification (the possibility was introduced with the previous refactor).   I think this is really and truly ready for merge now.

PR #740, opened 2017-03-14
Add m4a extension.

tingulfsen:
This PR implements support for the AAC audio format on the Mac platform. I've tested this on the various Mac browsers, my Android phone and in IE and the latest Edge for Windows 10, but the testing is not entirely complete yet. Nevertheless, I think this is a good step forward for supporting the new format.

### DIFF
--- a/lib/embed/stacks_media_stream.rb
+++ b/lib/embed/stacks_media_stream.rb
@@ -44,11 +44,7 @@ module Embed
     end
 
     def streaming_file_prefix
-      if video?
-        'mp4'
-      elsif audio?
-        'mp3'
-      end
+      Settings.streaming_prefix[file_extension]
     end
 
     def video?
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -56,3 +56,11 @@ stream:
     - mpeg
   audio:
     - mp3
+    - m4a
+# Mapping from file extensions to Wowza streaming prefixes
+streaming_prefix:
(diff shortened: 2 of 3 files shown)


lib/embed/stacks_media_stream.rb
```diff
@@ -47,7 +47,8 @@ def streaming_file_prefix
       if video?
         'mp4'
       elsif audio?
-        'mp3'
+        return 'mp3' if file_extension == 'mp3'
+        return 'mp4' if file_extension == 'm4a'
```
jkeck: This feels a bit unsatisfactory to me.  The entire method was predicated on the fact that audio always is `mp3` and video is always `mp4`.  I wonder if should rethink this method entirely now.  I don't have a good idea off the top of my head what that might look like, but perhaps a config hash that can have a default file prefix and then can configure variations from that default (e.g. `m4a` gets `mp4` instead of the default `mp3`).

That being said, this is a small enough change that if it felt like setting that up was more trouble that it was worth, we can wait until we come across the next incompatible file prefix to figure it out.  Totally up to you.

jmartin-sul: i hit approve, but this is also a reasonable point.  agreed on leaving it up to @tingulfsen and maybe waiting for another instance in case an obvious pattern emerges (though config hash seems like a sensible idea).

PR #756, opened 2017-06-03
Few minor tweaks to rails 5 upgrade

jkeck:


jkeck:
Ugh, looks like this is a ruby 2.3 vs 2.4 thing.  Oh well.

PR #759, opened 2017-06-07
Use uv for images

cbeer:


config/settings.yml
```diff
@@ -64,3 +64,4 @@ streaming_prefix:
   mpeg: 'mp4'
   m4a: 'mp4'
   mp3: 'mp3'
+image_viewer: Embed::Viewer::ImageX
```
ndushay: @mejackreed  could you please add a comment and indicate what the two options are?

.hound.yml
```diff
@@ -1,9 +1,9 @@
-javascript:
+jshint:
```
ndushay: this is awesome if it will shut up a lot of the hound messages.

mejackreed: 🙏

config/settings.yml
mejackreed: Added.. thanks for that. Great catch

PR #760, opened 2017-06-14
Update Universal Viewer directory name to reflect the version we want…

tingulfsen:
We want to use the UV version 3 branch, so update the directory name.


ndushay:
if we want/need this, the tests need to pass ...

PR #769, opened 2017-06-30
Hint to the browser that download links should trigger a download action

cbeer:
I suspect this doesn't work and/or isn't cross-browser compatible, but it might improve the download experience for some users. In any case, it aligns the geo and media viewers with what other download panes are doing.


mejackreed:
👍

jkeck:
Feel free to :shipit: as you need.

PR #775, opened 2017-07-07
[WIP] Swap webpacker for bower-rails

jcoyne:
webpacker is an official rails project and I anticipate it'll get more
support than bower-rails in the future.

I'm having a curious problem now where:
```
 LayoutStore is not defined
```
This is occuring because `ImageXViewer` is depending on this as a global.


jcoyne:
@aeschylus @mejackreed do you know anything about this `LayoutStore is not defined`? Should I be using https://github.com/webpack/docs/wiki/list-of-plugins#provideplugin ?

jcoyne:
@aeschylus @mejackreed alternatively, is it okay to refactor LayoutStore to `export default LayoutStore = function(options) { ... }`?

csbailey5t:
@jcoyne In the new standard JS way, `export default LayoutStore` should work for when you import that file with a generic import, such as `import 'modules/layout_store';` as long as that is the only function you want to export from the file. You could also try a partial import, such as `import LayoutStore from 'modules/layout_store';`.

mejackreed:
Yeah @jcoyne i think that will be fine

mejackreed:
One thing we haven't addressed here yet is the deployment story. Just adding a comment here as a reminder to test that.

jkeck:
This has been completed in another PR.

PR #777, opened 2017-07-07
Switch to using erb views

jcoyne:
This leverages view templates, which is where Rails typically constructs HTML.

This also gets rid of the stray `<!DOCTYPE` declarations scattered through the views.


app/controllers/embed_controller.rb
```diff
@@ -1,4 +1,5 @@
 class EmbedController < ApplicationController
+  append_view_path Rails.root.join('app', 'views', 'embed')
```
cbeer: Is this because of the singular name?

app/controllers/embed_controller.rb
```diff
@@ -12,7 +13,10 @@ def get
   end
 
   def iframe
-    render html: iframe_html.html_safe
+    # Trigger purl object validation (will raise Embed::PURL::ResourceNotAvailable)
```
cbeer: Ew. I know this isn't you, but a `?` method raising an exception is icky.

jcoyne: @cbeer Do you really want that fixed in this PR?

app/controllers/embed_controller.rb
jcoyne: Yes.

app/views/embed/body/_file.html.erb
```diff
@@ -0,0 +1,83 @@
+<div class='sul-embed-body sul-embed-file'
+      data-sul-embed-theme="<%= viewer.asset_url('file.css') %>"
```
mejackreed: Can we drop the `viewer` prepend here? I'm seeing some strange stuff locally when assets change.

mejackreed: Hmm.. maybe not? https://github.com/sul-dlss/sul-embed/blob/master/app/viewers/embed/viewer/common_viewer.rb#L15

jcoyne: Sounds good for another PR. There's already a lot of moving stuff in this one. I'm trying not to change more than necessary

PR #781, opened 2017-07-13
Hard code the default in the factory

jcoyne:
This makes the code simpler and easier to understand


mejackreed:
👍  on the mustache type

PR #786, opened 2017-07-13
Update openseadragon to 2.2.1

jcoyne:


mejackreed:
I think we want to kill the `map.js` file as it has caused us some problems in the past. https://github.com/sul-dlss/sul-embed/pull/336

Though not quite sure if that problem would have been resolved.

jcoyne:
@mejackreed I didn't see any problems with it on my box. Do you see any? Do you know what the nature of the problem was?

mejackreed:
I'm getting an error in the log:

```
http://localhost:3000/iframe?url=https://purl.stanford.edu/fw090jw3474&min_files_to_search=10 -> /assets/image_x-0392e4e42e8ebc34756824658ef583bf15d5da5bd50993971e080a080f73529c.js?_=1499970720870
Started GET "/pages/openseadragon.min.js.map" for 127.0.0.1 at 2017-07-13 14:32:00 -0400
Processing by PagesController#show as HTML
  Parameters: {"id"=>"openseadragon.min.js.map"}
Completed 404 Not Found in 8ms



ActionController::RoutingError (No such page: openseadragon.min.js.map):

```

jcoyne:
Okay, as far as I can tell, this only occurs when you are in Chrome and have the developer console open when you load the page.  Does that sound right?

jcoyne:
Here's an issue to track: https://github.com/rharriso/bower-rails/issues/213

jcoyne:
@mejackreed as far as I can tell, removing the map file does not make the problem go away.

mejackreed:
Ah thanks for catching that. We had previously also removed the reference from the `openseadragon.min.js` file too.

PR #791, opened 2017-07-20
remove margins from panels and attribution

mejackreed:
Closes https://github.com/sul-dlss/universalviewer/issues/26

screen shot 2017-07-20 at 2 31 17 pm


mejackreed:
cc @ggeisler @jvine

PR #811, opened 2017-08-22
change current view to zoomed region text

### LINKED ISSUE #807: Download link text is confusing
Testing : https://sul-purl-uat.stanford.edu/bc421tk1152 in Chrome on Mac

The download link text is confusing.  Maybe "Current View" can be changed to something like "Zoomed region only" or something.  Was also a little confused by two links that said "Whole Image" but with different sizes.  

Interested in @jvine and @ggeisler thoughts on this.  

image

mejackreed:
Closes #807 

cc @ggeisler 

screen shot 2017-08-22 at 5 11 04 pm

### DIFF
--- a/public/uv-config.json
+++ b/public/uv-config.json
@@ -25,6 +25,11 @@
                 "printEnabled": false,
                 "shareEnabled": true
             }
+        },
+        "downloadDialogue": {
+          "content": {
+            "currentViewAsJpg": "Zoomed region {0} x {1}px (jpg)"
+          }
         }
     }
 }


ggeisler:
Any objection to adding a colon after "Zoomed region" and "Whole image", as in my mockup in #807?

mejackreed:
Missed that.. @ggeisler I'll add those in. Thanks for catching that.

mejackreed:
Updated cc @ggeisler 

screen shot 2017-08-22 at 7 07 45 pm

mejackreed:
And more options here: https://github.com/UniversalViewer/universalviewer/blob/v3/src/extensions/uv-seadragon-extension/l10n/en-GB.json#L48-L67

ggeisler:
Thanks @mejackreed. Not sure if I'm misunderstanding something but your updated screenshot only shows the colon after "Zoomed region". I think we'd want it after all labels, or none of them.

mejackreed:
Akk. you're right! Looks like UV isn't respecting that config. I've reverted back to no colon and we can come back to this later.

PR #814, opened 2017-08-24
handle parsing of multiple extents by just using the first one

### LINKED ISSUE #813: Latlnt parsing is broken on some objects
### Expected behavior
bounds should be shown

### Actual behavior
javascript error
gk628xr0233

mejackreed:
Closes #813

### DIFF
--- a/app/models/embed/envelope.rb
+++ b/app/models/embed/envelope.rb
@@ -17,11 +17,11 @@ module Embed
     private
 
     def lower_corner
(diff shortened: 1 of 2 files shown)


jcoyne:
Should this have a test?

mejackreed:
Yeah.. So updating the fixture there gives us a breaking test. Want me to pop that out into something else?

jcoyne:
Or, maybe this is already being tested and the fixture change triggers the error?

Issue #837, opened 2017-10-10
Add automated integration tests for UV


mejackreed:
This is a technical debt ticket that needs to address automating UV testing for specific scenarios

mejackreed:
No longer relevant

PR #842, opened 2017-11-28
enables canvas index to be passed through to UV

### LINKED ISSUE #313: Allow deep-linking into the embedded content
One feature in the PURL gallery experience is the ability to deep-link to a specific image within an object (although it's highly imperfect). The embed widget should be able to handle deep-linking (and provide some way to expose that back up to the embedding application, using e.g. replaceState?) too.

mejackreed:
This PR, closes #313 and required for https://github.com/sul-dlss/exhibits/issues/827 enables a user who knows the api, to link to a specific canvas by zero-based canvas index order. UV does not yet support linking to a canvas by canvasID.

canvaslink

### DIFF
--- a/app/assets/javascripts/modules/uv_viewer.js
+++ b/app/assets/javascripts/modules/uv_viewer.js
@@ -12,7 +12,8 @@
         iiifResourceUri: options.uvUri,
         configUri: options.uvConfig,
         root: options.uvRoot,
-        deepLinkingEnabled: false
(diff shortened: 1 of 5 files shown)


mejackreed:
Yup we should def do that. I created a tech debt ticket a while back about this: https://github.com/sul-dlss/sul-embed/issues/837

But at the moment, we have zero UV integration tests :the_saddest_kitty_ever:

PR #844, opened 2018-01-09
[WIP] Allow an incoming 'q' parameter to initialize UV with that search term.

jkeck:
_**[WIP]** until we can update to a version of UV that doesn't require deep linking in order to initialize with a query term._

See UniversalViewer/universalviewer#527


jkeck:
This will be handled by M3 and we are starting to develop code to do that now.

PR #847, opened 2018-01-30
UV provides its own header and footer

cbeer:


mejackreed:
Were you seeing issues in the viewer without this?

cbeer:
Yes -- without this, the bottom 30 pixels that should be used by the viewer were blank, making it awkward to do https://github.com/sul-dlss/exhibits/issues/1063 (the iframe is 600px, the viewer was only 570px)

mejackreed:
Nice, great catch. Just was wondering :)

cbeer:
screen shot 2018-01-30 at 16 18 41

PR #848, opened 2018-01-30
Remove Image X viewer and related, unused files

cbeer:


aeschylus:
LGTM. I don't see anything that isn't used exclusively by the ImageX viewer or iiif-osd-viewer.

PR #935, opened 2018-11-26
Add feature inspector for geo

### LINKED ISSUE #651: Geoblacklight/Earthworks Map/Attrib Preview Window should be an oEmbed.
I'd like to see the map previews and attribute inspection in geoblacklight/Earthworks moved into an oembed, so that the purl page gets both, as well as providing the ability to embed the previews into other pages (Spotlight, Blacklight, blogs, etc...).

mejackreed:
Fixes #651 

Adds feature inspection for public stanford geo content

featureinspection

## Update
featureinspection

PURL: `cz128vq0535`


mejackreed:
I'm not sure we have much choice on the Hound scss complaints. We need to do a few of these things to override correctly. Open to suggestions on this.

mejackreed:
Thanks for that @camillevilla . I've updated the PR to hopefully avoid some of these things. Updates include: disabling max selector depth, and refactoring vendor prefixes to a mixin.

PR #1001, opened 2019-05-31
Fixes #992 by updating settings

### LINKED ISSUE #992: Configure default settings for Mirador 3 viewer
Below are the configuration settings to use for the Mirador 3 viewer that are different from the default settings. 

`window` settings

```
window: {
    allowClose: false, 

    defaultSideBarPanel: 'attribution', 

    sideBarOpenByDefault: true, 
  },
```

`workspace` settings

```
showZoomControls: true,
```

`workspaceControlPanel` settings

```
enabled: false,
```

Those settings should give us something like this on initial load of the viewer:

Screen Shot 2019-05-17 at 11 37 43 AM

Note that we might want to revisit the default side panel settings (open to the Rights panel) with Hannah, given the new viewer, to see whether she still thinks we need to have the Right info visible on initial load of the viewer.

Also, I created a [Mirador 3 issue](https://github.com/ProjectMirador/mirador/issues/2623) to add a configuration setting to enable hiding the window title in the title bar. Assuming that is implemented, I think we'd also want to set the title display off:

```
window: {
   hideWindowTitle: true, 
}
```

mejackreed:
When https://github.com/ProjectMirador/mirador/pull/2641 is shipped in a release, the fullscreen button should be added.

Screen Shot 2019-05-31 at 4 09 56 PM

### DIFF
--- a/app/assets/javascripts/modules/m3_viewer.js
+++ b/app/assets/javascripts/modules/m3_viewer.js
@@ -36,12 +36,15 @@
           ],
           window: {
             allowClose: false,
+            allowFullscreen: true,
             allowMaximize: false,
-            hideAnnotationsPanel: true,
+            defaultSideBarPanel: 'attribution',
+            hideWindowTitle: true,
             sideBarOpenByDefault: true
           },
           workspaceControlPanel: {
-            enabled: false
+            enabled: false,
+            showZoomControls: true
           }
         });
       },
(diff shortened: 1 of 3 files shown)


jkeck:
I don't believe that we have a release w/ ProjectMirador/mirador#2641 yet.  I'm planning on cutting one today though, so we should just avoid deploying this to uat until then.

cbeer:
Ah, sorry, my reading was "this is good to go, and when we cut a release the last thing'll fall into place".

jkeck:
No worries.  I had intended on cutting that release for the past 3 days so it's my fault. 😶

mejackreed:
> this is good to go, and when we cut a release the last thing'll fall into place

Yeah it should work like that

PR #1012, opened 2019-06-07
Add share plugin

mejackreed:
Starts integration of mirador-share-plugin into sul-embed.

There will be some console errors until https://github.com/ProjectMirador/mirador-share-plugin/pull/8 is shipped and integrated. (should be ok though)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,9 @@
   "dependencies": {
     "@rails/webpacker": "^4.0.7",
     "list.js": "1.1.1",
-    "mirador": "^3.0.0-alpha.9"
+    "mirador": "^3.0.0-alpha.9",
+    "react": "^16.8.6",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -16,7 +16,10 @@
   "dependencies": {
     "@rails/webpacker": "^4.0.7",
     "list.js": "1.1.1",
-    "mirador": "^3.0.0-alpha.9"
+    "mirador": "^3.0.0-alpha.9",
+    "mirador-share-plugin": "^0.1.0",
+    "react": "^16.8.6",
```
mejackreed: We have to add `react` and `react-dom` dependencies now that we are using the `es` version of the package. The `umd` has all of this baked into it (which we don't want).

PR #1013, opened 2019-06-07
Add download plugin and update mirador/share

mejackreed:


app/assets/javascripts/modules/m3_viewer.js
```diff
@@ -1,6 +1,9 @@
 import Mirador from 'mirador/dist/es/src/index.js';
 import miradorShareDialogPlugin from 'mirador-share-plugin/es/MiradorShareDialog.js';
 import miradorSharePlugin from 'mirador-share-plugin/es/miradorSharePlugin.js';
+import miradorDownloadPlugin from 'mirador-dl-plugin/es/miradorDownloadPlugin.js';
+import miradorDownloadDialogPlugin from 'mirador-dl-plugin/es/MiradorDownloadDialog';
```
jkeck: Will these need a `.js` extension like the other files?

mejackreed: doesn't I guess, but just pushed the change for consistency

PR #1026, opened 2019-06-19
Change up default panel logic

### LINKED ISSUE #1024: Update viewer configuration/instantiation related to Attribution panel
Based on conversations with @caaster and @mejackreed it seems the requirement for users to see the attribution information applies only to the Warhol collection (possibly Bassi, but it is okay to assume not unless and until we hear from them, and for sure not FRDA), and Jack thinks it is possible to deal with objects from a collection as a special case when showing the viewer. This seems to open up a new approach where we can:

- [x] Change the default configuration of the viewer to show no open panels.

- [x] If we know we are loading an object from Warhol (https://purl.stanford.edu/dc876nn4469), open the Attribution panel when the viewer is instantiated.

- [x] If we are dealing with a request via the "prepared search link" in #1022, open the Search panel when the viewer is instantiated, as described in that ticket.

@mejackreed Please correct me if I misinterpreted your thoughts.

@jvine Let us know if any of this doesn't seem like a good idea to you.

mejackreed:
Part of #1024 

### Object in Warhol http://127.0.0.1:3000/iframe?url=https://purl.stanford.edu/gm191tg1688
Screen Shot 2019-06-19 at 7 27 05 AM

### Object not in Warhol http://127.0.0.1:3000/iframe?url=https://purl.stanford.edu/fw090jw3474
Screen Shot 2019-06-19 at 7 27 14 AM

Object with search moved to `1024-search`

### DIFF
--- a/app/assets/javascripts/modules/m3_viewer.js
+++ b/app/assets/javascripts/modules/m3_viewer.js
@@ -60,11 +60,11 @@ export default class M3Viewer {
         allowFullscreen: true,
         allowMaximize: false,
         authNewWindowCenter: 'screen',
-        defaultSideBarPanel: 'attribution',
+        defaultSideBarPanel: (data.showAttribution === true) ? 'attribution' : 'info',
         hideAnnotationsPanel: true,
         hideSearchPanel: false,
         hideWindowTitle: (data.hideTitle === true),
-        sideBarOpenByDefault: true
+        sideBarOpenByDefault: (data.showAttribution === true)
       },
       workspaceControlPanel: {
         enabled: false,
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -108,6 +108,13 @@ module Embed
         end
     end
 
+    def collections
+      ng_xml.at_xpath(
+        '//fedora:isMemberOf',
+        'xmlns:fedora' => 'info:fedora/fedora-system:def/relations-external#'
+      )&.map { |_name, value| value.gsub('info:fedora/druid:', '') } || []
+    end
+
     private
 
     def cc_license
--- a/app/viewers/embed/viewer/m3_viewer.rb
+++ b/app/viewers/embed/viewer/m3_viewer.rb
@@ -34,6 +34,12 @@ module Embed
         0
       end
 
+      def show_attribution_panel?
+        purl_object.collections.any? do |druid|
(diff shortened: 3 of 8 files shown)


app/assets/javascripts/modules/m3_viewer.js
```diff
@@ -77,5 +87,30 @@ export default class M3Viewer {
...
       miradorDownloadPlugin,
     ]);
+    if (data.search) {
+      const state = mirador.store.getState();
+      const windowId = Mirador.selectors.getWindowIds(state)[0];
+      const companionWindowId = Mirador.selectors.getCompanionWindowIdsForPosition(state, { windowId: windowId, position: 'left' })[0];
+      let searchService = Mirador.selectors.getManifestSearchService(state, { windowId: windowId });
+      // Waits for 2 seconds for searchService to appear (manifest loaded and parsed)
+      let counter = 0;
+      const checkForSearchService = setInterval(() => {
```
mejackreed: Since state is async here.. we have to keep checking if the manifest comes in and is parsed, before we can get the search service. Perhaps we can provide a better API in Mirador? But this works for now.

jkeck: I'm wondering about possibly adding the ability to initiate a mirador viewer/window w/ a search initialized (since Content Search is in core).

UV has this feature and I think it would be beneficial to support something similar.  It would also allow us to avoid maintaining some of this code that is highly coupled to the Mirador API.

jkeck: I would be interested in exploring this in Mirador.  This is a feature UV has, and I think it would be beneficial for us to have it to.

jkeck: It seems like if this was in the Mirador API we could avoid any concern about having to set/clear intervals... which would be nice.

mejackreed: Totally agree! I could pop off the last commit if we want to get the other stuff in. We could then recircle to the mirador API stuff to handle search?

jkeck: I'll create a quick ticket over in Mirador.

mejackreed: K. I've done that. Draft PR of search stuff here: https://github.com/sul-dlss/sul-embed/pull/1027

PR #1037, opened 2019-06-26
Add polyfills and refactor the M3Viewer to use IE11-compatible code

cbeer:


mejackreed:
Thanks for looking at and fixing this. I started to look at https://babeljs.io/docs/en/babel-preset-env to do this automatically.. any luck with that in a try with this?

Issue #1038, opened 2019-06-26
Bust the browser caching on deploy


cbeer:

mejackreed:
Based off of this comment https://github.com/sul-dlss/sul-embed/pull/1039#issuecomment-506378975 do we still need this?

cbeer:
Yes, it'd be good to have.

PR #1039, opened 2019-06-26
Make node_modules and public/packs linked directories so that assets …

### LINKED ISSUE #1032: Investigate cached "packs" not being availabe
### Expected behavior
When caching is enabled, previous packs don't seem to be available on server.

E.g https://embed-uat.stanford.edu/packs/js/m3-6fa0de7205bb1210a010.js not available (previous cached version)
1. Open iframe version, to ensure cached js
2. Redeploy w/ js changes (new fingerprint)
3. Reopen w/ caching enabled in browser (should 404 js request)

jkeck:
…are not compiled unless they change and previous packs are available.

Closes #1032 (I took this for a spin in UAT, we should still do some sort of cache invalidation though so that users pick-up on the new viewer code).

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -21,7 +21,7 @@ set :log_level, :info
 set :linked_files, %w(config/honeybadger.yml config/newrelic.yml)
 
 # Default value for linked_dirs is []
-set :linked_dirs, %w(config/settings log tmp/pids tmp/cache tmp/sockets vendor/bundle public/system)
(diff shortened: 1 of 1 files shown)


cbeer:
And this is revertable once we figure out #1038 ?

jkeck:
I'm not sure we would want to revert it.  From what I understand, at least wrt `node_modules`, this will prevent compiling assets on every deploy when there have been no changed assets (which should help speed up deploy time)

PR #1052, opened 2019-07-03
Removes proptypes in production build.. saves 1KB over the wire

mejackreed:
Not sure if this helps much. but 1KB is 1KB /shrug


lwrubel:
Closing per discussion in purl + metrics WC discussion, not seeing a need for it.

PR #1054, opened 2019-07-03
PDF Viewer (using PDF.js)

jkeck:
pdf-viewer

## TODO
_Estimate 2 devs could complete w/i a day (potentially shorter if coupled w/ #1050 as some similar work will end up happening there)_
- [x] Tests
- [x] Use legit buttons
- [x] Update pdf viewer JS to match current patterns
- [x] Handle failure state
- [x]  Handle footer buttons/panels
  - [x] Download
  - [x] Embed This
  - [x] Full screen


mejackreed:
@jkeck one quick question I had was about the committing of the `pdf_worker.js` file. Is that file vendor generated file or is that something our build creates?

jkeck:
It's a vendor created file (generated and shipped w/ PDF.js)

mejackreed:
Is it something we can import/require into that file that so that we don't have to check it in? Or are we SOL on that.

jkeck:
Not that I'm aware of.  The worker needs to be available at a URL AFAICT.

camillevilla:
Nice work, @jkeck. I did notice that the focus styles for keyboard tab navigation don't seem to be working in Firefox (but work fine for Chrome and Safari). Is this already an known issue upstream? Or is there something about our styling that isn't being supported in Firefox? 

## Chrome
tabs-chrome

## Firefox
tabs-firefox

camillevilla:
Neeeeevermind. That's just a fundamental difference between Firefox and Chrome's default focus styles. 😭 The reason I noticed this is that UV augments overrides browser focus styles with its own. That's why you can still see the purple focus outline in Firefox on UV:

## UV
focus-styles

jvine:
Just some notes from poking at the focus indicator issue, no impact on this PR:  The Firefox focus issue is well-known, and it's "legal" to leave it as-is. Any explicit focus outline that would make Firefox's focus visible on black would also change the default behaviour for all other browsers. It's not possible to replicate Chrome's focus outline on other browsers. Related: it's not clear why we have overridden default focus behaviour on the embed frame.

anarchivist:
@jkeck A couple of lingering issues, which relate to `world (no download)` and `location: spec` objects:

* `location: spec`: bd742gh0511 - you just get a blank display, no error message
* `world (no-download)`: bd989qy8176, bf177xm2396

jkeck:
## Generic error (no-download) bd989qy8176
generic-no-download-bd989qy8176

## Location restricted error bd742gh0511
location-restricted-bd742gh0511

PR #1059, opened 2019-07-12
Add a box-shadow to the m3 viewer's container

### LINKED ISSUE #1055: Border on embedded viewer not displaying
Not sure if this was discussed while I was away but I'm not seeing a border around the viewer, although it appears there is a 1px border defined in the CSS:

Screen Shot 2019-07-05 at 1 40 43 PM

I think we need the border to display to give the viewer a clear visual definition, especially on the sides of the titlebar and the left side where the side panel menu just bleeds into the application background.

cbeer:
Fixes #1055 

Screen Shot 2019-07-11 at 21 30 47

I figured out what happened to our old shadow; we used to have all the mosaic cruft in the DOM and one of the styles we got from upstream was the `box-shadow` I've applied here.

### DIFF
--- a/app/assets/stylesheets/m3.scss
+++ b/app/assets/stylesheets/m3.scss
@@ -1,1 +1,8 @@
 @import 'common';
+
+.#{$namespace}-container {
+  border: 0;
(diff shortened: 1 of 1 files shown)


cbeer:
I'll ship to -stage and -uat.

PR #1076, opened 2019-09-06
Add honeybadger javascript error reporting

cbeer:


mejackreed:
Is this getting injected into the m3 viewer also?

cbeer:
I suppose it isn't. I guess there's a different way to do honeybadger with webpack.

jcoyne:
Fixed by #1580

PR #1121, opened 2020-03-12
Tell CodeClimate to use a newer version of Rubocop.

jkeck:


jkeck:
Not sure if CC uses the config from the branch it is testing or master. 😕

PR #1139, opened 2020-04-06
Update slider object navigation to use display none instead of…

jkeck:
…height/width/visibility updates.

### DIFF
--- a/app/assets/stylesheets/modules/thumb_slider.scss
+++ b/app/assets/stylesheets/modules/thumb_slider.scss
@@ -17,9 +17,7 @@ $thumb-slider-background-color: $sul-background;
 
 .#{$namespace}-hidden-slider-object,
 .#{$namespace}-hidden-slider-object * {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/modules/thumb_slider.scss
```diff
@@ -17,9 +17,7 @@ $thumb-slider-background-color: $sul-background;
 
 .#{$namespace}-hidden-slider-object,
 .#{$namespace}-hidden-slider-object * {
-  height: 0 !important;
-  visibility: hidden !important;
-  width: 0 !important;
+  display: none !important;
```
jkeck, line 20: I can't think of a particularly good reason to use height/width/visibility for this sort of display/show behavior (but totally will to entertain a good reason).  I thought at first it may have been using a transition but that didn't appear to be the case.  In general, this doesn't work super well w/ flexbox.

PR #1141, opened 2020-04-08
Don't hide/show video/audio elements…

jkeck:
… (they will be initialized by the time their parents are shown/hidden)

## Before
xk583vt6748-before

## After
xk583vt6748-after

### DIFF
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -49,7 +49,6 @@ module Embed
             controls='controls'
             aria-labelledby="access-restricted-message-div-#{file_index}"
             class="sul-embed-media-file #{'sul-embed-many-media' if many_primary_files?}"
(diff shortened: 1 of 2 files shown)


lib/embed/media_tag.rb
```diff
@@ -49,7 +49,6 @@ def media_element(file, type)
             controls='controls'
             aria-labelledby="access-restricted-message-div-#{file_index}"
             class="sul-embed-media-file #{'sul-embed-many-media' if many_primary_files?}"
-            style="display:none; height: 100%;"
```
jkeck, line 52: The `mediaObject.show()` above was causing some elements (different on chrome vs safari) to be given a `display: inline` instead of `display: block` and the content was not visible.  It doesn't appear that we should need to hide/show the video or audio elements as this is handled up the DOM.

PR #1142, opened 2020-04-10
Add our own check for stanford-only rights 

cbeer:
that considers an item stanford-only even if it is no-download

Believed to fix https://github.com/sul-dlss/media/issues/151

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -301,7 +301,9 @@ module Embed
         end
 
         def stanford_only?
-          @rights.stanford_only_unrestricted_file?(title)
+          value, _rule = @rights.stanford_only_rights_for_file(title)
+
+          value
         end
 
         def location_restricted?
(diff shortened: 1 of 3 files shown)


jkeck:
I _think_ sul-dlss/media#151 is different.  The issues is worded a bit oddly, but my interpretation from 
> Click the red link to go through auth and start playing. The video is able to play, but the red "login to play" link does not go away (it correctly does not show on page reload, or if already authed before getting to the page).

is that the issue being reported is that once the user has authenticated we're not dong anything to remove the login link from the viewer. 

If this change aims to address that then ignore (I had imagined a fix in the media js code for that).

cbeer:
I thought @hannahfrost confirmed that #151 was only an issue for no-download content. Now that the viewer will have the right initial state, I think things will work out right.

jkeck:
👍 I hadn't diagnosed the issue enough to determine that the initial state was the cause.  Happy for this to work out.

PR #1143, opened 2020-04-10
Add caption tracks

cbeer:


mjgiarlo:
@cbeer I believe we've got the ability to get transcription files for a PURL, and a way to generate tracks, in `main` now. OK to close?

PR #1147, opened 2020-04-17
Set a min height for the file viewer

### LINKED ISSUE #453: For short embed views (single file) panels are unusably short too
set min-height?

cbeer:
Fixes #453 by forcing the 1-file viewer to be no smaller than 2 files high, which is at least a little more usable?

Screen Shot 2020-04-17 at 08 34 36


cbeer:
If this seems like a reasonable thing to do, I'll figure out how to fix up that failing test in a meaningful way.

mjgiarlo (approved):

PR #1159, opened 2020-06-04
Remove import of unused translations

mejackreed:
No need to provide to browsers translations that they will never see. I wanted to be able to establish this pattern for others.

Before: 2.01 MB / 513.08 KB transferred
After: 1.96 MB / 501.26 KB transferred

2.3% reduction in JavaScript transferred

I've tested this on UAT. Currently deployed there.

### DIFF
--- a/app/javascript/empty_translation.js
+++ b/app/javascript/empty_translation.js
@@ -1,0 +1,1 @@
+{ translation: {}}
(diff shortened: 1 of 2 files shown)


cbeer:
I guess if/when we pass the locale from e.g. an exhibit to the embed, we can re-add these as needed?

PR #1162, opened 2020-06-18
Add image tools plugin to M3 on-demand

cbeer:


lib/embed/request.rb
```diff
@@ -70,7 +70,7 @@ def as_url_params
         :hide_title, :hide_embed, :hide_search, :hide_download,
         :min_files_to_search,
         :canvas_id, :canvas_index,
-        :search, :suggested_search
+        :search, :image_tools
```
jkeck, line 73: It looks like `suggested_search` was dropped here.

cbeer, line 73:
```suggestion
        :search, :suggested_search,
        :image_tools
```

cbeer: Just.. uh.. testing you.

PR #1165, opened 2020-06-23
Add a plugin to re-enable comparison views

cbeer:
Hopefully this is mostly temporary until an upstream approach is available. I don't even know if we'd  want to  merge this, or just deploy it out  to  -uat for Parker evaluation and  hope  something upstream comes along soon (https://github.com/ProjectMirador/mirador/pull/3156, to be released in an rc2 some day?)

TODO:
- [x] Mirador 3 rc2
- [x] update image tools to use `getWindowConfig` selector
- [x] new release of image tools
-


mejackreed:
Taking a look at this, I would be fine with shipping this or something similar very similar in sul-embed w/out the feature flag. It essentially "re-enables" previous behavior (DnD from UV) so it seems fine.

PR #1166, opened 2020-06-23
Bust the iframe cache every deploy

cbeer:


lib/embed/embed_this_panel.rb
```diff
@@ -74,7 +74,7 @@ def self.iframe_html(druid:, height:, width:, request:, title: '')
                      else
                        "#{height}px"
                      end
-      src = "#{Settings.embed_iframe_url}?url=#{Settings.purl_url}/#{druid}&#{request.as_url_params.to_query}"
+      src = "#{Settings.embed_iframe_url}?url=#{Settings.purl_url}/#{druid}&#{request.as_url_params.to_query}&_v=#{File.mtime(Rails.root)}"
```
jkeck, line 77: I suppose an unfortunate side affect of us re-using this code is that this will be given to users in the `<iframe>` snippet as well as what's served up through the oEmbed API.  One concern I have is that it may be detrimental/confusing to have this parameter in the `<iframe>` snippet that users copy/paste into their blog/website/etc.

Would it be possible to pass this in as a parameter so that it could be sent it when we're generating the oEmbed API response but not when we generate the iframe for users to embed?  Or alternative ways of disentangling these two things in a reasonable way?

cbeer: How about that?

jkeck: Perfect, thank you!

PR #1169, opened 2020-06-25
Analytics

cbeer:
TODO:

- [x] how long is a logged in session
- [x] token invalid
- [x] renews (auth flow + auth flow) (something something renew command in symphony)
- [ ] ~join waitlist (can we get current number of holds?) (symphony => WAITLIST / NEXT_UP)~
- [ ] ~checkout w/o hold (symphony comment => ACTIVE/COMPLETED/EXPIRED)~
- [ ] ~hold converted to checkout (symphony comment => ACTIVE/COMPLETED/EXPIRED)~
- [ ] ~where they're coming from (myaccount? sw? etc) (not tracking)~
- [ ] ~sym user profile or Stanford affiliation type?~ (symphony analytics from hold comment)


config/settings/production.yml
```diff
@@ -1,2 +1,3 @@
 static_assets_base: 'https://embed.stanford.edu'
 cache_life: <%= 15.minutes.to_s %>
+analytics_id: UA-7219229-37
```
mejackreed, line 3: In shared configs instead?

app/views/embed/iiif.html.erb
```diff
@@ -35,6 +35,17 @@
...
       </div>
-     </div>
+    </div>
+    <% if Settings.analytics_id %>
+      <script>
+        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+        })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
```
mejackreed, line 45: Anonymize here also?

app/assets/javascripts/plugins/analyticsPlugin.js
```diff
@@ -0,0 +1,167 @@
...
+  yield put({ type: 'mirador/analytics', payload: {
+    hitType: 'event',
+    eventCategory: manifestId,
+    eventAction: type,
+    eventLabel: profile,
+  }});
+}
+
+function* sendAnalyticsEvent({ payload }) {
+  window.ga && window.ga('send', payload);
```
mejackreed, line 140: Does this need to be:

```
window.ga('send',Object.values(payload));
```

Or does that automatically happen? Looking at https://developers.google.com/analytics/devguides/collection/analyticsjs/events

cbeer: Further down the doc, there's also:

```
ga('send', {
  hitType: 'event',
  eventCategory: 'Videos',
  eventAction: 'play',
  eventLabel: 'Fall Campaign'
});
```

jkeck (approved):

PR #1179, opened 2020-07-21
Update to rails 6

jkeck:
I'm getting a flapping geo viewer spec that's also failing for me on master.


cbeer:
Do you know if we need to add `skip_forgery_protection` to the JSON endpoints?

jkeck:
@cbeer I don't think that would be necessary for this app (but perhaps I'm not thinking of somewhere that would be helpful).

Is there a rails 6 changes WRT `skip_forgery_protection` that I should be keeping an eye on?

cbeer:
I'm not sure; I just remember we've been burned by it a couple times now.

jkeck:
Oddly the docs say that the change was to move the `default_protect_from_forgery` config `ActionController::Base` but it was kept `false` by default. Although our setup is similar to stacks where I believe we've run into this before, (e.g. no explicit `protect_from_forgery` call) things _seem_ to be working out (I deployed this to UAT).

I'm happy to add `skip_forgery_protection` to the. places where we might need it in case we want to be extra cautious.

PR #1192, opened 2020-08-26
Revert "Update javascript dependencies"

mejackreed:
Reverts sul-dlss/sul-embed#1191

Reverting for now due to platform issues installing on `linux`


mejackreed:
Nevermind, just updating node version on server

PR #1227, opened 2020-11-23
Time is fun

### LINKED ISSUE #1216: CDL: Translate loan periods >24 hours to accurately display due date/time
- [x] When not checked out, translate the loan period to " x day" or "x week" loans.  (e.g. 24-hour = "1 day", 168-hour = "1 week")

- [ ] When item is checked out, include the date the item is due (in addition to the time).  This needs to display to the user that has the item checked out AND anyone joining the waitlist.

Current view when item not checked out:
image

Current view for user with item checked out.  Missing is the information that this is due 1 week from today at 1:15pm PDT:
image

Current view for others when item is checked out.  Again, missing is that the item is due in 1 week at 1:15pm PDT:
image

 
Example record in Prod: http://purl.stanford.edu/cy779fp8559

mejackreed:
Fixes #1216

**Note** this assumes that the CDL loan period is defined in terms we would expect (minutes -> hours -> days -> weeks. So a loan period that say is 17 days, it would display as `1-week`. We could change this, but I'm not 💯 clear on when / why we should fall back to that.

## Due today
Screen Shot 2020-11-23 at 12 45 40 PM

## 1-week loan
Screen Shot 2020-11-23 at 12 44 32 PM

## Faked not today or tomorrow display
Screen Shot 2020-11-23 at 12 15 31 PM

## Due today
Screen Shot 2020-11-23 at 12 15 03 PM

## Faked due tomorrow
Screen Shot 2020-11-23 at 12 14 41 PM

## Other loan periods
Screen Shot 2020-11-23 at 12 49 00 PM
Screen Shot 2020-11-23 at 12 48 55 PM

### DIFF
--- a/app/assets/javascripts/components/DueDate.js
+++ b/app/assets/javascripts/components/DueDate.js
@@ -5,6 +5,28 @@ import Typography from '@material-ui/core/Typography';
 /** */
 export default class DueDate extends Component {
   /** */
+  static isToday(dueDate) {
+    const today = new Date();
+    if (today.getDate() === dueDate.getDate()
+        && today.getMonth() === dueDate.getMonth()
+        && today.getFullYear() === dueDate.getFullYear()) {
+      return true;
+    }
+    return false;
+  }
+
+  /** */
+  static isTomorrow(dueDate) {
+    const tomorrow = new Date(new Date().getTime() + 86400000);
+    if (tomorrow.getDate() === dueDate.getDate()
+        && tomorrow.getMonth() === dueDate.getMonth()
+        && tomorrow.getFullYear() === dueDate.getFullYear()) {
+      return true;
+    }
+    return false;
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/DueDate.js
```diff
@@ -4,6 +4,28 @@ import Typography from '@material-ui/core/Typography';
...
+        && today.getMonth() === dueDate.getMonth()
+        && today.getFullYear() === dueDate.getFullYear()) {
+      return true;
+    }
+    return false;
+  }
+
+  /** */
+  static isTomorrow(dueDate) {
+    const tomorrow = new Date(new Date().getTime() + 86400000);
```
mejackreed, line 20: 🤷 I think this is the way to do this without adding other dependencies.

PR #1242, opened 2021-02-17
Remove teaspoon in favor of Jest, paving the way for Ruby 3.0 support

mejackreed:


app/assets/javascripts/modules/common_viewer_behavior.js
```diff
@@ -28,6 +28,11 @@
     };
   })();
 
+  // Basic support of CommonJS module for import into test
+  if (typeof exports === "object") {
```
mejackreed, line 32: The "magic" allowing us to use `require` directly in our jest tests

cbeer (approved):

PR #1249, opened 2021-03-12
Update videojs; fixes #929

### LINKED ISSUE #929: Upgrade VideoJS
We are a few major versions behind.  We don't interact w/ the API directly, so hopefully there won't be too many breaking changes that affect us, but testing will need to be done.

cbeer:


mejackreed:
Did we never need to support flash here?

cbeer:
Flash is officially end-of-life, so 🤷‍♂️

PR #1265, opened 2021-05-06
Support the new style of licenses from h2

jcoyne:
See https://github.com/sul-dlss/purl/pull/486


mjgiarlo:
I think we settled on a different approach for licenses. Can this be closed, @jcoyne?

PR #1275, opened 2021-06-10
Look in MODS for license first

### LINKED ISSUE #1274: Change how we render license information
Now that public XML will include the license statement text all the access systems display, we will soon be out of the business of parsing identifiers and URIs and looking them up. The behavior we are aiming to support across the access & discovery stack is:

1. First, look in the MODS in public XML (`/publicObject/mods/accessCondition[@type="license"]`) for the license statement (as a single text child node). If it is non-empty, our work is done.
    * If the license node has a non-empty `@xlink:href` attribute, and the application in question linkifies license statements, use the URI; else, ignore it.
2. Failing :point_up:, which we can expect for older objects, using CC3 and/or ODC licenses, check in rights metadata (`/publicObject/rightsMetadata/use/human[@type="creativeCommons" or @type="openDataCommons"]`) and map the label (e.g., `Attribution-NonCommercial 3.0 Unported`) to a human-friendly license statement. At some point, we endeavor to be rid of these look-ups entirely, but for now we keep them for this data which is the bulk of our license-having data at this point (mid-2021).

mjgiarlo:
Fixes #1274

Also moves away from using the "machine" licenses since we are moving away from badges.

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -46,7 +46,7 @@ module Embed
     end
 
     def license
-      cc_license || odc_licence
+      cc_license || odc_license || mods_license
     end
 
     def embargo_release_date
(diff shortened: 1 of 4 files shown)


app/models/embed/purl.rb
```diff
@@ -46,7 +46,7 @@ def copyright
     end
 
     def license
-      cc_license || odc_licence
+      mods_license || cc_license || odc_license
```
cbeer, line 49: I don't think this will do the right thing with the old-style MODS licenses for CC v3 content, e.g.

https://purl.stanford.edu/gw298sq0504.xml

(Not that that particular object has the rights panel, but... presumably there's a real example out there?)

I think it'd be fine to check cc/odc first, or we could use the `mods_display` gem here to just render it?

mjgiarlo: @cbeer good :eye:. Take two is now up. LMK if that's sufficient or if we need the `mods_display` dependency. :man_shrugging:

app/views/embed/_metadata_panel.html.erb
```diff
@@ -28,8 +28,7 @@
         <% if viewer.purl_object.license.present? %>
           <dt>License</dt>
           <dd>
-            <span class='sul-embed-license-<%= viewer.purl_object.license[:machine] %>'></span>
```
mjgiarlo, line 31: Just calling this out. It's OK to move away from these spans and the "machine" biz entirely at this point, yeah?

cbeer: I assume they just drive the badgy things? I don't know if there are some css classes we get to drop?

mjgiarlo: @cbeer I searched around and found none of these `sul-embed-license-*` CSS classes but perhaps I just missed them or they're dynamic or something?

cbeer: 🤷‍♂️

Issue #1342, opened 2022-06-14
Inconsistent test results in rspec


marlo-longley:
### Expected behavior

A test will either pass or fail consistently if its code remains unchanged.

### Actual behavior

In CI, different rspec tests are regularly failing in both the Ruby 2.7 and 3.0 runs. These tests generally have been feature tests which require clicking an element and then expecting other elements to exist on the page. These tests don't fail consistently; it depends on the mood of CI at the time.

### Steps to reproduce the behavior

Run CI on the repo. Likely you will get some failed test. Run it again. You will get different failed tests, and the initial failures will pass.

### Solution

Looking into this with @thatbudakguy we think this has to do with race conditions in Capybara. We want to go through any tests that involve clicking and loading new DOM elements, and try to use the[ `find` Capybara method,](https://rubydoc.info/github/teamcapybara/capybara/master#the-dsl) which should wait for elements to load. We can keep track of which tests we've updated in a PR.

thatbudakguy:
@marlo-longley for posterity i think it'd be useful to list here every test that you've seen both pass & fail in CI at different times, i.e. the ones that seem to be unreliable.

thatbudakguy:
# failures
```
 embed this panel file objects are present after a user clicks the button
     Failure/Error: expect(page).to have_css('.sul-embed-embed-this-panel', visible: true)
       expected to find visible css ".sul-embed-embed-this-panel" but there were no matches. Also found "", which matched the selector but not all filters. 
     # ./spec/features/embed_this_panel_spec.rb:41:in `block (3 levels) in <top (required)>'

 embed this panel embed code includes the allowfullscreen no-scrolling, no-border, and no margin/padding attributes
     Failure/Error: expect(page.find('.sul-embed-embed-this-panel textarea').value).to match(/<iframe.*frameborder="0"/m)
     Capybara::ElementNotFound:
       Unable to find visible css ".sul-embed-embed-this-panel textarea"
     # ./spec/features/embed_this_panel_spec.rb:19:in `block (3 levels) in <top (required)>'

  geo viewer public loading geo viewer includes proper attributes for a _blank target on the download links
     Failure/Error:
       within '.sul-embed-download-panel' do
         within '.sul-embed-panel-body' do
           expect(page).to have_css('li a[target="_blank"][rel="noopener noreferrer"]', count: 3)
         end
       end
     Capybara::ElementNotFound:
       Unable to find visible css ".sul-embed-download-panel"
[...]

PR #1343, opened 2022-06-14
Flakey tests

marlo-longley:
Related to #1342


thatbudakguy:
based on an idea I got from @cbeer, it looks like the solution here was to run a virtual X display using `xvfb`, and then execute Chrome in headful mode and specify a sufficiently large window size for Capybara tests.

this PR also threw in `rubocop-rspec` (#1346) and reformatted our tests in the hope that it would cut down on errors, which did actually uncover one bugged test that was looking for an element capitalized via CSS.

corylown (approved):

PR #1344, opened 2022-06-20
Test changes to timemap behavior

mjgiarlo:
The Infrastructure team will soon be replacing openwayback with pywb, which implements slightly different Memento timemap behavior. Tweak the tests to gauge how much this will break in sul-embed.


spec/models/embed/was_time_map_spec.rb
```diff
@@ -31,6 +31,32 @@
...
+
+      it 'the memento lines are not all memento entries' do
+        expect(subject.timemap.count(&:memento?)).to eq 7
+      end
+
+      it 'memento lines have accessors' do
+        memento_line = subject.timemap.find(&:memento?)
+        expect(memento_line.url).to eq 'https://swap.stanford.edu/20090718213431/http://ennejah.info/'
+        # NOTE: this needed changing to pass! Problematic?
+        expect(memento_line.rel).to eq 'memento'
```
mjgiarlo, line 55: Heads-up. Not sure sul-embed otherwise special-cases "last memento" or "first memento"? if so, that's something that may trip us up.

PR #1355, opened 2022-07-14
Remove embed_iframe_url from settings in favor of controller data

marlo-longley:
We get `iframe_url` for free from the embed controller's `iframe` route.

This PR also removes any references to an unneeded asset_url method we'd put into the viewer code, and falls back to the rails `asset_url` helper.


thatbudakguy (approved):
all looks good!

PR #1368, opened 2022-08-29
Update leaflet-rails

cbeer:


marlo-longley:
This was the culprit
Screen Shot 2022-08-29 at 12 21 04

vs 1.7.0
Screen Shot 2022-08-29 at 12 21 51

PR #1376, opened 2022-10-28
Use literal slashes in stacks URLs with file-type content

mjgiarlo:
Connects to sul-dlss/dor-services-app#4256

This commit supports work to support hierarchy in filenames throughout SDR. Doing this to support analysis work. Depends on https://github.com/sul-dlss/stacks/pull/925

### DIFF
--- a/app/viewers/embed/viewer/common_viewer.rb
+++ b/app/viewers/embed/viewer/common_viewer.rb
@@ -34,11 +34,10 @@ module Embed
       # Creates a file url for stacks
       # @param [String] title
       # @param [Boolean] download
-      # @param [Boolean] allow_literal_slashes
       # @return [String]
-      def file_url(title, download: false, allow_literal_slashes: false)
-        title_parts = allow_literal_slashes ? title.split('/') : [title]
-        encoded_title = title_parts.map { |title_part| ERB::Util.url_encode(title_part) }.join('/')
+      def file_url(title, download: false)
+        # Allow literal slashes in the file URL (do not encode them)
+        encoded_title = title.split('/').map { |title_part| ERB::Util.url_encode(title_part) }.join('/')
         uri = URI.parse("#{stacks_url}/#{encoded_title}")
         uri.query = URI.encode_www_form(download: true) if download
         uri.to_s
--- a/app/viewers/embed/viewer/common_viewer.rb
+++ b/app/viewers/embed/viewer/common_viewer.rb
@@ -34,9 +34,12 @@ module Embed
       # Creates a file url for stacks
       # @param [String] title
       # @param [Boolean] download
+      # @param [Boolean] allow_literal_slashes
       # @return [String]
-      def file_url(title, download: false)
-        uri = URI.parse("#{stacks_url}/#{ERB::Util.url_encode(title)}")
+      def file_url(title, download: false, allow_literal_slashes: false)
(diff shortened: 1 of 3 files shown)


app/viewers/embed/viewer/common_viewer.rb
```diff
@@ -34,9 +34,12 @@ def external_url
       # Creates a file url for stacks
       # @param [String] title
       # @param [Boolean] download
+      # @param [Boolean] allow_literal_slashes
       # @return [String]
-      def file_url(title, download: false)
-        uri = URI.parse("#{stacks_url}/#{ERB::Util.url_encode(title)}")
+      def file_url(title, download: false, allow_literal_slashes: false)
+        encoded_title = ERB::Util.url_encode(title)
+        encoded_title.gsub!('%2F', '/') if allow_literal_slashes
```
cbeer, line 41: I don't know if rubocop would like it, but maybe something like this would be a little more robust?

```suggestion
        parts =  allow_literal_slashes ? title.split('/') : [title]
        encoded_title  = title.map { |x| ERB::Util.url_encode(x) }.join('/')
```

I get nervous about trying to undo encoded stuff after the fact 🤷‍♂️

app/viewers/embed/viewer/common_viewer.rb
```diff
@@ -34,9 +34,12 @@ def external_url
       # Creates a file url for stacks
       # @param [String] title
       # @param [Boolean] download
+      # @param [Boolean] allow_literal_slashes
       # @return [String]
-      def file_url(title, download: false)
-        uri = URI.parse("#{stacks_url}/#{ERB::Util.url_encode(title)}")
+      def file_url(title, download: false, allow_literal_slashes: false)
```
jcoyne, line 39: in which cases would we not want to allow literal slashes? can you add this to the inline documentation above? Do we need to keep this as an argument?

jcoyne (changes requested):

app/viewers/embed/viewer/common_viewer.rb
mjgiarlo: @jcoyne 💬 
> in which cases would we not want to allow literal slashes? 

That is the current behavior, so apparently all cases except the one we're adding for file-type objects? I don't really know. 

> can you add this to the inline documentation above?

Happy to add a blurb about why this arg was added, you bet.

> Do we need to keep this as an argument?

Good question, and I don't know. My intent here is to add a small seam to support the behavior needed for these filenames-with-path-info-in-them, without breaking anything else. If y'all think it's safe to move to literal slashes everywhere, I can make the change of course. That seems like it'd add significantly to the testing we'll have to do, and I'll need to lean on y'all for help with that.

jcoyne: I would rather change it to accept literal slashes in all cases an repeat the testing. While that is a greater burden now, over the long run it will be less complexity to support.

mjgiarlo: @jcoyne OK, that's fine with me, and I'm happy to re-run the test with the ~7-8 test objects in stage.

mjgiarlo: @jcoyne Mind having another look?

jcoyne (approved):

PR #1390, opened 2022-12-16
Surface containing directory information in file embed view

mjgiarlo:
Connects to #1380

This commit is WIP towards implementing a small change to surface containing directory information in the file embed view, intended to provide more context around files that were stored with path information.

Screenshot from 2022-12-15 16-27-58
Screenshot from 2022-12-15 16-28-17

NOTE: No tests have been written or run. Also, there is a bug in the current implementation causing the JavaScript item counter to overcount (likely due to the HTML elements added to show directory info).

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -33,6 +33,13 @@ module Embed
       end
     end
 
+    def containing_directories
+      all_resource_files
+        .map(&:containing_directory)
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@justinlittman I assume we can ditch this now. Confirm/deny?

justinlittman:
Confirmed.

PR #1392, opened 2022-12-16
Dev instructions

edsu:
This PR adds a few instructions for getting a working dev environment going. Please disregard if they aren't helpful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,23 +8,30 @@ An [oEmbed](http://oembed.com/) provider for embedding resources from the Stanfo
 
 There is an embedded static page available at `/pages/sandbox` in your development and test environments. Make sure that you use the same host on the service input (first text field) as you are accessing the site from (e.g. localhost or 127.0.0.1).
 
+To bring up a dev environment first you'll need to install Ruby and JavaScript dependencies. Note: NodeJS v14 and [yarn](https://yarnpkg.com/) must be installed:
+
+    bundle install
+    yarn install
+
+Then start up the Rails app in one terminal window:
+
+    bundle exec bin/rails server
+
+Then start the NodeJS process to make the JavaScript assets available:
+
+    bin/webpack-dev-server
+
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -38,3 +38,5 @@ yarn.lock
 /yarn-error.log
 yarn-debug.log*
 .yarn-integrity
+
+.node-version
```
mjgiarlo, line 42: You may want to consider adding this to your global `.gitignore`, too (or instead?)

mjgiarlo (approved):
LGTM! We should prolly defer to @sul-dlss/access-team folk on merging.

.gitignore
edsu: Good point, I just removed it.

README.md
```diff
@@ -8,23 +8,30 @@ An [oEmbed](http://oembed.com/) provider for embedding resources from the Stanfo
...
 
-To install needed JavaScript dependencies make sure to install them using:
+## oEmbed specification details
+
+URL scheme: `http://purl.stanford.edu/*`
 
-```sh
-$ yarn install
-```
+API endpoint: `TBD`
```
thatbudakguy, line 32: what's it mean that this is TBD?

thatbudakguy (approved):
Thanks for adding this!

README.md
thatbudakguy: ...I guess that was already there. I think this should be `embed.stanford.edu` now, probably?

mjgiarlo: I guess?

mjgiarlo: Either that or we can consider renaming the prod embed service `TBD.stanford.edu`!

edsu: This was out of the scope of the changes I was making to just make it a bit easier to get up and running for development. But I just pushed a change to make this hostname and the example to hopefully reflect reality.

PR #1397, opened 2022-12-20
Switch to shakapaker

jcoyne:


bin/webpacker
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env ruby
```
marlo-longley, line 1: @jcoyne this file has very similar contents to bin/webpack-dev-server, except for a different runner. What is its purpose?

jcoyne: It builds the javascript. See https://github.com/shakacode/shakapacker#development

bin/webpacker-dev-server
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env ruby
```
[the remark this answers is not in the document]
jcoyne: @marlo-longley hasn't that already been done? You're pointing at a commit not in this branch.

corylown (commented):
This could just be an issue with my local setup or maybe it's expected, but I had to run `yarn add @blueprintjs/core` and `yarn add @blueprintjs/icons` to get `bin/webpacker-dev-server` to start without errors. I was seeing things like: `Module not found: Error: Can't resolve '@blueprintjs/core' in '/Users/corylown/workspace/sul-embed/node_modules/react-mosaic-component/lib/util'`

Update: not harmful and due to blueprintjs being an optional dependency

corylown (approved):

PR #1413, opened 2023-01-10
Display of hierarchical files.

justinlittman:
refs #1380

Supersedes #1390


app/assets/javascripts/modules/dir_toggle.js
```diff
@@ -0,0 +1,50 @@
+// Module adds behavior to Preview/Close preview for file objects
+
+(function( global ) {
+  var Module = (function() {
+
+    return {
+      init: function() {
+        $('.sul-embed-container').each(function(){
```
jcoyne, line 8: Please avoid using jquery for this as we are trying to eliminate jQuery as a dependency here.

jcoyne (changes requested):

app/assets/javascripts/modules/dir_toggle.js
justinlittman: This is just a variation on `file_preview.js`. However, I got rid of all of the jquery except `.slideDown()` and `.slideUp()`. I'm open to suggestions on how to get rid of them, but since they are in use elsewhere not sure if it is desirable to have multiple slide implementations that may look different.

app/assets/javascripts/modules/dir_toggle.js
```diff
@@ -0,0 +1,48 @@
+// Module adds behavior to Preview/Close preview for file objects
+
+(function( global ) {
+  var Module = (function() {
+
+    return {
+      init: function() {
+        document.querySelectorAll('.sul-embed-container').forEach(function(container){
```
jcoyne, line 8: Please use `data-*` selectors for behavior rather than classes.  This will help differentiate behavior from style.

app/assets/javascripts/modules/dir_toggle.js
```diff
@@ -0,0 +1,48 @@
...
+        container.querySelectorAll('[data-sul-embed-dir-toggle="true"]').forEach(function(toggle){
+          toggle.addEventListener('click', function(){
+            DirToggle.toggleDirToggleIcon(toggle);
+            DirToggle.toggleDirWindow(toggle);
+          });
+        });
+      },
+      toggleDirToggleIcon: function(toggle) {
+        var icon = toggle.querySelector('i');
+        if(icon.classList.contains('sul-i-arrow-right-8')) {
```
jcoyne, line 22: Can you add a class here like `open` when it is open and change the icon in the css?

app/models/embed/purl.rb
```diff
@@ -145,6 +152,7 @@ def purl_xml_url
     def response
       @response ||=
         begin
+          Rails.logger.info(purl_xml_url)
```
jcoyne, line 155: cruft?

app/models/embed/purl.rb
```diff
@@ -160,6 +168,46 @@ def response
         end
     end
 
+    def file_contents
```
jcoyne, line 171: This class is too big. I don't know what it's responsibilities are.  Can you extract these new methods to a new class?

config/settings/development.yml
```diff
@@ -0,0 +1,3 @@
+valid_purl_url: <%= /http?:\/\/localhost:3000\/*/ %>
```
jcoyne, line 1: Why is this in erb?

justinlittman: Based on https://github.com/sul-dlss/sul-embed/blob/master/config/settings.yml#L6

jcoyne: Let's not perpetuate bad choices. Here's a fix https://github.com/sul-dlss/sul-embed/pull/1423

app/assets/javascripts/modules/dir_toggle.js
justinlittman: This is taken from https://github.com/sul-dlss/sul-embed/blob/master/app/assets/javascripts/modules/file_preview.js#L8. I'm reluctant to introduce different approaches in the codebase.

app/assets/javascripts/modules/dir_toggle.js
justinlittman: Same here too: https://github.com/sul-dlss/sul-embed/blob/master/app/assets/javascripts/modules/file_preview.js#L24-L29

justinlittman:
@ggeisler @astridu  Here is a first cut at the hierarchical display. Note that the file count and search still has not been addressed.

https://user-images.githubusercontent.com/588335/213291312-2f391301-6c65-43d6-9820-315af21a6d93.mov

ggeisler:
@justinlittman This looks good to me. One super minor thing, that I would defer to @astridu since she's more familiar with the intended design: could we move the indentation of nested levels to the right by a couple of pixels, so that the left-edge of the leftmost element aligns left with the left border of the image thumbnail in the level above it? 

What I mean is, in the example below the first indentation looks nice and left-aligned, but see how the next two levels (the expand/collapse control in the first, the number "12" in the second) are slightly too far left to be left-aligned with the left border of the thumbnail above them)? We'd lose two or three horizontal pixels at each level doing this, but I think it would look a bit sharper if everything lined up nicely on a left edge, so that might be a worthwhile sacrifice.

Screen Shot 2023-01-18 at 3 08 43 PM

Without directly working with the code it's hard for me to be sure this is a do-able suggestion, so feel free to ignore if it is more complicated than I'm suggesting.

app/assets/javascripts/modules/dir_toggle.js
jcoyne: Please don't use existing code as a good example of what we want in the future. Things were done differently 8 years ago because they had requirements that no longer exist.

andrewjbtw:
@justinlittman @ggeisler @astridu 

Astrid and I went over a number of sample items in hierarchy on stage. I think the main challenge remains the search/filter box. But I'll leave that to the end of this comment. Overall, I think the implementation looks good for both hierarchies and objects with flat file lists.

I have some sample druids in stage under the "Project : hfs" tag: https://argo-stage.stanford.edu/catalog?f%5Btag_ssim%5D%5B%5D=Project+%3A+hfs 

**Changes**

1. Shifting the alignment of the indents: if it's do-able it makes a lot of sense to line things up as Gary suggests

2. If possible, show a scrollbar if a filename extends past the viewer border. In my testing, I found the text wraps on spaces and dashes but not on long strings connected only by underscores, where the name just gets cut off:

Screen Shot 2023-01-18 at 5 08 27 PM

**Image previews

Unfortunately, image previews don't work for JP2s in hierarchies. I'm not sure what we can do about that. My guess is that the image server, rather than Stacks, must be serving up the previews and it's not parsing the hierarchy even when the Stacks links work. 

There are at most 20 production objects that use image previews in a way that could be considered functional. It's more common even in the current implementation to find image previews that display blank squares and empty space where an image would be, and even then the number of attempts to use previews is under 200. Given that, and the unlikelihood of users trying to make use of this functionality with hierarchy, I think it would be acceptable to have hierarchy without working image previews. 

(And to go further I would recommend that image preview functionality be removed from the file viewer in the future unless it could be implemented in a way that aligns better with our existing data, rights settings, and deposit workflows.)

**Filename search**

This is the big one. It turns out that when the hierarchy implementation is deployed, search stops working even on non-hierarchical objects. Regardless of whether or not an object uses hierarchy, the top bar of the embed viewer shows "2 items" and typing text into the search box does not filter the list as expected:

https://user-images.githubusercontent.com/8975583/213332773-9915dc6d-bca7-4293-9a37-71a846902592.mov

That leaves us with three options:

1. re-implement search with the same behavior for any object in the file viewer
2. if possible, restore search for flat lists using the existing implementation but exclude the search box altogether from hierarchies
3. remove search altogether from all objects

Astrid and I came to the conclusion that option 1 would be ideal, as it wouldn't require a different logic for hierarchies and flat lists. Everything would have a search. If we implement the search filter for hierarchies, then requirements are:

- match text in filenames
- match text in the object "labels" (the optional field where users can describe files)
- display only the files that match the search criteria
- it's ok to ignore folders for search purposes

The idea behind option 2 is that it might be possible to maintain the existing search for objects that currently use it. A search that handles hierarchy could then be added in a later iteration.

Option 3 of removing search is the most straightforward but would have the most noticeable impact on users. I analyzed all existing file objects and found that about 1000 currently use the search bar, mostly self-deposit and ETD objects. We unfortunately are short on time to do the research we need to determine how this function is actually used, which makes me reluctant to suggest removing it.

jcoyne:
@justinlittman for accessability, this should probably implement the ARIA treeview https://www.w3.org/WAI/ARIA/apg/patterns/treeview/  See especially the section on keyboard interaction.

justinlittman:
Yeah, I'm working on switching to tree following https://iamkate.com/code/tree-views/. Seems like a potential win from simplicity (no JS) and accessability.

jcoyne:
@justinlittman I consulted with Kate (from #code4lib) and she indicated that's not a great pattern to follow with regards to accessability + hierarchy.  She urged looking at the treeview pattern as it is better for accessability.

jcoyne:
Consider https://sap.github.io/ui5-webcomponents/playground/components/Tree/ perhaps?

justinlittman:
@andrewjbtw :
1. I'd like to defer the scrollbar to a separate ticket. I don't understand the implications for the rest of sul-embed.
2. For this ticket, my plan is to pull search for all file objects. Option 2 isn't possible. Option 1 is feasible, but not in the scope of this PR. I'm happy to work on it once this is completed.

andrewjbtw:
@justinlittman Both of those points are ok with me. I'll split scrollbar and search to their own tickets.

config/settings/development.yml
jcoyne: This value is no longer used in the code. Please remove it.

spec/features/file_hierarchy_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'file viewer with hierarchy', js: true do
```
jcoyne, line 5:
```suggestion
RSpec.describe 'file viewer with hierarchy', js: true do
```

spec/features/file_hierarchy_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    expect(page).to have_css('.sul-embed-treeitem', count: 2)
+    # One of the files is nested
+    expect(page).to have_css('[role="treeitem"] [role="treeitem"] [role="treeitem"] .sul-embed-media', count: 1)
+  end
+
+  it 'toggles hierarchy' do
+    stub_purl_response_with_fixture(hierarchical_file_purl)
+    visit_iframe_response
+
+    expect(page).to have_css('.sul-embed-media', count: 2)
```
jcoyne, line 23:
```suggestion
```

Can we combine these?  It's a single feature.  The feature tests are slow so we avoid having more than necessary.

spec/lib/embed/media_tag_spec.rb
```diff
@@ -158,27 +158,27 @@
   describe '#media_wrapper' do
     describe 'data-stanford-only attribute' do
       it 'true for Stanford only files' do
-        resource_file = double(Embed::Purl::Resource::ResourceFile, stanford_only?: true, location_restricted?: false, label: 'ignored', duration: nil)
+        resource_file = double(Embed::Purl::ResourceFile, stanford_only?: true, location_restricted?: false, label: 'ignored', duration: nil)
```
jcoyne, line 161:
```suggestion
        resource_file = instance_double(Embed::Purl::ResourceFile, stanford_only?: true, location_restricted?: false, label: 'ignored', duration: nil)
```

spec/lib/embed/media_tag_spec.rb
```diff
@@ -158,27 +158,27 @@
...
         media_wrapper = Capybara.string(subject_klass.send(:media_wrapper, file: resource_file))
         expect(media_wrapper).to have_css('[data-stanford-only="true"]')
       end
 
       it 'false for public files' do
-        resource_file = double(Embed::Purl::Resource::ResourceFile, stanford_only?: false, location_restricted?: false, label: 'ignored', duration: nil)
+        resource_file = double(Embed::Purl::ResourceFile, stanford_only?: false, location_restricted?: false, label: 'ignored', duration: nil)
```
jcoyne, line 167:
```suggestion
        resource_file = instance_double(Embed::Purl::ResourceFile, stanford_only?: false, location_restricted?: false, label: 'ignored', duration: nil)
```

spec/lib/embed/media_tag_spec.rb
```diff
@@ -158,27 +158,27 @@
...
         media_wrapper = Capybara.string(subject_klass.send(:media_wrapper, file: resource_file))
         expect(media_wrapper).to have_css('[data-stanford-only="false"]')
       end
     end
 
     describe 'duration' do
       it 'sets the duration when the resource file has duration' do
-        resource_file = double(Embed::Purl::Resource::ResourceFile, label: 'ignored', duration: '1:02')
+        resource_file = double(Embed::Purl::ResourceFile, label: 'ignored', duration: '1:02')
```
jcoyne, line 175:
```suggestion
        resource_file = instance_double(Embed::Purl::ResourceFile, label: 'ignored', duration: '1:02')
```

spec/lib/embed/media_tag_spec.rb
```diff
@@ -158,27 +158,27 @@
...
         media_wrapper = Capybara.string(subject_klass.send(:media_wrapper, file: resource_file))
         expect(media_wrapper).to have_css('[data-duration="1:02"]')
       end
 
       it 'leaves the duration empty when the resource file is missing duration' do
-        resource_file = double(Embed::Purl::Resource::ResourceFile, label: 'ignored', duration: nil)
+        resource_file = double(Embed::Purl::ResourceFile, label: 'ignored', duration: nil)
```
jcoyne, line 181:
```suggestion
        resource_file = instance_double(Embed::Purl::ResourceFile, label: 'ignored', duration: nil)
```

spec/lib/embed/media_tag_spec.rb
```diff
@@ -188,27 +188,27 @@
     describe 'data-location-restricted attribute' do
       context 'stanford_only' do
         it 'true for location restricted files' do
-          resource_file = double(Embed::Purl::Resource::ResourceFile, stanford_only?: true, location_restricted?: true, label: 'ignored', duration: nil)
+          resource_file = double(Embed::Purl::ResourceFile, stanford_only?: true, location_restricted?: true, label: 'ignored', duration: nil)
```
jcoyne, line 191:
```suggestion
          resource_file = instance_double(Embed::Purl::ResourceFile, stanford_only?: true, location_restricted?: true, label: 'ignored', duration: nil)
```

spec/lib/embed/media_tag_spec.rb
```diff
@@ -188,27 +188,27 @@
...
           media_wrapper = Capybara.string(subject_klass.send(:media_wrapper, file: resource_file))
           expect(media_wrapper).to have_css('[data-location-restricted="true"]')
         end
 
         it 'false for public files' do
-          resource_file = double(Embed::Purl::Resource::ResourceFile, stanford_only?: true, location_restricted?: false, label: 'ignored', duration: nil)
+          resource_file = double(Embed::Purl::ResourceFile, stanford_only?: true, location_restricted?: false, label: 'ignored', duration: nil)
```
jcoyne, line 197:
```suggestion
          resource_file = instance_double(Embed::Purl::ResourceFile, stanford_only?: true, location_restricted?: false, label: 'ignored', duration: nil)
```

spec/lib/embed/media_tag_spec.rb
```diff
@@ -188,27 +188,27 @@
...
           media_wrapper = Capybara.string(subject_klass.send(:media_wrapper, file: resource_file))
           expect(media_wrapper).to have_css('[data-location-restricted="false"]')
         end
       end
 
       context 'not stanford_only' do
         it 'true for location restricted files' do
-          resource_file = double(Embed::Purl::Resource::ResourceFile, stanford_only?: false, location_restricted?: true, label: 'ignored', duration: nil)
+          resource_file = double(Embed::Purl::ResourceFile, stanford_only?: false, location_restricted?: true, label: 'ignored', duration: nil)
```
jcoyne, line 205:
```suggestion
          resource_file = instance_double(Embed::Purl::ResourceFile, stanford_only?: false, location_restricted?: true, label: 'ignored', duration: nil)
```

spec/lib/embed/media_tag_spec.rb
```diff
@@ -188,27 +188,27 @@
...
           media_wrapper = Capybara.string(subject_klass.send(:media_wrapper, file: resource_file))
           expect(media_wrapper).to have_css('[data-location-restricted="true"]')
         end
 
         it 'false for public files' do
-          resource_file = double(Embed::Purl::Resource::ResourceFile, stanford_only?: false, location_restricted?: false, label: 'ignored', duration: nil)
+          resource_file = double(Embed::Purl::ResourceFile, stanford_only?: false, location_restricted?: false, label: 'ignored', duration: nil)
```
jcoyne, line 211:
```suggestion
          resource_file = instance_double(Embed::Purl::ResourceFile, stanford_only?: false, location_restricted?: false, label: 'ignored', duration: nil)
```

spec/models/embed/purl/resource_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Embed::Purl::Resource do
```
jcoyne, line 5:
```suggestion
RSpec.describe Embed::Purl::Resource do
```

.rubocop_todo.yml
```diff
@@ -62,7 +63,8 @@ RSpec/ContextWording:
     - 'spec/lib/embed/stacks_media_stream_spec.rb'
     - 'spec/lib/embed/viewer/file_spec.rb'
     - 'spec/lib/embed/viewer_factory_spec.rb'
-    - 'spec/models/embed/purl_spec.rb'
+    - 'spec/models/embed/purl/resource_file_spec.rb'
+    - 'spec/models/embed/purl/resource_spec.rb'
```
jcoyne, line 67: I'm not seeing this in resource_spec.rb

spec/models/embed/purl/resource_file_spec.rb
```diff
@@ -0,0 +1,279 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Embed::Purl::ResourceFile do
```
jcoyne, line 5:
```suggestion
RSpec.describe Embed::Purl::ResourceFile do
```

spec/models/embed/purl/resource_file_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+
+    it 'is the file id when no resource description is available' do
+      file = described_class.new(resource, resource_file, double('rights'))
+      expect(file.label).to eq 'The File ID'
+    end
+  end
+
+  describe '#thumbnail' do
+    let(:resource_with_thumb) do
+      double(
```
jcoyne, line 57: Can these be instance_double?

spec/models/embed/purl/resource_file_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+        'Purl::Resource', files: [
+          double(thumbnail?: false, title: 'Non thumb'),
+          double(thumbnail?: true, title: 'The Thumb')
+        ]
+      )
+    end
+    let(:resource_without_thumb) do
+      double(
```
jcoyne, line 65: Can these be instance_double?

spec/models/embed/purl/resource_file_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+      expect(f).to receive(:xpath).with('./*[@duration]').and_return([audio_data_el])
+      rf = described_class.new(double('Resource'), f, double('Rights'))
+      expect(Honeybadger).to receive(:notify).with("ResourceFile#media duration ISO8601::Errors::UnknownPattern: 'invalid'")
+      expect(Embed::MediaDuration).to receive(:new).and_call_original
+      expect(rf.duration).to be_nil
+    end
+  end
+
+  describe 'video_data' do
+    context 'valid videoData' do
```
jcoyne, line 268: Can you use `with valid videoData` and then remove the exception from .rubocop_todo?

.rubocop_todo.yml
```diff
@@ -88,11 +90,11 @@ RSpec/IteratedExpectation:
...
     - 'spec/models/embed/was_time_map_spec.rb'
 
-# Offense count: 56
-# Configuration parameters: .
+# Offense count: 67
+# Configuration parameters: EnforcedStyle.
 # SupportedStyles: have_received, receive
 RSpec/MessageSpies:
-  EnforcedStyle: receive
+  Enabled: false
```
jcoyne, line 97: Why are you changing this?

jcoyne (changes requested):

.rubocop_todo.yml
justinlittman: I just re-ran autogen.

jcoyne: please don't do that, at least not in this PR. It adds a bunch of additional changes that we have to review. Instead fix the errors or add inline exceptions.

justinlittman: The errors are because I moved tests around. The tests were not changed. This just reflects the movement of those tests, not the introduction of any new rubocop exceptions. Fixing the existing tests are beyond the scope of this ticket.

Gemfile
```diff
@@ -22,6 +22,7 @@ gem 'iso8601' # to parse durations, since ActiveSupport::Duration doesn't get a
 gem 'nokogiri', '>= 1.7.1'
 
 group :development, :test do
+  gem 'byebug'
```
jcoyne, line 25: Please remove byebug, the 'debug' gem is now included in Ruby 3.1+

jcoyne (changes requested):

jcoyne (approved):

andrewjbtw:
Based on further feedback from PSM, with hierarchy and search both deployed. Everyone was overall positive about where things are in the display. Nothing looks like a blocker.

The main thing that jumped out was the sequence numbering that appears to the left of each filename:

Screen Shot 2023-01-20 at 12 59 14 PM

That's part of the default display for flat lists of files, and it seems useful there at least for helping align the lists visually. But people found the numbering confusing in the hierarchy context and would prefer to drop it in that situation.

Is it possible to not show those sequence numbers when there is a hierarchy? The general consensus was that we shouldn't drop the number for flat lists, but it would be nice if we could for hierarchies. All of this with the understanding that we should come back to the sequence number issue in the future, as there might be another visual approach to flat lists that could replace it.

The other question that came up was how the order is being determined in the display. I see that files display before folders for each level of hierarchy. Is there any other sorting being applied in the viewer? The comments here were less about users being able to determine a sort order and more about understanding how the order is being determined. In the past, the display order followed the Cocina order. There may be requests for user-specified orders in the future, but in the shorter term we'd like to be able to explain to people how order is being set.

justinlittman:
@andrewjbtw Looking at the sul-embed code, I don't see any sorting applied (other than placing files before folders). I can't say whether there is sorting elsewhere in the stack.

Looking at removing the numbering.

andrewjbtw:
I’ve been cycling through a lot of objects and it’s looking good to me. @Astrid Usong is checking on accessibility.

The only thing I see that isn’t loading as expected is the file count. Sometimes it loads slowly or never loads. The upper limit seems to be around 7000 items.

- [7000 items](https://sul-purl-stage.stanford.edu/tb758mb4939) (loads in firefox and Safari, not Chrome)
- [8000 items](https://sul-purl-stage.stanford.edu/qp984hm2822) (no count in any browser)

Issue #1403, opened 2023-01-05
Migrate sassc to dart sass


jcoyne:
Because libsass is deprecated

PR #1414, opened 2023-01-11
Replace libsass with Dart Sass

jcoyne:
Fixes https://github.com/sul-dlss/sul-embed/issues/1403

Test on https://sul-purl-stage.stanford.edu/bb546qn3753


Rakefile
```diff
@@ -12,7 +12,10 @@ rescue LoadError
   # should only be here when gem group development and test aren't installed
 end
 
-task default: [:javascript_tests, :spec, :rubocop]
+# remove default rspec task
+task(:default).clear
```
thatbudakguy, line 16: was this causing issues?

thatbudakguy (approved):

Rakefile
jcoyne: Yes, if you don't clear it, it just appends to it rather than overwriting it.

Issue #1383, opened 2022-11-18
Migrate Universal Analytics (UA) to Google Analytics 4 (GA4)


corylown:
Universal Analytics (UA) properties will [stop collecting data as of July 1, 2023](https://support.google.com/analytics/answer/11583528?hl=en). The data in the existing UA property will remain available for an unspecified period of time. In order to collect data after July 1, 2023 we will need to create a new GA4 property and add the newer `gtag.js` snippet to the application in order to collect data into the new GA4 property.

Spreadsheet that tracks properties and whether each property is tracking events: https://docs.google.com/spreadsheets/d/1WWt3l-RFqfb_L2j01Y7JBUWJ2OnX0Hi4xLmIW8abNLw/edit?usp=sharing

- [x] [Add a Google Analytics 4 property](https://support.google.com/analytics/answer/9744165#zippy=%2Cin-this-article)
- [x] [Add the Google tag to the application](https://support.google.com/analytics/answer/9304153#zippy=%2Cadd-your-tag-using-google-tag-manager%2Cadd-the-google-tag-directly-to-your-web-pages)
- [x] [Set up event tracking as desired](https://support.google.com/analytics/answer/11091025#zippy=%2Cin-this-article). NOTE: _everything_ is an event in GA4 and it automatically collects some events.
- [ ] Optionally, download and archive the historical UA data.

NOTE: There is currently a [setting available](https://support.google.com/analytics/answer/11150547) that will send events from `analytics.js` (UA) to GA4 properties. However, Google does not guarantee this will continue to work after July 1, 2023. This is not recommended as a migration path for event tracking.

marlo-longley:
Note: the events being tracked on the existing UA property are all Mirador events, e.g. `mirador/SET_CANVAS`

marlo-longley:
This is the new embed code: 
```

<!-- Google tag (gtag.js) -->
<script async src="https://www.googletagmanager.com/gtag/js?id=G-80J719XWE5"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'G-80J719XWE5');
</script>
```

marlo-longley:
@corylown -- do you know where we stand on your optional step of archiving the old analytics? This could be closed by #1417. I marked the Migrated? column to YES in the spreadsheet as well.

corylown:
[...]

PR #1417, opened 2023-01-11
Migrate Google Analytics to GA4

marlo-longley:
Ref https://github.com/sul-dlss/sul-embed/issues/1383

Following the pattern of other recent GA4 implementations, I put the Google identifier directly into the code.
Debug mode will remain on unless on the production server (see https://github.com/sul-dlss/shared_configs/pull/1881)

This code successfully translates the existing custom Mirador events from UA to GA4, as verified in the GA4 admin panel (debug mode):
Screenshot 2023-01-11 at 18 31 47

It also sends the automatically-collected new GA4 events (see https://support.google.com/analytics/answer/9234069?hl=en)

I found it weird to put the script between `<head>` and `<body>` but that is what the migration [guide](https://support.google.com/analytics/answer/9744165?hl=en#zippy=%2Cin-this-article%2Cinstall-a-google-tag) seemed to say!

> copy and paste your entire Google tag in the code of every page of your website, immediately after the head element.

### DIFF
--- a/app/assets/javascripts/plugins/analyticsPlugin.js
+++ b/app/assets/javascripts/plugins/analyticsPlugin.js
@@ -7,7 +7,6 @@ import { getManifest, getCompanionWindow } from 'mirador/dist/es/src/state/selec
 function* onSetCanvas({ type, windowId, canvasId }) {
   const { id: manifestId } = yield select(getManifest, { windowId });
   yield put({ type: 'mirador/analytics', payload: {
-    hitType: 'event',
     eventCategory: manifestId,
     eventAction: type,
     eventLabel: canvasId,
@@ -16,7 +15,6 @@ function* onSetCanvas({ type, windowId, canvasId }) {
 
 function* onAddResource({ type, manifestId }) {
   yield put({ type: 'mirador/analytics', payload: {
-    hitType: 'event',
     eventCategory: manifestId,
     eventAction: type,
   }});
@@ -28,7 +26,6 @@ function* onAddCompanionWindow({ payload: { position, content }, type, windowId 
   const { id: manifestId } = yield select(getManifest, { windowId });
 
   yield put({ type: 'mirador/analytics', payload: {
-    hitType: 'event',
     eventCategory: manifestId,
     eventAction: type,
     eventLabel: `${content}/${position}`,
@@ -42,7 +39,6 @@ function* onUpdateCompanionWindow({ id, payload: { position, content }, type, wi
   const { content: existingContent } = yield select(getCompanionWindow, { companionWindowId: id });
 
   yield put({ type: 'mirador/analytics', payload: {
-    hitType: 'event',
     eventCategory: manifestId,
     eventAction: type,
     eventLabel: `${content || existingContent}/${position}`,
@@ -52,7 +48,6 @@ function* onUpdateCompanionWindow({ id, payload: { position, content }, type, wi
 
 function* onReceiveManifest({ manifestId, type }) {
   yield put({ type: 'mirador/analytics', payload: {
-    hitType: 'event',
     eventCategory: manifestId,
     eventAction: type,
   }});
@@ -62,7 +57,6 @@ function* onRequestSearch({ query, type, windowId }) {
   const { id: manifestId } = yield select(getManifest, { windowId });
 
   yield put({ type: 'mirador/analytics', payload: {
-    hitType: 'event',
     eventCategory: manifestId,
     eventAction: type,
     eventLabel: query,
@@ -71,7 +65,6 @@ function* onRequestSearch({ query, type, windowId }) {
 
 function* onAddWindow({ type, window: { canvasId, manifestId } }) {
   yield put({ type: 'mirador/analytics', payload: {
-    hitType: 'event',
     eventCategory: manifestId,
     eventAction: type,
     eventLabel: canvasId,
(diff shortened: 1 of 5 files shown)


config/settings.yml
```diff
@@ -61,4 +61,6 @@ use_custom_pdf_viewer: true
 use_custom_3d_viewer: true
 collections_to_show_attribution:
   - dc876nn4469
-analytics_id:
+# Google Analytics will report in debug mode by default
+# in shared_configs for embed-prod, we set this value to 0  
+analytics_debug: 1
```
jcoyne, line 66: Can we keep the default as 0, and set it to 1 where we need to debug?

jcoyne (changes requested):

config/settings.yml
marlo-longley: @jcoyne I guess the logic here is a "safety" style logic -- we don't want to mistakenly send data to our real analytics, so better to default to debug mode and explicitly set production.

jcoyne (approved):

marlo-longley:
@jcoyne thanks. I actually just noticed something I need to amend here -- reverting to draft.

marlo-longley:
@jcoyne fixed up -- I forgot to migrate a custom dimension for the item's parent collection. Working now:
Screenshot 2023-01-12 at 14 37 52

app/assets/javascripts/plugins/analyticsPlugin.js
```diff
@@ -7,18 +7,16 @@ import { getManifest, getCompanionWindow } from 'mirador/dist/es/src/state/selec
 function* onSetCanvas({ type, windowId, canvasId }) {
   const { id: manifestId } = yield select(getManifest, { windowId });
   yield put({ type: 'mirador/analytics', payload: {
-    hitType: 'event',
-    eventCategory: manifestId,
-    eventAction: type,
-    eventLabel: canvasId,
+    event_category: manifestId,
```
jcoyne, line 10: not a huge concern, but aren't parameters in javascript typically camelcased (as they were before), why did you migrate away from this convention?

marlo-longley: This is the syntax used in all the official Google guides, e.g. https://developers.google.com/analytics/devguides/migration/ua/analyticsjs-to-gtagjs#measure_events

jcoyne: I only see that in the parameters to the `gtag` call.  I don't think we need to change all the other parameters in this plugin.

marlo-longley: Ah I see what you mean. I can change that back, I guess keeping them uniform just was cleaner in my mind.

jcoyne: 🤷

marlo-longley: I'll change it

app/views/embed/_analytics.html.erb
```diff
@@ -0,0 +1,12 @@
+<script async src="https://www.googletagmanager.com/gtag/js?id=G-80J719XWE5"></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){dataLayer.push(arguments);}
+  gtag('js', new Date());
+  gtag('config', 'G-80J719XWE5', {
+    'debug_mode': Boolean('<%= Settings.analytics_debug %>'),
+    // 'collection' is a custom event parameter for getting an item's parent collection, if applicable
```
jcoyne, line 8: How about putting this in a erb comment (e.g. `<%#  like this %>`) so that we don't print it out to the HTML source?

config/settings.yml
jcoyne, line 66: Is there a reason this isn't `true`?  Why do we configure this as a integer?

marlo-longley: Yes when we implemented this same debug flag in VT we ran into an issue and solved it using integer, so I went with the same thing here.

jcoyne: What issue?  It seems unnecessarily complex that we have to use javascript to convert a string to a boolean to handle this, when the config gem itself can do this for us.

jcoyne (changes requested):

config/settings.yml
marlo-longley: @jcoyne I don't know a good way to store this in a setting/config file for use in JS. Sure we can store it in in the config file as a Bool but in JS it's going to load as a string. It seems more complex to do logic like if "true" then ... -- than to just store a 1 or 0. Maybe I'm missing something obvious tho.

marlo-longley: Boolean("false") in JS still evaluates to true -- 0/1 seemed safer

marlo-longley: Boolean("false") in js still evaluates to true -- 1/0 seemed safer

jcoyne: Why do you need to wrap it in Boolean? just render the boolean in the JS.

marlo-longley: It looks like I made an error here, but I still think we need to use Boolean.

`typeof('<%= Settings.analytics_debug %>'))` is string whether the value is "false", false, or 0.
`typeof(Number('<%= Settings.analytics_debug %>')))` is number when the value is 0.
Boolean("false") === true
Boolean(0) === false

So -- I think we need `Boolean(Number('<%= Settings.analytics_debug %>')))` to actually get a false into JS.
I agree this is complex. I don't know a better approach if we want to set the debug value out of shared_configs/settings. Maybe setting another way is better.

marlo-longley: @jcoyne you were right -- I was forcing it to a string by mistake. Thank you for catching that. Pushed a fix. The values in settings can be true or false, not integers.

jcoyne (approved):

PR #1418, opened 2023-01-12
[WIP] Remove pdfjs and use the native pdf viewer

jcoyne:
We don't need to ship a PDF library as this is now included in all browsers.
Test with:
https://embed-stage.stanford.edu/iframe?url=https://sul-purl-stage.stanford.edu/rr785jw7640
or https://sul-purl-stage.stanford.edu/cv071xh8846

Need to get buy in from stakeholders.


cbeer:
I wonder whether we're concerned about:

- inconsistent behavior, controls, and features in different browsers
- some level of consistency between the different sul-embed viewers; this seems to get rid of the sul-embed chrome with various controls (download, attribution, etc)
- whether the native chrome (which seems.. chunky) is appropriate in an embedded context
- how (or whether?) the native chrome handles stanford-only, location-restricted or dark content

lwrubel:
@thatbudakguy will talk with @cbeer about this to determine next steps if any.

PR #1420, opened 2023-01-13
Add download-all for File objects

### LINKED ISSUE #1381: User can download all files in an object with a single click
**Request**

Add a download button to the bottom of the the embed viewer when:

- content type is file
- rights allow download

The button should provide access to the download all files API: https://api.library.stanford.edu/docs/digital-stacks/api#downloading-all-files-for-a-given-object

Do not show a download icon when rights do not allow download.

**Design**

image

Notes on the design

- Ideally, we would give people an indication of the total download size before they click the button. If that's not possible, just showing a download icon would be acceptable.
- Only show the Stanford 'S' icon if rights are restricted to "Stanford + download"
  - The 'S' icon should be the same icon used when single files are restricted to Stanford ([example](https://purl.stanford.edu/bb057yp4089))
  - If rights are restricted to Stanford, the link should trigger SUnet authentication. That might need an additional flow but if possible it should follow what happens when people click on a Stanford-restricted single file.

If possible, once the button is clicked, show an in-progress indicator and disable the button to prevent multiple clicks in a short period:

image (2)

Because the download starts being built in the background before the browser initiates download, there's a concern that users will repeatedly click the button if they don't realize the system is already working on the request.

**Additional considerations**:
- If one file within an object has different access rights (e.g. dark) than others, do not show the "download all" button. Handling cases where file-level rights vary within the same object will require more analysis.
- We are not placing a limit on size of download, either in terms of number of files or size of files. However, it's likely that there is an upper limit to what the system will handle. This may need to be revisited but there are no current limits on the size of individual file downloads and policy concerning maximum download size requires further discussion.
- At the risk of doing a premature optimization we may want to limit how much crawlers will discover and follow links to stacks.stanford.edu to generate the ZIP files since this the rapid, dynamic generation of ZIP files could create some performance issues.

**Additional context**
Objects with many files can take a lot of clicks to download all files, and this will become more of an issue with hierarchical objects with many files.  Users should be able to get all of the files in an object with a single click. This would also benefit the user experience for already existing content that is presented in long file lists in sul-embed.

Some previous work design work for image viewer here: #1187

edsu:
Changes to add Download All button for objects of type File.

Closes #1381

### DIFF
--- a/app/assets/javascripts/modules/download_all.js
+++ b/app/assets/javascripts/modules/download_all.js
@@ -1,0 +1,39 @@
+// This module ensures Download All button doesn't get clicked repeatedly. The first
+// click on Download button will cause the button text to be changed to
+// "Initializing download" and disabled, then the download will start.
+//
+// An <a> element is used to enclose the <button> because Chrome and
+// Stanford's SingleSignOn don't seem to like creating <a> tags dynamically and
+// then clicking on them--maybe this a security feature? So the first click
+// will also replace the <a> element in the DOM with the <button> which will
+// prevent further clicks on the <a> since it will no longer exist.
+
+(function(global) {
+  var Module = (function() {
+    return {
+
+      init: function() {
+        const a = document.getElementById('sul-embed-footer-download-all');
+        if (a) {
+          a.addEventListener('click', e => this.download(e));
+        }
+      },
+
+      download: function(event) {
+        const el = event.target;
+        if (el.localName == 'button') {
+          el.disabled = true;
+          el.innerText = " Initializing download...";
+
+          // replace the anchor tag with the button to prevent further clicks
+          const a = el.parentElement;
+          a.replaceWith(el);
+        }
+      }
+
+    }
+  })();
+
+  global.DownloadAll = Module;
+
+})(this);
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -110,6 +110,7 @@ body {
     .#{$namespace}-footer-toolbar {
       padding-top: 0;
       vertical-align: middle;
+      color: $white-color;
     }
 
     .sul-i-download-3 {
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 module Embed
+  # rubocop:disable Metrics/ClassLength
   class Purl
     require 'embed/media_duration'
     require 'dor/rights_auth'
@@ -39,6 +40,10 @@ module Embed
 
     def hierarchical?
       @hierarchical ||= hierarchical_contents.files.length != contents.sum { |resource| resource.files.length }
+    end
+
+    def size
+      contents.sum(&:size)
     end
 
     def rights
@@ -167,6 +172,7 @@ module Embed
           nil
         end
     end
+    # rubocop:enable Metrics/ClassLength
 
     class ResourceNotAvailable < StandardError
       def initialize(msg = 'The requested PURL resource was not available')
--- a/app/views/embed/footer/_generic.html.erb
+++ b/app/views/embed/footer/_generic.html.erb
@@ -10,7 +10,6 @@
         data-sul-embed-toggle= 'sul-embed-embed-this-panel'></button>
     <% end %>
 
-
     <% if viewer.show_download? %>
       <button
         class='sul-embed-footer-tool sul-embed-btn sul-em bed-btn-toolbar sul-embed-btn-default sul-i-download-3'
--- a/spec/models/embed/purl_spec.rb
+++ b/spec/models/embed/purl_spec.rb
@@ -99,6 +99,18 @@ describe Embed::Purl do
     end
   end
 
+  describe '#size' do
+    subject { purl.size }
+
+    before do
+      stub_purl_response_with_fixture(multi_file_purl)
+    end
+
+    let(:purl) { described_class.new('12345') }
+
+    it { is_expected.to eq 12_345 * 2 }
+  end
+
   describe '#downloadable_files' do
     it 'returns a flattened array of downloadable resource files' do
       stub_purl_response_with_fixture(multi_resource_multi_type_purl)
--- a/app/assets/javascripts/file.js
+++ b/app/assets/javascripts/file.js
@@ -6,6 +6,7 @@
 //= require modules/file_preview
 //= require modules/popup_panels
 //= require modules/embed_this
+//= require modules/download_all
 
 CssInjection.injectFontIcons();
 CssInjection.appendToHead();
@@ -13,3 +14,4 @@ CommonViewerBehavior.initializeViewer();
 FilePreview.init();
 PopupPanels.init();
 EmbedThis.init();
+DownloadAll.init();
--- a/app/assets/stylesheets/file.scss
+++ b/app/assets/stylesheets/file.scss
@@ -45,6 +45,18 @@ body {
     max-width: 100%;
   }
 
+  .#{$namespace}-footer-toolbar .sul-i-download-3 {
+    padding: 0px 5px 0px 5px;
+  }
+
+  .#{$namespace}-footer-toolbar .sul-embed-location-restricted-text {
+    background-color: $white-color;
+  }
+
+  .#{$namespace}-footer-toolbar a {
+    color: $white-color;
+  }
+
   .#{$namespace}-download {
     display: inline;
     font-size: .9em;
--- a/app/models/embed/purl/resource.rb
+++ b/app/models/embed/purl/resource.rb
@@ -38,6 +38,10 @@ module Embed
         end
       end
 
+      def size
+        files.sum(&:size)
+      end
+
       def primary_file
         files.find(&:primary?)
       end
--- a/app/viewers/embed/viewer/file.rb
+++ b/app/viewers/embed/viewer/file.rb
@@ -61,6 +61,25 @@ module Embed
         super || display_file_search?
       end
 
+      # Returns true or false whether the viewer should display the Download All
+      # link. The limits were determined in testing and may need to be adjusted
+      # based on experience with download performance and any changes in the
+      # Stacks API. It returns false when there is just one file because the
+      # file download link will suffice for that.
+      def display_download_all?
+        @purl_object.size < 10_737_418_240 &&
+          @purl_object.downloadable_files.length > 1 &&
+          @purl_object.downloadable_files.length < 3000
+      end
+
+      def any_stanford_only_files?
+        @purl_object.all_resource_files.any?(&:stanford_only?)
+      end
+
+      def download_url
+        "#{Settings.stacks_url}/object/#{@purl_object.druid}"
+      end
+
       private
 
       def default_height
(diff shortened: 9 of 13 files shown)


app/models/embed/purl.rb
```diff
@@ -208,6 +216,10 @@ def files
         end
       end
 
+      def size
```
jcoyne, line 219: There are two methods with this name? https://github.com/sul-dlss/sul-embed/pull/1420/files#diff-804f518d2a4799fc0c48e440678aa911984edb69a40efbb0a097a99392b875a8R36

jcoyne: NM, this is a Resource class.

app/models/embed/purl.rb
```diff
@@ -208,6 +224,11 @@ def files
...
+        # need to guard against file.size being nil when summing
+        files.filter_map { |file| file.size unless file.size.nil? }.sum
```
mjgiarlo, line 229: I *think* the `unless file.size.nil?` is not needed here; `#filter_map` will automatically drop all `nil` values.

edsu: yep, works! thanks 🙇

.rubocop_todo.yml
```diff
@@ -26,7 +26,7 @@ Metrics/AbcSize:
 # Offense count: 3
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 130
+  Max: 135
```
jcoyne, line 29: Please don't make this change globally as this allows every file to get worse.  Instead, add an inline exception only on the file that requires this.

app/models/embed/purl.rb
```diff
@@ -266,8 +287,10 @@ def image?
           mimetype =~ %r{image/jp2}i
         end
 
+        # Return the size of the file in bytes as an integer or nil if it is not present.
```
jcoyne, line 290: How about we never return `nil`? That means we don't have to do any guarding?

app/models/embed/purl.rb
```diff
@@ -266,8 +287,10 @@ def image?
...
         def size
-          @file.attributes['size'].try(:value)
+          bytes = @file.attributes['size'].try(:value)
+          bytes.blank? ? nil : bytes.to_i
```
jcoyne, line 293:
```suggestion
          bytes.to_i
```

both `nil.to_i` and `''.to_i` both return 0

app/models/embed/purl.rb
```diff
@@ -208,6 +224,11 @@ def files
         end
       end
 
+      def size
+        # need to guard against file.size being nil when summing
```
jcoyne, line 228: Let's not guard.

jcoyne (changes requested):

app/models/embed/purl.rb
edsu: I think there are cases where there actually is no value, which some of the logic depends on. I initially had it return 0 when it was undefined and some tests broke.

app/models/embed/purl.rb
edsu: See discussion of zero vs nil above.

app/models/embed/purl.rb
jcoyne: Why is that important?  Where do we use that distinction? It looks to me like we have to go out of our way and add complexity to the code to accomodate this distinction without any user benefit.

edsu: I changed size to return an integer rather than a string. There was a test which broke when I started returning 0 instead of nil. Would it be helpful to know which test that was?

jcoyne: Yes.

edsu: These three tests failed when I simply changed `ResourceFile.size` to return an integer:

```
rspec ./spec/models/embed/purl_spec.rb:268 # Embed::Purl Purl::Resource Purl::Resource::ResourceFile attributes gets the size from the size attribute
rspec ./spec/views/embed/template/_file.html.erb_spec.rb:77 # embed/template/_file without file size displays Download as the link text
rspec ./spec/views/embed/template/_file.html.erb_spec.rb:88 # embed/template/_file with empty file size displays Download as the link text
```

jcoyne: Let's do this then. https://github.com/sul-dlss/sul-embed/pull/1426 I think it's more important to us to end this work with simple, maintainable code then it is to value changing the fewest number of lines.

.rubocop_todo.yml
jcoyne: Can you please revert this change?

jcoyne (changes requested):

app/assets/stylesheets/common.scss
```diff
@@ -110,13 +110,22 @@ body {
...
       margin-right: 1px;
       padding-right: 1px;
     }
 
+    .#{$namespace}-footer-toolbar .sul-i-download-3 {
+      padding: 0px 5px 0px 5px;
+    }
+
+    .#{$namespace}-footer-toolbar .sul-embed-location-restricted-text {
+      background-color: white;
```
jcoyne, line 126: Why `white` here and `$white-color` on 113?  Can  you make these consistent?

app/views/embed/footer/_generic.html.erb
```diff
@@ -10,8 +10,21 @@
         data-sul-embed-toggle= 'sul-embed-embed-this-panel'></button>
     <% end %>
 
-
-    <% if viewer.show_download? %>
+    <% if viewer.show_download_all? && viewer.purl_object.size > 1073741824 %>
```
jcoyne, line 13: Can you put this in `app/views/footer/_file.html.erb` and then we can remove the `show_download_all?` method?    This will also require changing this line: https://github.com/sul-dlss/sul-embed/blob/305b3c12924a3de19040e8c96f3335e9e26ce013/app/views/embed/template/_file.html.erb#L6

jcoyne (changes requested):

app/views/embed/footer/_generic.html.erb
edsu: I thought about doing this too, will it be ok to duplicate the rest of `footer/_generic.html.erb` in`footer/_file.html.erb`?

jcoyne: Yes, that's fine. If you wanted to make it really dry, you could extract the common parts out to a template included by both.

.rubocop_todo.yml
```diff
@@ -27,6 +27,9 @@ Metrics/AbcSize:
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
   Max: 130
+  Exclude:
+    - 'app/models/embed/purl.rb'
```
jcoyne, line 31: Can you move this to an inline exception in the file itself?  Otherwise this will get overwritten if someone regenerates this file.

jcoyne (changes requested):

spec/models/embed/purl_spec.rb
```diff
@@ -99,6 +99,13 @@
     end
   end
 
+  describe 'size' do
```
jcoyne, line 102: Can you extract:
```ruby
let(:purl) { described_class.new('12345') }
subject { purl.size } 
before do
  stub_purl_response_with_fixture(multi_file_purl)
end
it { is_expected.to eq 12_345 * 2 }
```

On these? I think that helps the reader because it highlights the subject of the test and separates out the setup for the test.

spec/models/embed/purl_spec.rb
```diff
@@ -123,6 +130,37 @@
...
+    end
+
+    it 'returns false when one file is not downloadable' do
+      stub_purl_response_with_fixture(world_restricted_download_purl)
+      expect(described_class.new('12345').all_files_downloadable?).to be false
+    end
+  end
+
+  describe 'any_stanford_only_files' do
+    it 'returns true when one or more files are stanford only' do
```
jcoyne, line 146: `when one or more files are stanford only` is a context so you should use the rspec `context` method to define those.

app/views/embed/footer/_generic.html.erb
edsu: I went with splitting off the template for now, without the extra template complexity.

app/assets/javascripts/modules/download_all.js
```diff
@@ -0,0 +1,34 @@
...
+
+(function(global) {
+  var Module = (function() {
+    return {
+
+      init: function() {
+        this.waitForClick();
+      },
+
+      waitForClick: function() {
```
jcoyne, line 13: Is there a reason this can't be merged into the `init()` function?

app/models/embed/purl.rb
```diff
@@ -69,10 +74,22 @@ def downloadable_files
       end
     end
 
+    def all_files_downloadable?
```
jcoyne, line 77: Is this method needed?  I'm not seeing where it is used.

jcoyne (changes requested):

app/assets/stylesheets/common.scss
```diff
@@ -110,13 +110,22 @@ body {
...
       vertical-align: middle;
+      color: $white-color;
     }
 
     .sul-i-download-3 {
       margin-right: 1px;
       padding-right: 1px;
     }
 
+    .#{$namespace}-footer-toolbar .sul-i-download-3 {
```
jcoyne, line 121: Can these changes be moved to https://github.com/sul-dlss/sul-embed/blob/master/app/assets/stylesheets/file.scss ?

app/viewers/embed/viewer/common_viewer.rb
```diff
@@ -66,6 +66,10 @@ def show_download?
         self.class.show_download? && !@request.hide_download?
       end
 
+      def show_download_all?
```
jcoyne, line 69: Is this method used?

jcoyne (changes requested):

app/assets/javascripts/modules/download_all.js
edsu: Not that I'm aware of.

app/viewers/embed/viewer/common_viewer.rb
edsu: No, thanks for noticing!

app/models/embed/purl.rb
edsu: Ah good catch, this is no longer used/needed after https://github.com/sul-dlss/stacks/pull/950

spec/models/embed/purl_spec.rb
edsu: Thanks for the tip.

spec/models/embed/purl_spec.rb
edsu: @jcoyne I think I understand but it would help me to walk through this one with you if you are available Weds AM (EST)?

edsu: @jcoyne I've updated the setup for these tests--can you take a look and see if I did what you were expecting?

jcoyne (changes requested):

jcoyne (approved):

jcoyne:
@edsu it seems like rubocop is failing.

edsu:
This is now waiting for feedback from @andrewjbtw about whether the behavior on mobile is acceptable or if we want to make some more changes.

PR #1426, opened 2023-01-19
Make the size method always return an integer

jcoyne:


edsu:
I've merged this into #1420 so we can close this if you want. I'm not entirely sure why the unit tests pass for #1420  but not here though.

jcoyne:
@edsu that's a known flappy spec.

corylown (approved):

PR #1431, opened 2023-01-20
Search for file viewer

justinlittman:
https://user-images.githubusercontent.com/588335/213778263-5dbcaf1b-f96f-48a5-99e5-72850315069c.mov

### DIFF
--- a/app/assets/javascripts/tree.js
+++ b/app/assets/javascripts/tree.js
@@ -12,20 +12,9 @@
 
 'use strict';
 
-/**
- * ARIA Treeview example
- *
- * @function onload
- * @description  after page has loaded initialize all treeitems based on the role=treeitem
- */
-
 window.addEventListener('load', function () {
-  var trees = document.querySelectorAll('[role="tree"]');
-
-  for (var i = 0; i < trees.length; i++) {
-    var t = new Tree(trees[i]);
-    t.init();
-  }
+  var treeNodes = document.querySelectorAll('[role="tree"]');
+  treeNodes.forEach((treeNode) => new Tree(treeNode).init());
 });
 
 /*
@@ -48,11 +37,13 @@ var Tree = function (node) {
   this.domNode = node;
 
   this.treeitems = [];
+  this.fileTreeitems = [];
   this.firstChars = [];
 
   this.firstTreeitem = null;
   this.lastTreeitem = null;
   this.selectedItem = null;
+  this.searchInput = null;
 };
 
 Tree.prototype.init = function () {
@@ -67,6 +58,7 @@ Tree.prototype.init = function () {
         ti.init();
         tree.treeitems.push(ti);
         tree.firstChars.push(ti.label.substring(0, 1).toLowerCase());
+        if(ti.isSearchable) tree.fileTreeitems.push(ti);
       }
 
       if (elem.firstElementChild) {
@@ -83,8 +75,10 @@ Tree.prototype.init = function () {
   }
 
   findTreeitems(this.domNode, this, false);
-
   this.updateVisibleTreeitems();
+  this.updateFileCount();
+  this.searchInput = document.querySelector('[data-search-input]')
+  if(this.searchInput) this.searchInput.addEventListener('keyup', this.updateSearch.bind(this));
 
   this.firstTreeitem.domNode.tabIndex = 0;
 };
@@ -216,6 +210,19 @@ Tree.prototype.updateVisibleTreeitems = function () {
   }
 };
 
+Tree.prototype.updateFileCount = function () {
+  const count = this.fileTreeitems.filter(fileTreeItem => fileTreeItem.isSearchVisible).length;
+  const count_text = `${count} ${count == 1 ? "item" : "items"}`;
+  document.querySelectorAll('.sul-embed-item-count').forEach((elem) => elem.innerHTML = count_text);
+};
+
+Tree.prototype.updateSearch = function () {
+  const search = this.searchInput.value.toLowerCase();
+  this.fileTreeitems.forEach(ti => ti.updateSearch(search));
+  this.updateFileCount();
+  this.treeitems.forEach(ti => ti.updateSearch(search));
+};
+
 Tree.prototype.setFocusByFirstCharacter = function (currentItem, char) {
   var start, index;
 
@@ -323,6 +330,10 @@ var Treeitem = function (node, treeObj, group) {
     RIGHT: 39,
     DOWN: 40,
   });
+
+  this.isSearchable = node.dataset.hierarchyType === 'file';
+  this.isSearchVisible = true;
+  this.searchText = null;
 };
 
 Treeitem.prototype.init = function () {
@@ -341,6 +352,12 @@ Treeitem.prototype.init = function () {
     this.domNode.addEventListener('mouseover', this.handleMouseOver.bind(this));
     this.domNode.addEventListener('mouseout', this.handleMouseOut.bind(this));
   }
+
+  if(this.isSearchable) {
+    const filepath = this.domNode.querySelector('[data-filepath]').dataset.filepath.toLowerCase();
+    const description = this.domNode.querySelector('[data-description]').textContent.toLowerCase();
+    this.searchText = [filepath, description].join('')
+  }
 };
 
 Treeitem.prototype.isExpanded = function () {
@@ -349,6 +366,17 @@ Treeitem.prototype.isExpanded = function () {
   }
 
   return false;
+};
+
+Treeitem.prototype.updateSearch = function (search) {
+  if(this.isSearchable && search != undefined) {    
+    this.isSearchVisible = this.searchText.includes(search);
+    this.domNode.setAttribute('aria-hidden', !this.isSearchVisible);
+  }
+  if(!this.isSearchable) {
+    const hasVisibleDescFiles = this.domNode.querySelectorAll('[data-hierarchy-type="file"][aria-hidden="false"]').length === 0
+    this.domNode.setAttribute('aria-hidden', hasVisibleDescFiles);  
+  }
 };
 
 /* EVENT HANDLERS */
--- a/app/views/embed/template/_file.html.erb
+++ b/app/views/embed/template/_file.html.erb
@@ -4,4 +4,5 @@
   <%= render 'metadata_panel', viewer: viewer %>
   <%= render 'embed_this/file', viewer: viewer %>
   <%= render 'footer/generic', viewer: viewer %>
- </div>
+  <script src="<%= asset_url('tree.js') %>"></script>
+</div>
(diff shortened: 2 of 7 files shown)


andrewjbtw:
Results of some preliminary testing:

- filename filtering looks to be working as expected
- the search is matching folder names so that if a folder name matches, all files in the folder display
- the search is not matching the text in the file labels
  - to replicate, search for "object" on this item: https://sul-purl-stage.stanford.edu/sx562mk0566

I should be able to get some feedback on the search behavior for folders. I don't have a strong opinion on the folder name matches. It seems like a nice thing to have and you can collapse the folder in the filtered view.

justinlittman:
I included the folder name based on my experience with Github file searches (where it is helpful). I could add the labels if desired.

andrewjbtw:
I just finished reviewing with PSM:

- positive feedback on including folders, so let's keep it
- please add labels to the search

One additional bit of feedback on the layout: when the number of items is large, the text starts to encroach on the search box:

Screen Shot 2023-01-20 at 12 24 22 PM

I don't know if it would be better to dynamically move it to the right depending on the text, or to shift it based on some expected maximum length, like no more than 1 million files. Might be a question for @astridu?

astridu:
I vote for dynamically moving it to the right, OR just moving it to the right about an inch. In cases where there aren't that many items, there may be a gap, but that's ok. 
image

app/views/embed/template/_file.html.erb
```diff
@@ -4,4 +4,16 @@
   <%= render 'metadata_panel', viewer: viewer %>
   <%= render 'embed_this/file', viewer: viewer %>
   <%= render 'footer/generic', viewer: viewer %>
- </div>
+  <script src="<%= asset_url('tree.js') %>"></script>
+  <script>
+    document.onreadystatechange = () => {
+      if (document.readyState === "complete") {
```
jcoyne, line 10: was it determined that `addEventListener('load', () => ...` would definitely not work here?

spec/features/file_search_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'file viewer search bar', js: true do
```
jcoyne, line 5:
```suggestion
RSpec.describe 'file viewer search bar', js: true do
```

spec/features/file_search_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  include PurlFixtures
+  it 'limits shown files when text is entered' do
```
jcoyne, line 7: Please use context blocks for each of these e.g. `context 'when text is entered' ...`

app/assets/javascripts/tree.js
```diff
@@ -341,6 +353,12 @@ Treeitem.prototype.init = function () {
     this.domNode.addEventListener('mouseover', this.handleMouseOver.bind(this));
     this.domNode.addEventListener('mouseout', this.handleMouseOut.bind(this));
   }
+
+  if(this.isSearchable) {
+    var filepath = this.domNode.querySelector('[data-filepath]').getAttribute('data-filepath').toLowerCase();
```
jcoyne, line 358:
```suggestion
    const filepath = this.domNode.querySelector('[data-filepath]').dataset.filepath.toLowerCase();
```

app/assets/javascripts/tree.js
```diff
@@ -341,6 +353,12 @@ Treeitem.prototype.init = function () {
...
+    var description = this.domNode.querySelector('.sul-embed-description').textContent.toLowerCase();
```
jcoyne, line 359:
```suggestion
    const description = this.domNode.querySelector('.sul-embed-description').textContent.toLowerCase();
```

jcoyne: Could you please add a data selector on this node, rather than using a class?

app/assets/javascripts/tree.js
```diff
@@ -323,6 +331,10 @@ var Treeitem = function (node, treeObj, group) {
     RIGHT: 39,
     DOWN: 40,
   });
+
+  this.isSearchable = node.getAttribute('data-hierarchy-type') === 'file';
```
jcoyne, line 335:
```suggestion
  this.isSearchable = node.dataset.hierarchyType === 'file';
```

app/assets/javascripts/tree.js
```diff
@@ -83,8 +70,10 @@ Tree.prototype.init = function () {
   }
 
   findTreeitems(this.domNode, this, false);
-
   this.updateVisibleTreeitems();
+  this.updateFileCount();
+  this.searchInput = document.getElementById('sul-embed-search-input');
```
jcoyne, line 75: Please use a data attribute for this selector rather than using an id. This makes it easier to understand that the node is bound to some javascript.

app/assets/javascripts/tree.js
```diff
@@ -216,6 +205,25 @@ Tree.prototype.updateVisibleTreeitems = function () {
   }
 };
 
+Tree.prototype.updateFileCount = function () {
+  var count = this.fileTreeitems.filter(fileTreeItem => fileTreeItem.isSearchVisible).length;
```
jcoyne, line 209:
```suggestion
  const count = this.fileTreeitems.filter(fileTreeItem => fileTreeItem.isSearchVisible).length;
```

app/assets/javascripts/tree.js
```diff
@@ -216,6 +205,25 @@ Tree.prototype.updateVisibleTreeitems = function () {
...
+  var count_text = `${count} ${count == 1 ? "item" : "items"}`;
```
jcoyne, line 210:
```suggestion
  const count_text = `${count} ${count == 1 ? "item" : "items"}`;
```

app/assets/javascripts/tree.js
```diff
@@ -216,6 +205,25 @@ Tree.prototype.updateVisibleTreeitems = function () {
...
+  document.querySelectorAll('.sul-embed-item-count').forEach((elem) => elem.innerHTML = count_text);
+};
+
+Tree.prototype.updateSearch = function () {
+  const search = this.searchInput.value.toLowerCase();
+  for (var i = 0; i < this.fileTreeitems.length; i++) {
```
jcoyne, line 216: Can you use `this.fileTreeitems.forEach((ti) => { })` here?

app/assets/javascripts/tree.js
```diff
@@ -216,6 +205,25 @@ Tree.prototype.updateVisibleTreeitems = function () {
...
+    var ti = this.fileTreeitems[i];
+    ti.updateSearch(search);
+  }
+  this.updateFileCount();
+  for (var i = 0; i < this.treeitems.length; i++) {
```
jcoyne, line 221: consider using `this.fileTreeitems.forEach((ti) => { })`

app/assets/javascripts/tree.js
```diff
@@ -351,6 +369,17 @@ Treeitem.prototype.isExpanded = function () {
   return false;
 };
 
+Treeitem.prototype.updateSearch = function (search) {
+  if(this.isSearchable && search != undefined) {    
+    this.isSearchVisible = this.searchText.includes(search);
+    this.domNode.setAttribute('aria-hidden', !this.isSearchVisible);
+  }
+  if(!this.isSearchable) {
+    var hasVisibleDescFiles = this.domNode.querySelectorAll('[data-hierarchy-type="file"][aria-hidden="false"]').length == 0
```
jcoyne, line 378:
```suggestion
    const hasVisibleDescFiles = this.domNode.querySelectorAll('[data-hierarchy-type="file"][aria-hidden="false"]').length === 0
```

app/views/embed/template/_file.html.erb
jcoyne: Can this go back in the tree.js now that this is not being loaded via jQuery.ajax?

app/assets/javascripts/tree.js
justinlittman: I was just following the W3C example code. It seemed odd to me, but figured they had a reason.

app/assets/javascripts/tree.js
justinlittman: Done.

app/assets/javascripts/tree.js
```diff
@@ -341,6 +352,12 @@ Treeitem.prototype.init = function () {
     this.domNode.addEventListener('mouseover', this.handleMouseOver.bind(this));
     this.domNode.addEventListener('mouseout', this.handleMouseOut.bind(this));
   }
+
+  if(this.isSearchable) {
+    const filepath = this.domNode.querySelector('[data-filepath]').dataset.filepath.toLowerCase();
+    const description = this.domNode.querySelector('.sul-embed-description').textContent.toLowerCase();
```
jcoyne, line 358: Can we put a data selector on here?

spec/features/file_search_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+RSpec.describe 'file viewer search bar', js: true do
+  include PurlFixtures
+
+  context 'when text is entered with a file list' do
+    it 'limits shown files' do
+      stub_purl_response_with_fixture(file_purl)
+      visit_iframe_response('abc123', min_files_to_search: 1)
+      expect(page).to have_css('.sul-embed-count', count: 1)
+      expect(page).to have_css '.sul-embed-item-count', text: '1 item'
+      fill_in 'sul-embed-search-input', with: 'test'
```
jcoyne, line 14:
```suggestion
      fill_in 'Search this list', with: 'test'
```
Using labels helps test accessability.

spec/features/file_search_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      expect(page).not_to have_css('.sul-embed-count')
+      expect(page).to have_css '.sul-embed-item-count', text: '0 items'
+      fill_in 'sul-embed-search-input', with: 'File1'
```
jcoyne, line 17:
```suggestion
      fill_in 'Search this list', with: 'File1'
```

spec/features/file_search_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    it 'hides empty directories' do
+      stub_purl_response_with_fixture(hierarchical_file_purl)
+      visit_iframe_response('abc123', min_files_to_search: 1)
+      expect(page).to have_css '.sul-embed-item-count', text: '2 items'
+      expect(page).to have_content('dir1')
+      expect(page).to have_content('dir2')
+      expect(page).to have_content('Title_of_2_PDF')
+      expect(page).to have_content('Title_of_the_PDF')
+
+      fill_in 'sul-embed-search-input', with: 'Title_of_the_PDF'
```
jcoyne, line 32:
```suggestion
      fill_in 'Search this list', with: 'Title_of_the_PDF'
```

jcoyne (changes requested):
Looks to be pretty close now.

app/assets/javascripts/tree.js
justinlittman: Done.

jcoyne (approved):

PR #1436, opened 2023-01-23
Give jsx components a proper suffix

jcoyne:
This will help webpack to find the appropriate compiler


corylown:
Blocked until the following are merged:
https://github.com/sul-dlss/sul-embed/pull/1431
https://github.com/sul-dlss/sul-embed/pull/1420
https://github.com/sul-dlss/sul-embed/pull/1413

PR #1437, opened 2023-01-24
Remove unnecessary partial

jcoyne:
This was only ever used one time, so it just makes the code harder to follow


corylown:
Blocked until the following are merged:
https://github.com/sul-dlss/sul-embed/pull/1431
https://github.com/sul-dlss/sul-embed/pull/1420
https://github.com/sul-dlss/sul-embed/pull/1413

cbeer:
I'm not clear what was blocking this, but those are merged now 🤷‍♂️

PR #1441, opened 2023-01-24
Upgrade list.js to 1.5.0

jcoyne:


justinlittman:
Note that PR for updates to file viewer removes list.js.

PR #1447, opened 2023-01-26
Remove unused template

jcoyne:


cbeer:
Looks like we can remove `[app/views/embed/download/_image_x.html.erb](https://github.com/sul-dlss/sul-embed/blob/25c84b3c9396ff30693bba03264edde8763ea3f7/app/views/embed/download/_image_x.html.erb#L4)` too?

jcoyne:
@cbeer Thanks. Added that one.

PR #1462, opened 2023-02-01
Validate timemap url before acting on it

### LINKED ISSUE #1460: was-seed viewer throws 500 when archived site URL isn't correct
### Expected behavior

When the PURL XML contains an archived website URL that isn't properly formatted then sul-embed should probably display something helpful about the problem?

### Actual behavior

When the PURL XML contains an archived website URL that lacks a `*` it throws a 500 error:

Screenshot 2023-02-01 at 3 05 59 PM

Load up some PURL XML for a was-seed object containing a URL like this into the sandbox viewer:

```xml
<location>
  <url displayLabel="Archived website">https://swap.stanford.edu/was/20220831210052/https://wexarts.org/talks-more/tongues-untied</url>
</location>
```

### Browser / Environment

Firefox/MacOS

See https://github.com/sul-dlss/was_robot_suite/issues/279 for related work to prevent was-seed objects from getting published with incorrect URLs.

jcoyne:
Fixes #1460

### DIFF
--- a/app/viewers/embed/viewer/was_seed.rb
+++ b/app/viewers/embed/viewer/was_seed.rb
@@ -30,7 +30,13 @@ module Embed
       end
 
       def archived_timemap_url
-        archived_site_url.sub('/*/', '/timemap/')
+        if archived_site_url.include?('/*/') # Not a valid timemap url
+          archived_site_url.sub('/*/', '/timemap/')
+        else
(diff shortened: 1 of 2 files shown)


edsu (approved):
It looks like we will now get HB errors with this, and the embed view will be empty (instead of the big error message)? Seems like a big improvement to me!

PR #1464, opened 2023-02-02
Install the react-refresh-webpack-plugin

### LINKED ISSUE #1463: $RefreshReg$ is not defined
If you clear your assets (`bin/rails assets:clobber`) and then start `bin/webpacker-dev-server` and `bin/dev` and then go to 

http://localhost:3000/iframe?url=https://purl.stanford.edu/vh115kc5665

You get this error:
```
Uncaught ReferenceError: $RefreshReg$ is not defined
    jsx shareMenuPlugin.jsx:53
    Webpack 11
shareMenuPlugin.jsx:53
```

This can be worked around if you don't use hot reloading (e.g. precompile javascript via `bin/webpacker` or `bin/rails assets:precompile`)

It's possibly this issue? https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/397  But I still see this persisting even if I set NODE_ENV=development before starting bin/webpacker-dev-server

jcoyne:
Fixes #1463

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -1,7 +1,9 @@
 const { generateWebpackConfig, merge } = require('shakapacker')
 const webpack = require('webpack')
+const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
 
-// See the shakacode/shakapacker README and docs directory for advice on customizing your webpackConfig.
+const isDevelopment = process.env.NODE_ENV !== 'production';
+const baseWebpackConfig = generateWebpackConfig()
 const options = {
   resolve: {
       fallback: {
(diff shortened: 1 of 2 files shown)


config/webpack/webpack.config.js
```diff
@@ -1,6 +1,8 @@
 const { webpackConfig: baseWebpackConfig, merge } = require('shakapacker')
+const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
```
marlo-longley, line 2: @jcoyne maybe I am missing something but -- do we need this in package.json?

jcoyne: It's already here: https://github.com/sul-dlss/sul-embed/blob/master/package.json#L67

marlo-longley (approved):

lwrubel (approved):

lwrubel:
Works for me now locally. Approved pending tests passing.

PR #1469, opened 2023-02-08
Use VTT transcript

edsu:
If an object's video or audio file has an associated VTT file, add a `<track>` element which should allow the player to display the transcript during playing of the video or audio.

You can see [some SDR objects](https://argo.stanford.edu/catalog?f%5Bexploded_tag_ssim%5D%5B%5D=SMPL+%3A+avtc) already have VTT files added. Their PURL data looks something like:

```xml
<resource id="cocina-fileSet-bd111gd4290-bd111gd4290_1" sequence="1" type="video">
  <label>Video file</label>
    <file id="bd111gd4290_sl.mp4" mimetype="video/mp4" size="1077075674" publish="yes" shelve="yes" preserve="yes">
  </file>
  <file id="bd111gd4290_thumb.jp2" mimetype="image/jp2" size="296021" publish="yes" shelve="yes" preserve="yes">
    <imageData height="480" width="640"/>
  </file>
  <file id="bd111gd4290_cap.vtt" mimetype="text/plain" size="145678" publish="yes" shelve="yes" preserve="yes">
  </file>
</resource>
```

Ideally the VTT files would have a `text/vtt` mimetype? Also it might be possible for there to be multiple language transcriptions for a media file, which is not currently handled in this PR. Handling multiple languages should be doable if there was a convention or mechanism for determining the language.

Also, I think testing/research with `<audio>` should be done. I did a bit of quick looking around and [it seems](https://iandevlin.com/blog/2015/12/html5/webvtt-and-audio/) that maybe captions don't get displayed unless you say it's video instead? That might be stale information though.

This was quick exploratory in response to questions from @pleonard212 and @dinahhandel.

### DIFF
--- a/app/models/embed/purl/resource_file.rb
+++ b/app/models/embed/purl/resource_file.rb
@@ -48,7 +48,7 @@ module Embed
       end
 
       def vtt?
-        mimetype == 'text/vtt' && title.end_with?('.vtt')
+        mimetype == 'text/vtt'
       end
 
       def pdf?
--- a/app/models/embed/purl/resource_file.rb
+++ b/app/models/embed/purl/resource_file.rb
@@ -43,6 +43,14 @@ module Embed
         Settings.resource_types_that_contain_thumbnails.include?(resource.type)
       end
 
+      def vtt
+        resource.files.find(&:vtt?)
+      end
+
+      def vtt?
+        mimetype == 'text/plain' && title.end_with?('.vtt')
+      end
+
       def pdf?
         mimetype == 'application/pdf'
       end
--- a/app/models/embed/purl/resource_file.rb
+++ b/app/models/embed/purl/resource_file.rb
@@ -48,7 +48,7 @@ module Embed
       end
 
       def vtt?
-        mimetype == 'text/plain' && title.end_with?('.vtt')
+        mimetype == 'text/vtt' && title.end_with?('.vtt')
       end
 
       def pdf?
--- a/spec/fixtures/purl_fixtures.rb
+++ b/spec/fixtures/purl_fixtures.rb
@@ -972,7 +972,7 @@ module PurlFixtures
         <contentMetadata type="media">
           <resource sequence="1" id="abc123_1" type="video">
             <file id="abc_123.mp4" mimetype="video/mp4" size="152000000"></file>
-            <file id="abc_123_cap.vtt" mimetype="text/vtt" size="176218"></file>
+            <file id="abc_123_cap.webvtt" mimetype="text/vtt" size="176218"></file>
           </resource>
         </contentMetadata>
         <rightsMetadata>
--- a/spec/fixtures/purl_fixtures.rb
+++ b/spec/fixtures/purl_fixtures.rb
@@ -972,7 +972,7 @@ module PurlFixtures
         <contentMetadata type="media">
           <resource sequence="1" id="abc123_1" type="video">
             <file id="abc_123.mp4" mimetype="video/mp4" size="152000000"></file>
-            <file id="abc_123_cap.vtt" mimetype="text/plain" size="176218"></file>
+            <file id="abc_123_cap.vtt" mimetype="text/vtt" size="176218"></file>
           </resource>
         </contentMetadata>
         <rightsMetadata>
--- a/spec/fixtures/purl_fixtures.rb
+++ b/spec/fixtures/purl_fixtures.rb
@@ -963,6 +963,32 @@ module PurlFixtures
     XML
   end
 
+  def video_purl_with_vtt
+    <<-XML
+      <publicObject>
+        <identityMetadata>
+          <objectLabel>Title of the single video</objectLabel>
+        </identityMetadata>
+        <contentMetadata type="media">
+          <resource sequence="1" id="abc123_1" type="video">
+            <file id="abc_123.mp4" mimetype="video/mp4" size="152000000"></file>
+            <file id="abc_123_cap.vtt" mimetype="text/plain" size="176218"></file>
+          </resource>
+        </contentMetadata>
+        <rightsMetadata>
+          <access type="read">
+            <machine>
+              <location>spec</location>
+            </machine>
+          </access>
+        </rightsMetadata>
+        <oai_dc:dc xmlns:oai_dc="http://www.openarchives.org/OAI/2.0/oai_dc/" xmlns:dc="http://purl.org/dc/elements/1.1/">
+          <dc:title>title of video with vtt</dc:title>
+        </oai_dc>
+      </publicObject>
+    XML
+  end
+
   def video_purl
     <<-XML
       <publicObject>
--- a/app/models/embed/purl/resource.rb
+++ b/app/models/embed/purl/resource.rb
@@ -48,6 +48,11 @@ module Embed
       def thumbnail
         files.find(&:thumbnail?)
       end
+
+      # @return [ResourceFile]
+      def vtt
+        files.find(&:vtt?)
+      end
     end
   end
 end
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -4,6 +4,7 @@ module Embed
   # Utility class to handle generating HTML <video> and <audio> tags
   # Currently, MPEG-DASH is used at the <video> element level (in a data attribute to be picked up by javascript)
   # and HLS is used as a <source> within the <video> or <audio> tag.
+  # rubocop:disable Metrics/ClassLength
   class MediaTag
     SUPPORTED_MEDIA_TYPES = %i[audio video].freeze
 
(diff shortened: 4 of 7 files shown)


cbeer:
See also #1143.

edsu:
@andrewjbtw created a test object in Stage for testing. It didn't seem easy to get video streaming to work in my dev environment, maybe because of CORS?

andrewjbtw:
Two notes about the test object:

1. I didn't assign the vtt file any role. In OCR, all the ALTO files are assigned "role=transcription" in order to be picked up for search. In preliminary discussion of caption files, we discussed following that precedent in some way but did not settle on a term. I could set it to "transcription." Cocina validates roles so we would need to add a role to the vocabulary if we wanted something like "caption".

2. If you look at the filenames, the video file and the caption file differ by both name and extension. Some players like VLC will detect captions if the names are the same and only the extensions differ. I can make a test like that but ultimately it would be more robust if the video file and caption file could differ in name. Plus that would be essential if we offered subtitles in multiple languages.

edsu:
I pushed this to stage where there is a test video object, and it seems to work!

https://sul-purl-stage.stanford.edu/dw202dq8174

spec/fixtures/purl_fixtures.rb
```diff
@@ -963,6 +963,32 @@ def single_video_purl
...
+  def video_purl_with_vtt
+    <<-XML
+      <publicObject>
+        <identityMetadata>
+          <objectLabel>Title of the single video</objectLabel>
+        </identityMetadata>
+        <contentMetadata type="media">
+          <resource sequence="1" id="abc123_1" type="video">
+            <file id="abc_123.mp4" mimetype="video/mp4" size="152000000"></file>
+            <file id="abc_123_cap.vtt" mimetype="text/plain" size="176218"></file>
```
jcoyne, line 975: Isn't `text/vtt` the registered mime-type for vtt files?

app/models/embed/purl/resource_file.rb
```diff
@@ -43,6 +43,14 @@ def thumbnail?
         Settings.resource_types_that_contain_thumbnails.include?(resource.type)
       end
 
+      def vtt
+        resource.files.find(&:vtt?)
+      end
+
+      def vtt?
+        mimetype == 'text/plain' && title.end_with?('.vtt')
```
jcoyne, line 51: Shouldn't this be `text/vtt`? https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API#webvtt_files

jcoyne (changes requested):

app/models/embed/purl/resource_file.rb
edsu: I think that's right @jcoyne. I think it might be coded in the example XML that way? Perhaps it should allow for both?

jcoyne: I think the example XML is just bad data.

edsu: It looks there are 45 text/vtt files at the moment: https://argo.stanford.edu/catalog?all=true&f[content_file_mimetypes_ssim][]=text%2Fvtt 

I guess I'd need to run a report to see if there are many/any `.vtt` files that need the correct media type.

spec/fixtures/purl_fixtures.rb
edsu: Yes it is - nice catch.

app/models/embed/purl/resource_file.rb
edsu: Now I'm wondering if it should ignore the extension altogether and only key of the media type?

jcoyne: Yes, I don't think the extension is something we care about.

andrewjbtw: It's not exactly bad data. vtt is a text file. What we need is more precise identification of text files. Infrastructure will need to build that. We have similar identification problems with lots of text file types. The more specifc the consuming applications need mimetypes, the more precisely the accessioning apps need to identify them.

We don't want a situation where a person has to manually correct mimetypes after each accession. If you ignore the extension here, I think the most likely the path to identifying VTT files as text/vtt will be to use the extension during accessioning. Which may be the best way to do it, but there's both trade-offs and cross-team implications to whatever strategy we choose.

edsu: That's an important point @andrewjbtw -- that would suck to have to in and manually fix anything with a transcript.  Does techmd do the mediatype detection for us at the moment? Do we need a ticket in there, and perhaps a data remediation ticket as well?

jcoyne: Improvements to mime-type detection should go in https://github.com/sul-dlss/assembly-objectfile/blob/229520e3c590dc3e8c83157c5d62c00cd3e53eb7/lib/assembly/object_file.rb#L103. 

It does have a problem with this type as it prefers the value returned from `file` before the value from `extension`

```ruby
file = Assembly::ObjectFile.new('foo.vtt')
file.send :exif_mimetype
=> nil
file.send :file_mimetype
=> "text/plain"
file.send :extension_mimetype
=> "text/vtt"
```

andrewjbtw: There are a couple of different places where mime-types are determined. I think `sdr-api` does it as well. Maybe counterintuitively, the mime-type in the structural metadata is generated and stored separately from the mime-type in the technical metadata, which has been the case going back to the Fedora era. The access systems don't use the technical metadata.

jcoyne (approved):

andrewjbtw:
@jcoyne Does merging this PR mean this will go out to prod and items with VTT will start showing captions next week? I ask because I'm not sure we're ready for that in terms of the styling/UI.

edsu:
@andrewjbtw my apologies, this is my fault for taking the PR out of draft, which signaled to @jcoyne that it was in fact ready.

One thing that came up when we were discussing this in the #dlss-av-captions meeting on April 21st was the need for users to be able to turn off the transcript from playing. It does appear that VideoJS should support adding this if the existing control isn't obvious enough:

https://videojs.com/guides/text-tracks/#working-with-text-tracks

There was also some concern about the length of the each line of text. I believe this is a function of the lines in the VTT file itself, and isn't behavior we can easily change programmatically. 

Fortunately there are very few ([44](https://argo.stanford.edu/catalog?all=true&f%5Bcontent_file_mimetypes_ssim%5D%5B%5D=text%2Fvtt)) object files coded with a media type of `text/vtt` at the moment, so it shouldn't be too disruptive (as long as it works correctly?).

andrewjbtw:
Looking at some of those items, I think this will cause vtt captions to appear over burned-in captions. For example, the videos with VTT in https://argo.stanford.edu/view/druid:bb761mb4522 are the two videos with burned in English and German captions. I think with VTT captioning turned on, this will overlay the VTT captions onto the video.

It's a small number for sure, but most of them (36) appear to be items where care was taken to generate burned-in captions specifically for a high-profile project. The other 9 appear to be zoom captions.

There's a larger number where the VTT was identified as "text/plain" but those won't show captions until remediated.

edsu:
Would it be helpful for me to add a feature flag to turn off VTT transcripts in settings for now?

jcoyne:
@edsu If andrew doesn't want this out yet, then yes, that would be a good idea.

andrewjbtw:
Let me see if I can get some quick feedback. I do appreciate being able to have captions, just wasn't expecting this this week.

The part of the UI that doesn't look great to me is the settings menu, not really the captions themselves.

andrewjbtw:
@edsu After making another sample for stage (of one of the VT objects) and getting feedback from PSM, we should not turn this on yet. I'm happy to file a separate issue for keeping it off.

jcoyne:
Agreed. The styles are messed up:
Screenshot 2023-05-17 at 2 07 33 PM

app/models/embed/purl/resource_file.rb
mjgiarlo: FWIW, I just deposited a VTT file using the SDR API in stage, and it appears both sdr-api and the techmd service applied the correct MIME type: https://argo-stage.stanford.edu/view/druid:pc587kh4617

assembly-objectfile may be the one spot needing attention here as @jcoyne shared above.

mjgiarlo: I was wrong about the above: @edsu already patched `assembly-objectfile` such that is correctly picks the VTT mime type, and that's in the latest version of the gem.

dinahhandel:
### Expected behavior

A user should be able to choose if they want captions to display by clicking a button or otherwise selecting the functionality.
The CC icon as well as caption settings options must display in a way that does not interfere with playback of the media (as dropdown menus? As a pop-up over the media?).

### Other considerations

What if there is no caption file? Should an option to play CC still display? Is it possible for a viewer to only show an option to play CC if there is a web.vtt file present? Or is CC hard-coded into the viewer and we need to notify a user that if there is no web.vtt file, captions are not yet available for playback? If this is the case, can we build in better functionality for requesting that the media be captioned?

PR #1499, opened 2023-05-19
Add feature flag for transcripts

### LINKED ISSUE #1497: Add feature flag to turn off caption display for now
PR: https://github.com/sul-dlss/sul-embed/pull/1469 to add display of VTT captions has been merged but caption display is not ready for production release.

This ticket is a follow-on to discussion in that PR to use a feature flag to prevent caption display from being turned on until more time is dedicated to captions.

edsu:
The `Settings.enabled_features.transcripts` flag has been added to control
whether the viewer will display a `<track>` element containing the transcript for a
video.

The setting can be forcibly turned on by adding `trascripts=true` to the URL
query parameter for the embed viewer. Note, this will require you to look at
the view directly instead of in an embedded frame in PURL or SearchWorks.

Until these shared_configs changes are merged the feature will be on by default:

- https://github.com/sul-dlss/shared_configs/pull/1997
- https://github.com/sul-dlss/shared_configs/pull/1998

Resolves #1497

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -64,3 +64,5 @@ collections_to_show_attribution:
 # in shared_configs for embed-prod, we set this value to 0  
 analytics_debug: true
 min_files_to_search_default: 10
+enabled_features:
+  transcripts: false
--- a/lib/embed/media_tag.rb
+++ b/lib/embed/media_tag.rb
@@ -12,8 +12,9 @@ module Embed
 
     attr_accessor :file_index
 
-    def initialize(viewer)
+    def initialize(viewer, include_transcripts: false)
       @viewer = viewer
+      @include_transcripts = include_transcripts
       @purl_document = viewer.purl_object
       @file_index = 0
     end
@@ -94,7 +95,7 @@ module Embed
     end
 
     def transcript(file)
-      return unless file.vtt
+      return unless @include_transcripts && file.vtt
 
       <<~HTML
         <track default src="#{viewer.stacks_url}/#{file.vtt.title}" />
--- a/app/views/embed/body/_media.html.erb
+++ b/app/views/embed/body/_media.html.erb
@@ -1,6 +1,6 @@
 <div class='sul-embed-body sul-embed-media'
      data-sul-embed-theme="<%= stylesheet_path('media.css') %>"
      data-sul-icons="<%= stylesheet_path('sul_icons.css') %>">
-  <%= Embed::MediaTag.new(viewer).to_html.html_safe %>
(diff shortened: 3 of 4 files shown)


lib/embed/media_tag.rb
```diff
@@ -94,7 +94,7 @@ def enabled_streaming_sources(file)
     end
 
     def transcript(file)
-      return unless file.vtt
+      return if !file.vtt || !Settings&.enabled_features&.transcripts
```
jcoyne, line 97: Why do you have the safe navigation operator here? There's no way for these to be nil, is there?

What's more, is that I'd prefer this value to be injected into the initializer of MediaTag.  Thus ```initialize(viewer, display_transcript: false)```

This allows you to remove `set_transcripts_flag_per_param` from the controller and pass `MediaTag.new(viewer, params[:transcript] == 'true')` in https://github.com/sul-dlss/sul-embed/blob/e1194564101d7e6ec209d920bd1c457d9e6a8fdb/app/views/embed/body/_media.html.erb#L4

jcoyne (changes requested):

config/settings.yml
```diff
@@ -64,3 +64,5 @@ collections_to_show_attribution:
 # in shared_configs for embed-prod, we set this value to 0  
 analytics_debug: true
 min_files_to_search_default: 10
+enabled_features:
+  transcripts: true
```
jcoyne, line 68: Please set the default to false. Then turn it on in any environment you want to see it in.

app/controllers/embed_controller.rb
```diff
@@ -54,6 +55,10 @@ def embed_request
     render body: e.to_s, status: :unsupported_media_type
   end
 
+  def set_transcripts_flag_per_param
+    Settings.enabled_features.transcripts = true if params[:transcripts] == 'true'
```
jcoyne, line 59: This settings object is a global, so you can't change it on a per-request basis.

jcoyne: @edsu I don't think that's a viable pattern in Argo either.  Any other request served by the same process will be affected by that setting.

edsu:
Thanks for the feedback @jcoyne -- I've made some adjustments. Let me know if anything else jumps out.

lib/embed/media_tag.rb
edsu: I moved it to the constructor, with a named parameter. Let me know if that doesn't make sense, and thanks for this suggestion.

config/settings.yml
edsu: Ok, it is set to false by default now. Maybe we don't need the shared_config changes now?

jcoyne (approved):

PR #1534, opened 2023-10-16
Add blueprintjs to avoid error in dev

edsu:
Without adding this you get some warnings in dev that make development difficult (clicking through to ignore the warning). See this thread for context: https://stanfordlib.slack.com/archives/C060ZTZRV/p1680278501914259?thread_ts=1680275819.309989&cid=C060ZTZRV

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "defaults"
   ],
   "devDependencies": {
+    "@blueprintjs/core": "^5.5.0",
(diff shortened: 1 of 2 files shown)


jcoyne:
blueprintjs missing is a warning, not an error. We are not using blueprintjs and so we don't want to add the extra weight to the build.

edsu:
I understand not wanting to burden the build with additional dependencies. What are the steps to get sul-embed to use the ignore plugin?

PR #1554, opened 2023-10-18
Extract media metadata component

jcoyne:


app/components/embed/media/metadata_component.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Embed
+  module Media
+    class MetadataComponent < ViewComponent::Base
```
thatbudakguy, line 5: is it intentional that we flipped the order of modules? the previous partial was `/embed/metadata/media` which I feel like would translate to `Embed::Metadata::MediaComponent` not `Embed::Media::MetadataComponent`

jcoyne: Yes, it was intentional.  I wanted to make a module just for "media" things.   I think that is the most important axis for dividing these responsibilities. Otherwise we'd have modules for things that may only be used by one viewer.

thatbudakguy: OK, just wanted to check

thatbudakguy (approved):

PR #1556, opened 2023-10-18
Switch from sprockets-rails to propshaft

jcoyne:


jcoyne:
@edsu thanks. It's now rebased.

PR #1559, opened 2023-10-18
Move file_url from the viewer to the file model

jcoyne:
This makes it possible to use the method in more places and more consistently

### DIFF
--- a/app/models/embed/purl/resource.rb
+++ b/app/models/embed/purl/resource.rb
@@ -3,10 +3,16 @@
 module Embed
   class Purl
     class Resource
-      def initialize(resource, rights)
+      # @param [String] druid identifier without a namespace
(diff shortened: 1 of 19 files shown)


app/models/embed/purl/resource.rb
```diff
@@ -3,11 +3,17 @@
...
+      # @param [Nokogiri::XML::Element] resource
+      # @param [Dor::RightsAuth] rights
+      def initialize(druid, resource, rights)
+        @druid = druid
         @resource = resource
         @rights = rights
       end
 
+      attr_reader :druid
+
```
thatbudakguy, line 16: can we not pick the druid out of the resource XML? I'd assume it'd be in there somewhere

jcoyne: It is not.

PR #1565, opened 2023-10-20
Switch to use Rails 7.0 default config

jcoyne:


config/application.rb
```diff
@@ -22,7 +22,7 @@
 module SulEmbed
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 6.1
+    config.load_defaults 7.0
```
edsu, line 25: Does it matter if this isn't 7.1 (version of rails that's installed)?

edsu (approved):

PR #1569, opened 2023-10-22
Simplify call to route helper

jcoyne:
This makes it possible to use ViewComponent previews, which makes developing the UI quite a bit easier.


ndushay (approved):
LGTM but what do I know?

Is there a place where vc_test_controller is defined that can now go away?

PR #1572, opened 2023-10-23
Use data attributes to bind javascript rather than classes

jcoyne:
the .sul-embed-media class was used to bind behaviors, for the bootstrap media class and for custom media styles.  This change helps separate all three of those concerns


thatbudakguy (approved):
this all makes sense, but where's the "legacy" prefix come from? what does that mean in this context?

jcoyne:
legacy as in "not the new component we're developing" I'm thinking this one will be feature flagged until we can ship the new version

PR #1574, opened 2023-10-23
Convert MediaTag to a view component

jcoyne:


app/components/embed/media_component.rb
```diff
@@ -7,5 +7,22 @@ def initialize(viewer:)
...
+      Settings.enabled_features.transcripts || params[:transcripts] == 'true'
+    end
+
+    def primary_files_count
+      purl_object.contents.count do |resource|
+        resource.primary_file.present?
+      end
+    end
+
+    def many_primary_files
```
mjgiarlo, line 24:
```suggestion
    def many_primary_files?
```

jcoyne: But then we can't pass it as an implicit argument `many_primary_files:`

mjgiarlo (approved):
Other than a trivial suggestion, approving.

mjgiarlo (approved):
Other than a trivial suggestion, approving.

PR #1580, opened 2023-10-24
Add honeybadger to the frontend

jcoyne:
This allows developers to be informed of errors without relying on users to file a bug report.

### DIFF
--- a/app/views/embed/iframe.html.erb
+++ b/app/views/embed/iframe.html.erb
@@ -5,6 +5,15 @@
     <script>
       <%= 'if (window.jQuery) { jQuery.fx.off = true; }' if Rails.env.test? %>
     </script>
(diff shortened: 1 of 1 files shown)


app/views/embed/iframe.html.erb
```diff
@@ -5,6 +5,15 @@
     <script>
       <%= 'if (window.jQuery) { jQuery.fx.off = true; }' if Rails.env.test? %>
     </script>
+    <script src="//js.honeybadger.io/v6.5/honeybadger.min.js" type="text/javascript"></script>
```
mjgiarlo, line 8: What does the `//` mean here? I'm unfamiliar with this syntax.

jcoyne: That means, use whatever protocol delivered this page.

mjgiarlo (approved):

PR #1586, opened 2023-10-24
Boolean method should end in a question mark

jcoyne:


mjgiarlo:
It's better as is. Thanks for working this up.

PR #1595, opened 2023-10-24
Fix up companion window media component to use the new MediaTagComponent

jcoyne:
Now that MediaComponent uses this, the MediaWithCompanionWindowsComponent can too

### DIFF
--- a/app/assets/stylesheets/media.scss
+++ b/app/assets/stylesheets/media.scss
@@ -135,7 +135,6 @@
   .media-component-body {
     display: flex;
     flex-grow: 1;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/media.scss
```diff
@@ -135,7 +135,6 @@
   .media-component-body {
     display: flex;
     flex-grow: 1;
-    // height: 100%;
```
mjgiarlo, line 138: Cruft?

mjgiarlo (approved):

PR #1614, opened 2023-10-26
Replaces max-age caching with conditional get caching.

### LINKED ISSUE #1578: Investigate caching settings
We've had multiple reports of users experiencing undesirable caching and using alternative methods (incognito, etc.) in order to view changes to purl pages.

Some possibly related tickets:
#700 
#1028
#1496

justinlittman:
closes #1578

image


app/controllers/embed_controller.rb
```diff
@@ -71,14 +57,31 @@ def linted_params
 
   private
 
-  def set_cache
-    return unless Rails.env.production?
+  def embed_request
```
justinlittman, line 60: This is the same; just moved below private.

app/controllers/embed_controller.rb
```diff
@@ -3,12 +3,14 @@
 class EmbedController < ApplicationController
   append_view_path Rails.root.join('app/views/embed')
   before_action :embed_request
-  before_action :set_cache
```
justinlittman, line 6: No longer set max-age.

jcoyne (approved):

PR #1621, opened 2023-10-28
handling image size for poster URL

### LINKED ISSUE #1604: Poster should show for video with thumbnail
This item has a thumbnail, so it should display as the poster: https://sul-purl-stage.stanford.edu/cm856pm4228

hudajkhan:
Closes #1604

### DIFF
--- a/app/components/embed/media_tag_component.rb
+++ b/app/components/embed/media_tag_component.rb
@@ -5,6 +5,8 @@ module Embed
     include StacksImage
     with_collection_parameter :resource
     SUPPORTED_MEDIA_TYPES = %i[audio video].freeze
+    MAX_WIDTH = 800
+    MAX_HEIGHT = 600
 
     def initialize(resource:, resource_iteration:, druid:, include_transcripts:, many_primary_files:)
       @resource = resource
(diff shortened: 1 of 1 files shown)


cbeer:
Should the size prefix `!` handle this check?

> The extracted region is scaled so that the width and height of the returned image are not greater than w and h, while maintaining the aspect ratio. The returned image must be as large as possible but not larger than the extracted region, w or h, or server-imposed limits.

I'm not sure why.. something.. is trying to up-scale the image

hudajkhan:
I wondered that @cbeer , but in practice, if the !w,h request is asking for w + h that are larger than the original image size, it is throwing an error.  I have these notes on the github issue (the original image width was 640): 

Getting to the image server directly, requesting "iiif/2/cm%2F856%2Fpm%2F4228%2Fgt507vy5436_thumb.jp2/full/!641,600/0/default.jpg" using gives the following error

"403 Forbidden

Requests for scales in excess of 100% are not allowed.

edu.illinois.library.cantaloupe.resource.ScaleRestrictedException: Requests for scales in excess of 100% are not allowed."

cbeer:
But we have control over stacks, so couldn't we fix it there?

hudajkhan:
Do you mean image server configuration? I wanted to ask you that, because I'm not sure what's expected/what we can change. I'm seeing the same kind of behavior in production as well for some examples.

cbeer:
I mean, we have a ton of logic in stacks to convert from our stacks IIIF URLs to the cantaloupe request:

https://github.com/sul-dlss/stacks/blob/master/app/models/projection.rb

Why wouldn't we add the same sorts of checks you're proposing here to that code so all `!`-style image requests work correctly?

PR #1625, opened 2023-10-30
Add a way to click on a thumbnail to bring up the viewer

jcoyne:


app/javascript/src/controllers/media_tag_controller.js
```diff
@@ -2,11 +2,13 @@ import { Controller } from "@hotwired/stimulus"
 import videojs from 'video.js';
 import validator from '../modules/validator.js'
 import mediaTagTokenWriter from '../modules/media_tag_token_writer.js'
+import buildThumbnail from '../modules/media_thumbnail_builder.js'
 
 export default class extends Controller {
   static targets = [ "mediaTag"]
```
mjgiarlo, line 8: Maybe I'm misreading this, but I'm curious why the newly added `list` target (for the media tag controller) is not registered here? Wouldn't that help on line 39 by being able to call `this.listTarget` instead of the `querySelector` call? Or is there a reason not to use Stimulus targets here?

mjgiarlo (approved):

app/javascript/src/controllers/media_tag_controller.js
jcoyne: Good catch! I've made that change.

mjgiarlo (approved):

PR #1626, opened 2023-10-30
Prefilter resources to ensure they have primary files

### LINKED ISSUE #1627: Clicking thumbnail shows the wrong media
Go to https://purl.stanford.edu/gt507vy5436, then open the drawer for the media. Click on the third thing and the second item will display in the viewer.

jcoyne:
Otherwise the index in MediaTagComponent might skip the sequence.  This is a problem because the `thumb_slider.js` expects sequential indexes.

Fixes #1627

### DIFF
--- a/app/components/embed/legacy_media_component.rb
+++ b/app/components/embed/legacy_media_component.rb
@@ -16,8 +16,12 @@ module Embed
     end
 
     def primary_files_count
-      purl_object.contents.count do |resource|
-        resource.primary_file.present?
+      resources_with_primary_file.count
+    end
+
+    def resources_with_primary_file
+      @resources_with_primary_file ||= purl_object.contents.select do |purl_resource|
+        purl_resource.primary_file.present?
       end
     end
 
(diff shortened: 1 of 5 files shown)


app/components/embed/legacy_media_component.rb
```diff
@@ -16,8 +16,12 @@ def include_transcripts
...
-      purl_object.contents.count do |resource|
-        resource.primary_file.present?
+      resources_with_primary_file.count
+    end
+
+    def resources_with_primary_file
+      @resources_with_primary_file ||= purl_object.contents.select do |purl_resource|
+        purl_resource.primary_file.present?
       end
     end
```
mjgiarlo, line 26: Naïvely, this code looks similar (identical?) to lines 22-26 in `app/components/embed/media_with_companion_windows_component.rb`. Do you think it's worth considering extracting them into a model or a service class, or do you expect these methods to diverge? (Alternatively, maybe there's nothing to do because the legacy component is not long for this world?)

mjgiarlo (approved):
Approve pending consideration of comment

app/components/embed/legacy_media_component.rb
jcoyne: Yeah, the legacy component is supposed to go away, so I don't want to build connections between the two.

Issue #1631, opened 2023-10-31
Add headers for various companion windows


jcoyne:

mjgiarlo:
Alan will drop tab headers roughly here-ish.

alundgard:
(Mirador [example](https://purl.stanford.edu/bb627py6670) for comparison)

**[[Material UI Icon](https://mui.com/material-ui/material-icons/) Name] Header Title**

Left Companion Window

[Info] About this item
[Copyright] Rights
[Share] Share

Right Companion Window

[FormatListBulleted] Media Content
[Download] Download

mjgiarlo:
We think this is done. Closing.

PR #1662, opened 2023-11-02
Add headers to companion window components

mjgiarlo:
Connects to #1631

This commit adds headers to both the left and right companion window components. It also extracts the rights component out of the metadata list component. Note that this commit only affects the HTML DOM and JavaScript; the CSS to style the headers similar to Mirador is not included. (Too time-consuming given my current level of CSS skill.)

Screenshot from 2023-11-02 13-02-33


app/components/embed/media_with_companion_windows_component.html.erb
```diff
@@ -53,9 +77,21 @@
     <div class="right-drawer collapse" data-media-target="rightDrawer">
       <div class="inner">
         <div data-media-target="contents" hidden>
+          <div class="MuiToolbar-root MuiToolbar-regular">
```
jcoyne, line 80: Can you remove these Mui* classes?  We aren't using them. I think just the h3 will be fine.

jcoyne (changes requested):

app/components/embed/media_with_companion_windows_component.html.erb
mjgiarlo: Good eye, that was indeed cruft.

spec/components/metadata_list_component_spec.rb
```diff
@@ -9,19 +9,11 @@
 
   let(:viewer) { instance_double(Embed::Viewer::Media, purl_object:) }
   let(:purl_object) do
-    instance_double(Embed::Purl,
-                    use_and_reproduction: 'The materials are open for research use',
-                    copyright: 'The Board of Trustees of the Leland Stanford Junior University',
-                    license: 'cc-0',
-                    purl_url: 'https://purl.stanford.edu/gt507vy5436')
+    instance_double(Embed::Purl, purl_url: 'https://purl.stanford.edu/gt507vy5436')
```
jcoyne, line 12: Will this change trigger network access in the tests?

mjgiarlo: The purl URL was there before. This change is removing code that was already here (extracting rights stuff to its own component spec).

jcoyne (approved):

PR #1667, opened 2023-11-03
Use default thumbnail for media if dataset thumbnail URL is blank

### LINKED ISSUE #1640: Broken image in right drawer when there is no thumbnail
Seen in https://embed-stage.stanford.edu/iframe?new_component=true&url=https://sul-purl-stage.stanford.edu/nr582tm3161&_v=1698775996
Screenshot 2023-10-31 at 2 23 53 PM

mjgiarlo:
Not just an empty string.

Fixes #1640

Screenshot from 2023-11-03 11-14-59

### DIFF
--- a/app/javascript/src/modules/media_thumbnail_builder.js
+++ b/app/javascript/src/modules/media_thumbnail_builder.js
@@ -9,10 +9,10 @@ export default function(dataset, index) {
     }
 
     let thumbnailIcon = '';
-    if (dataset.thumbnailUrl !== '') {
+    if (dataset.thumbnailUrl) {
       thumbnailIcon = `<img class="sul-embed-media-square-icon" src="${dataset.thumbnailUrl}" />`
     } else {
-      thumbnailIcon = `<i class="${dataset.defaultIcon}"></i>`
+      thumbnailIcon = `<i class="${dataset.defaultIcon} default-thumbnail-icon"></i>`
     }
 
     const isLocationRestricted = dataset.locationRestricted === "true"
--- a/app/assets/stylesheets/media.scss
+++ b/app/assets/stylesheets/media.scss
@@ -123,7 +123,7 @@
   .sul-embed-media-file {
     height: calc(100% - 54px);;
   }
-  
+
   .button-label {
     width: 100%;
     display: flex;
@@ -169,12 +169,12 @@
     background-color: var(--active-color);
     color: white;
     padding: .5rem;
-    
+
     .loginMessage {
       margin-left: 1rem;
     }
 
-    button { 
+    button {
       margin-left: 0.75rem;
       background-color: white;
       color: var(--active-color);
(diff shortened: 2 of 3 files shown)


app/javascript/src/modules/media_thumbnail_builder.js
```diff
@@ -9,7 +9,7 @@ export default function(dataset, index) {
     }
 
     let thumbnailIcon = '';
-    if (dataset.thumbnailUrl !== '') {
+    if (dataset.thumbnailUrl && dataset.thumbnailUrl !== '') {
```
mjgiarlo, line 12: The bug here was that the `img@src` attr below was being set to `undefined` because the check above only looks at whether `thumbnailUrl` is an empty string, not if it's defined.

If folks prefer that this bug should be fixed upstream of this code, holler at me.

jcoyne: Yes, I'm wondering why it wasn't defined. I think we should track that down too.

mjgiarlo: @jcoyne Will do

mjgiarlo: @jcoyne When the file has no thumbnail (when `file.thumbnail` is falsey), `file_thumb` is set to `nil` here: https://github.com/sul-dlss/sul-embed/blob/db08304bd05e518bfe11bd0495b6108b13e369ef/app/components/embed/media_tag_component.rb#L37

jcoyne:
This also doesn't feel like the design is right.  I was thinking this icon should stretch to fill the area that would have been occupied by the thumbnail if it were present. Can you add that?

app/javascript/src/modules/media_thumbnail_builder.js
jcoyne: okay, so `data: { thumbnail_url: nil }` ends up not generating any attribute, thus we need to do the check on `dataset.thumbnailUrl`

jcoyne: How about we switch it to `data: { thumbnail_url: @thumbnail.presence }` and then we can drop the empty string check in the javascript?

mjgiarlo: Sounds good, as I don't think it's ever finding the empty string now.

mjgiarlo:
@jcoyne 💬 
> This also doesn't feel like the design is right. I was thinking this icon should stretch to fill the area that would have been occupied by the thumbnail if it were present. Can you add that?

This is an icon from the sul_styles icon font, I think? I've tried mucking with `width` and `height` but they are invalid properties in this context. I can make the icon larger by choosing `font-size: xxx-large` but it doesn't fill the space. If you have CSS advice or want to pair on this, I'm all ears.

jcoyne:
@mjgiarlo have you tried something like `font-size: 52px`?

jcoyne:
@mjgiarlo I think a px value makes sense here because we're matching to an image thumbnail that has a particular px value.

mjgiarlo:
@jcoyne Just pushed up the change with an updated screenshot ☝🏻

jcoyne (approved):
Thanks @mjgiarlo . I appreciate all the changes. 🙇

PR #1674, opened 2023-11-06
Improve development experience

mjgiarlo:
Includes:
* Bundle JS as part of bin/dev processes
* Add remote debugger to dev web server and document how to attach debuggers
* Dump dockerized asset compilation (because this was only needed when using unsupported webpacker gem---now using supported shakapacker gem)


bin/bundle
```diff
@@ -1,3 +1,114 @@
 #!/usr/bin/env ruby
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
-load Gem.bin_path('bundler', 'bundle')
+# frozen_string_literal: true
```
mjgiarlo, line 2: Regenerated the bundle binstub because bundler told me to when running the `bin/rdbg` binstub.

jcoyne (approved):

PR #1678, opened 2023-11-06
Add aria-controls to the relevant nav components of the viewer

### LINKED ISSUE #1670: Add aria-controls to the buttons for the media viewer
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

aaron-collier:
Fixes #1670

There may be a few more buttons that need these values. I can put in follow up PRs if this is reviewed/merged before I identify any more controls.

### DIFF
--- a/app/components/embed/media_with_companion_windows_component.html.erb
+++ b/app/components/embed/media_with_companion_windows_component.html.erb
@@ -1,14 +1,14 @@
 <div class="sul-embed-container media-component" id='sul-embed-object' style='display:none;' data-controller="media media-tag">
   <header>
     <nav aria-label="Window navigation">
-      <button data-action="click->media#toggleLeft">
+      <button data-action="click->media#toggleLeft" aria-label="Display sidebar" aria-controls="left-drawer">
         <span class="button-label"><svg class="MuiSvgIcon-root" focusable="false" viewBox="0 0 24 24" aria-hidden="true"><path d="M3 18h18v-2H3v2zm0-5h18v-2H3v2zm0-7v2h18V6H3z"></path></svg></span>
       </button>
       <h2 class=""><%= viewer.purl_object.title %></h2>
-      <button data-action="click->media#toggleContents" aria-label="Content">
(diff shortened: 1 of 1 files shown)


app/components/embed/media_with_companion_windows_component.html.erb
```diff
@@ -34,35 +34,35 @@
     </div>
   </div>
 
-  <div class="media-component-body">
+  <div class="media-component-body" id="left-drawer">
```
jcoyne, line 37: This is more than the left drawer.

jcoyne (changes requested):

app/components/embed/media_with_companion_windows_component.html.erb
```diff
@@ -1,14 +1,14 @@
 <div class="sul-embed-container media-component" id='sul-embed-object' style='display:none;' data-controller="media media-tag">
   <header>
     <nav aria-label="Window navigation">
-      <button data-action="click->media#toggleLeft">
+      <button data-action="click->media#toggleLeft" aria-label="Display left menu" aria-controls="left-drawer">
```
jcoyne, line 4: Is calling out "left" important here for someone who may be visually impaired? Is there a better description of this drawer we could use @alundgard ?

app/components/embed/media_with_companion_windows_component.html.erb
aaron-collier: Sure, I don't think in this context it's being used informatively, it's just the link to the `aria-controls` value, but perhaps I'm misunderstanding.

jcoyne: I'm saying that this is the parent container of the left drawer and contains the media player and the right drawer too.

aaron-collier: Ah, gotcha.. Yeah, that's not right, I probably just put this in the wrong place, apologies.

jcoyne (approved):

PR #1691, opened 2023-11-07
Add appropriate tab, tablist, and tabpanel attributes to left drawer

### LINKED ISSUE #1677: Add roles for tab, tablist & tabpanel to the media viewer
The buttons on the left side function as tabs so we should add the appropriate roles there. 

See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role

The tab should also contain the `aria-controls` attribute and the `aria-selected` attribute

aaron-collier:
Fixes #1677

### DIFF
--- a/app/components/embed/media_with_companion_windows_component.html.erb
+++ b/app/components/embed/media_with_companion_windows_component.html.erb
@@ -43,33 +43,33 @@
 
   <div class="media-component-body">
     <aside class="left-drawer collapse" data-media-target="leftDrawer" id="left-drawer">
-      <nav class="vert-tabs">
-        <button data-action="click->media#displayMetadata" data-media-target="leftButton" class="active" aria-label="About this item" aria-controls="about">
+      <nav class="vert-tabs" role="tablist">
+        <button data-action="click->media#displayMetadata" data-media-target="leftButton" class="active" aria-label="About this item" aria-controls="about" aria-selected="true" role="tab">
           <svg class="MuiSvgIcon-root" focusable="false" viewBox="0 0 24 24" aria-hidden="true"><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm1 15h-2v-6h2v6zm0-8h-2V7h2v2z"></path></svg>
         </button>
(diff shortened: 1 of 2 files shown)


app/components/embed/media_with_companion_windows_component.html.erb
```diff
@@ -43,33 +43,33 @@
 
   <div class="media-component-body">
     <aside class="left-drawer collapse" data-media-target="leftDrawer" id="left-drawer">
-      <nav class="vert-tabs">
-        <button data-action="click->media#displayMetadata" data-media-target="leftButton" class="active" aria-label="About this item" aria-controls="about">
+      <nav class="vert-tabs" role="tablist">
+        <button data-action="click->media#displayMetadata" data-media-target="leftButton" class="active" aria-label="About this item" aria-controls="about" aria-selected="true" role="tab">
```
jcoyne, line 47: Can you please add the logic to change the `aria-selected` when one of the other tabs is pressed?

jcoyne (changes requested):

app/components/embed/media_with_companion_windows_component.html.erb
aaron-collier: Done.

app/javascript/packs/media.js
```diff
@@ -15,4 +15,25 @@ $(document).ready(function() {
...
+
+  // Add a click event handler to each tab
+  tabs.forEach((tab) => {
+    tab.addEventListener("click", changeTabs);
+  });
 });
+
+function changeTabs(e) {
+  const target = e.target;
+  const parentButton = target.parentNode;
```
jcoyne, line 29: This seems a bit fragile as it's tightly bound to the structure of the dom.  Can you use `.closest('[role="tabitem"]')` here and same for `tablist` below?

aaron-collier: Sure - I think this makes sense for the tablist select, I'm curius about this for parent... the issue is that the click event happens on the svg object, not the button, so I need to grab the parent button of the svg item... would `closest` be sufficient? And is `tabitem` correct vs `tab`?

I'll work through it and see what happens.

aaron-collier: Figured it out, I think it does need to be `tab` :-)

jcoyne (approved):

PR #1698, opened 2023-11-07
Combine the tab changing logic

jcoyne:
We don't want it spread out over two different places


aaron-collier (approved):
LGTM. I think a suggestion in my PR would have been helpful (though maybe it wasn't clear there was duplication at the time) for those of us pretty new to the codebase and patterns.

jcoyne:
@aaron-collier sorry, about that I didn't want to discourage you by making you do the moving, so I just did a follow up refactor.

aaron-collier:
@jcoyne no worries, I like this change and approach... Personally, I'm open to refactoring suggestions, but maybe that's not super common :-)

PR #1713, opened 2023-11-08
Set appropriate aria-role, aria-selected, and aria-controls for the content thumbnails

### LINKED ISSUE #1688: Media content window thumbnails should get aria roles tab/tablist
and aria-controls attribute

aaron-collier:
Closes #1688

### DIFF
--- a/app/components/embed/media_with_companion_windows_component.html.erb
+++ b/app/components/embed/media_with_companion_windows_component.html.erb
@@ -87,7 +87,7 @@
         </section>
       </div>
     </aside>
-    <div class='sul-embed-body sul-embed-media'>
+    <div class='sul-embed-body sul-embed-media' id='sul-embed-body'>
       <%= render Embed::MediaTagComponent.with_collection(resources_with_primary_file,
                                                           include_transcripts:,
                                                           druid:,
--- a/app/components/embed/media_with_companion_windows_component.html.erb
+++ b/app/components/embed/media_with_companion_windows_component.html.erb
@@ -76,7 +76,7 @@
             <h3>Media Content</h3>
           </div>
 
-          <ul class="contentsList" data-media-tag-target="list"></ul>
+          <ul class="contentsList" data-media-tag-target="list" role="tablist"></ul>
         </section>
         <section id="rights" data-media-target="rights" role="tabpanel" hidden>
           <div class="header-background">
--- a/app/javascript/src/controllers/thumbnail_controller.js
+++ b/app/javascript/src/controllers/thumbnail_controller.js
@@ -8,5 +8,15 @@ export default class extends Controller {
     evt.preventDefault();
     const event = new CustomEvent('thumbnail-clicked', { detail: { index: evt.params.index } })
     window.dispatchEvent(event)
+
+    const tabs = this.element.closest('[role="tablist"]').querySelectorAll('[role="tab"]')
+    tabs.forEach((target) => {
+      target.classList.remove('active')
+      target.setAttribute("aria-selected", false)
+    })
+
+    const target = this.element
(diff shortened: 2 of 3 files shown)


app/components/embed/media_with_companion_windows_component.html.erb
```diff
@@ -76,7 +76,7 @@
             <h3>Media Content</h3>
           </div>
 
-          <ul class="contentsList" data-media-tag-target="list"></ul>
+          <ul class="contentsList" data-media-tag-target="list" role="tablist" data-controller="thumbnail"></ul>
```
jcoyne, line 79: This data-controller should not be here. It's on the list items themselves.  You might consider adding a new controller for the list.

app/javascript/src/controllers/thumbnail_controller.js
```diff
@@ -1,12 +1,21 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
-  static targets = [ ]
+  static targets = [ "mediaSliderThumb" ]
```
jcoyne, line 4: I think I'd do this without a target and just get all the sibling tabs via `evt.target.closest('[role="tablist"]').querySelectorAll('[role="tab"]')`

app/javascript/src/modules/media_thumbnail_builder.js
```diff
@@ -28,7 +28,7 @@ export default function(dataset, index) {
     const fileLabel = dataset.fileLabel || '';
     const duration = dataset.duration || '';
     // Note: the "position: relative" is required for the stretched-link style.
-    return `<li class="sul-embed-media-slider-thumb ${activeClass}" data-controller="thumbnail" data-action="click->thumbnail#activate" data-thumbnail-index-param="${index}" style="position: relative;">
+    return `<li class="sul-embed-media-slider-thumb ${activeClass}" data-action="click->thumbnail#activate" data-thumbnail-index-param="${index}" data-thumbnail-target="mediaSliderThumb" style="position: relative;" aria-controls="sul-embed-body" role="tab">
```
jcoyne, line 31: finally, restore the data-controller here.

jcoyne (changes requested):

app/components/embed/media_with_companion_windows_component.html.erb
aaron-collier: Ah, gotcha. I had to move it up so that the targets were in scope but your suggestion below moves away from that. I'll give it a whirl. 👍

app/javascript/src/controllers/thumbnail_controller.js
aaron-collier: Done, this made for a bit of a simpler change, thanks. :-)

app/javascript/src/controllers/thumbnail_controller.js
```diff
@@ -8,5 +8,15 @@ export default class extends Controller {
...
     const event = new CustomEvent('thumbnail-clicked', { detail: { index: evt.params.index } })
     window.dispatchEvent(event)
+
+    const tabs = evt.target.closest('[role="tablist"]').querySelectorAll('[role="tab"]')
+    tabs.forEach((target) => {
+      target.classList.remove('active')
+      target.setAttribute("aria-selected", false)
+    })
+
+    const target = evt.target.closest('[role="tab"]')
```
jcoyne, line 18:
```suggestion
    const target = this.element
```

jcoyne (changes requested):

aaron-collier:
@jcoyne done, thanks.

jcoyne (approved):

PR #1721, opened 2023-11-09
Add locked-media-poster

### LINKED ISSUE #1706: Add default poster for locked media with no thumbnail
When media objects are (b)locked due to access restrictions, an access restriction message is shown to users. Also, an empty (blank white) native video player is still rendered. This player appears to users as broken: non-interactive, non-functional.
- When the media object has a corresponding thumbnail image, use this thumbnail as the "poster" rendered in the native video player. (NB: the player should remain non-functional, as the media content is still locked.)
- When the media object has no thumbnail image, use the default `locked-media-poster.svg` thumbnail for locked media attached below.
Image

jcoyne:
Fixes #1706


mjgiarlo (approved):

mjgiarlo:
@jcoyne Approved w/ a conflict

Issue #1683, opened 2023-11-06
Send analytics data to the metrics service


thatbudakguy:
We need to use [ahoy.js](https://github.com/ankane/ahoy.js) to communicate with the [new metrics API](https://github.com/sul-dlss/metrics).

For all viewers, we want to track:
- views (when the `/iframe` response is initially rendered, call `ahoy.trackView()`)
- downloads (call `ahoy.track('download', ...)`)
- where the event originates from (i.e. the domain outside the iframe)

jcoyne:
@thatbudakguy should download events be tracked via Stacks?  Not all downloads have to originate from sul-embed.

thatbudakguy:
is there a way to distinguish "the user asked to download this" from "some other program is requesting the file" in stacks?

it looks like we set the `download` attribute (in the file viewer at least) on the download links; is that what powers [the download detection in stacks](https://github.com/sul-dlss/stacks/blob/561e51eb4842206d325dafa103758ec3139fb910/app/controllers/file_controller.rb#L32-L36)?

jcoyne:
yes, that attribute impacts the content-disposition. I'm wondering if the users want to know when their file is downloaded when someone posts a link to social media and doesn't get it from purl.

thatbudakguy:
it sounds like that might not be too hard to do. i'll run this by the workcycle team and test it out locally.

jcoyne:
I'm not sure what the "it" is you're referring to.

thatbudakguy:
I guess what I'm worried about is platforms using the stacks URL to serve, display, or otherwise use the content that isn't "the user asked to download this". Don't we need to make sure to only track a file download if the user was explicitly invoking a download action? the presence of the download attribute seems like the right way to unambiguously indicate that.

thatbudakguy:
@jcoyne looking at stacks a bit more, I can see now that there's:
* `/object/:id`, which is `ObjectsController#show`, that downloads a .zip of all the files
* `/file/:id/*file_name`, which is `FilesController#show`, that serves a single file

it seems like the first one, at least, is an unambiguous "this object was downloaded", since it's the whole object, and the user gets a zipfile. the second one i'm less sure about, because if the disposition _wasn't_ provided, we could just be serving the file for use in a viewer or some other thing that isn't an explicit download to the user's machine.

what do you think?

thatbudakguy:
[...]

PR #1742, opened 2023-11-13
Send view events to the SDR metrics service when embedding

thatbudakguy:
Ref #1683

### DIFF
--- a/app/views/embed/_analytics.html.erb
+++ b/app/views/embed/_analytics.html.erb
@@ -16,3 +16,24 @@
   }
   gtag('config', 'G-80J719XWE5', args)
 </script>
+<% if Settings.metrics_api_url %>
+<script src="https://unpkg.com/ahoy.js@0.4.3"
+  integrity="sha384-s9X57Wbu8jIIErBMhudZJEAS1ZkelgbMI7/HUqsgjFBBm9IXyfbpKQ+ZoWUUvpgE"
+  crossorigin="anonymous"></script>
+<script>
+  // Set up Ahoy
+  ahoy.configure({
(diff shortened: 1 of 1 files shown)


app/views/embed/_analytics.html.erb
```diff
@@ -16,3 +16,20 @@
   }
   gtag('config', 'G-80J719XWE5', args)
 </script>
+<script src="https://unpkg.com/ahoy.js"></script>
```
jcoyne, line 19: Can we pin to a specific version?  Should we add an integrity attribute https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity ?

thatbudakguy: Both good suggestions! Added.

thatbudakguy:
related config change: https://github.com/sul-dlss/shared_configs/pull/2107/files

config/settings.yml
```diff
@@ -5,6 +5,7 @@ purl_conn_timeout: 2
 purl_feedback_email: 'some-email-address@distribution-list-host.stanford.edu'
 stacks_url: 'https://stacks.stanford.edu'
 iiif_info_url: 'https://library.stanford.edu/iiif/viewers'
+metrics_api_url: 'https://sdr-metrics-api-prod.stanford.edu'
```
jcoyne, line 8: Do we want to configure all of our instances to send the data to the prod metrics server?  It seems this even will send the data I generate by running it locally.

thatbudakguy: Updated so that it's not provided by default, and if there's no value set, we don't track views.

jcoyne (approved):

PR #1750, opened 2023-11-15
Aria label for download must include count displayed in viewer

### LINKED ISSUE #1723: Download Link - Visible label and accessible name do not match
Site Improve reports the following accessibility issue with the PDF viewer:

Visible label and accessible name do not match; however, that's slightly misleading. In actuality, the label text (visual) must be included in the accessible name of the element. I suggest the following fix:

Remove the `aria-label` on the `span` element. Change the `aria-label` on the `button` element to 'Download 1 file' (or whatever the number is, but for the PDF viewer, I believe it will always be 1). Here are the before and after code snippets tested with Site Improve. 

Currently:
Screenshot 2023-11-10 at 10 40 31 AM

Recommended Change:
Screenshot 2023-11-10 at 10 31 41 AM

aaron-collier:
Closes #1723

### DIFF
--- a/app/components/embed/footer_component.html.erb
+++ b/app/components/embed/footer_component.html.erb
@@ -11,9 +11,6 @@
     <% end %>
 
     <% if viewer.show_download? %>
-      <% file_count = viewer.purl_object.downloadable_files.length %>
-      <% download_file_label = 'No files available for download' %>
-      <% download_file_label = "#{file_count} #{'file'.pluralize(file_count)} available for download" if file_count > 0 %>
       <button
         class='sul-embed-footer-tool sul-embed-btn sul-em bed-btn-toolbar sul-embed-btn-default sul-i-download-3'
         aria-expanded='false' aria-label='<%= download_file_label %>'
--- a/app/components/embed/footer_component.html.erb
+++ b/app/components/embed/footer_component.html.erb
@@ -11,14 +11,15 @@
     <% end %>
 
     <% if viewer.show_download? %>
+      <% file_count = viewer.purl_object.downloadable_files.length %>
+      <% download_file_label = 'No files available for download' %>
+      <% download_file_label = "#{file_count} #{'file'.pluralize(file_count)} available for download" if file_count > 0 %>
       <button
         class='sul-embed-footer-tool sul-embed-btn sul-em bed-btn-toolbar sul-embed-btn-default sul-i-download-3'
-        aria-expanded='false' aria-label='open download panel'
+        aria-expanded='false' aria-label='<%= download_file_label %>'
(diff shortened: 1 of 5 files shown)


jcoyne (changes requested):

app/components/embed/footer_component.html.erb
```diff
@@ -11,14 +11,14 @@
     <% end %>
 
     <% if viewer.show_download? %>
+      <% file_count = viewer.purl_object.downloadable_files.length if viewer.show_download_count? %>
+      <% download_file_label = file_count > 1 ? "#{file_count} files available for download" : "1 file available for download" %>
```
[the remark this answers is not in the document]
aaron-collier: @jcoyne good point, I think this may be a bit better now.

app/components/embed/footer_component.html.erb
```diff
@@ -11,14 +11,15 @@
     <% end %>
 
     <% if viewer.show_download? %>
+      <% file_count = viewer.purl_object.downloadable_files.length %>
+      <% download_file_label = 'No files available for download' %>
+      <% download_file_label = "#{file_count} #{'file'.pluralize(file_count)} available for download" if file_count > 0 %>
```
jcoyne, line 16: Would you mind moving this code to the `.rb` part of the component?

jcoyne (changes requested):

app/components/embed/footer_component.html.erb
jcoyne: And maybe consider using the i18n pluralization tool for doing this: https://guides.rubyonrails.org/i18n.html#pluralization

aaron-collier: @jcoyne added a commit to use i18n, good suggestion, I'm happy to address any further concerns/suggestions.

app/components/embed/footer_component.rb
```diff
@@ -4,8 +4,10 @@ module Embed
   class FooterComponent < ViewComponent::Base
     def initialize(viewer:)
       @viewer = viewer
+      @file_count = @viewer.purl_object.downloadable_files.length
+      @download_file_label = I18n.t :download_files, count: @file_count
```
jcoyne, line 8: Why is this an ivar rather than a method call?

mjgiarlo: Both of these values hang off the viewer, so both seem ripe for ditching the ivar and replacing the `attr_reader`s with methods.

mjgiarlo (approved):

mjgiarlo:
@jcoyne Have another look?

jcoyne (approved):

PR #1754, opened 2023-11-16
Improve caption labels when a resource has multiple captions with languages

### LINKED ISSUE #1527: Disambiguate "Video file" in download list of media objects
https://purl.stanford.edu/bd111gd4290

Thumbs should probably not be made available in the download list for world objects (it’s confusing because they have the same title as the streaming video file). Need to change the way thumbs are either generated post-accessioning, or the way they’re handled in the SMPL accessioning process. Cocina has a field that allows you to label the thumb file (every file has a label field now) differently but none of the displays are capable of reading it yet.

mjgiarlo:
Fixes #1527 

- Improve caption labels when a resource has multiple captions with languages
- Add a ViewComponent preview for media with multiple language captions

- [ ] Add logic around thumbnails


mjgiarlo:
Superseded.

Issue #1531, opened 2023-10-11
Indicate Stanford-only access after authentication on media objects


jcoyne:
### Expected behavior
There should be an indicator that the media item requires authentication even after you have authenticated so you know that when you share this URL, the user on the other end may not have access.

### Actual behavior

Once you are logged in there is no indicator that you needed to log in to see this.

### Browser / Environment

See https://github.com/sul-dlss/media/issues/152

alundgard:
- See the Access Banner Prototype in the [Figma Design](https://www.figma.com/file/ADNitlk4aWHDiPDWD1dHdF/Media-Work-Cycle---Media-Player?type=design&node-id=0-1&mode=design). [To launch the interactive prototype in Figma, click the "Present" play arrow in the top-right corner. From this tab, click Access Banner Prototype under the "Flows" panel.]
- The current design approach introduces an "authentication tab" as a persistent indicator.
![Image](https://github.com/sul-dlss/sul-embed/assets/2343316/84343596-bb3b-4ee1-9801-d92b6d9c09a6)
- Contact me for additional context and/or design revisions due to technical feasibility.

mjgiarlo:
Simpler approach: a persistent banner across the tab (i.e., not collapsible). That is also fine.

alundgard:
Here's a [Mirador example](https://purl.stanford.edu/bb627py6670) of the simpler approach (persistent banner across the top).

PR #1763, opened 2023-11-16
[WIP] indicate stanford-only restriction even after login

jmartin-sul:
this is the most basic approach to dealing with #1531, from which we can iterate and do something more like the ideal mockup.  (per [comment on the ticket](https://github.com/sul-dlss/sul-embed/issues/1531#issuecomment-1789250403): "Simpler approach: a persistent banner across the tab (i.e., not collapsible). That is also fine.")

This has one bit of UI weirdness that makes the player harder to use, and if fixing that is difficult, or even just not-super-simple, maybe we should just go ahead and do the slightly more complex thing the mockup suggests.

TODO to get this out of draft:
- [ ] fix the UI issue where the post-login banner pushes up the menu bar with the hamburger icon that lets you open the companion window (see screenshot)
- [ ] get an SVG for the unlocked icon and replace the locked icon SVG with that after login (or have both included in the page initially, with the unlocked icon initially hidden, and the hidden state of the locked and unlocked icons toggled after successful login)
- [ ] have the banner display on initial page load when already logged in (noticed the lack of display in this case when i went to see if i could reproduce something i screenshotted where the post-login banner was displayed and everything about the embed window was visible, but i suspect that was actually the immediate post-login state, with the captions slightly pushed below the viewable area to start; once play is hit, the whole video player becomes visible, and pushes the menu bar up, causing it to be partially obscured)

Screen Shot 2023-11-16 at 12 02 50 PM

The menu still works, it's just hard to see the icon to open it:
Screen Shot 2023-11-16 at 12 03 22 PM


jmartin-sul:
pointers for logout link implementation:

- example from mirador from alan: https://purl.stanford.edu/bb627py6670
- mike says see H2 -- https://stanfordlib.slack.com/archives/C05SVRXGER1/p1700590285579929
  - First, configure webauth library with the magic shib URL: https://github.com/sul-dlss/happy-heron/blob/362f154952b856e6d861c38280b0f2906bb07ae2/config/initializers/devise_remote_user.rb#L13
  - Next, wire up devise to destroy the session: https://github.com/sul-dlss/happy-heron/blob/362f154952b856e6d861c38280b0f2906bb07ae2/config/routes.rb#L9-L11
  - Last, create a link in the UI: https://github.com/sul-dlss/happy-heron/blob/362f154952b856e6d861c38280b0f2906bb07ae2/app/views/layouts/application.html.erb#L76

jmartin-sul:
question for @alundgard:

What should I do about the logged in banner either pushing the player controls under the bottom of the embed iframe in PURL, or pushing the hamburger icon for the companion window above it (which behavior occurs depends on whether the user just logged in, or already logged in before loading the page).

From post-standup discussion, the options seemed like:
* shrink the video so that it fits in with the banner without chopping off the controls or laying the banner over the video
* try the fancier design from the first mockup, with the unlock icon off to the side, and either minimized or expanded with companion window, overlayed on video (jcoyne observes: downside of this approach is that on iphone size screens or for wide aspect ratio videos, you'll always be covering part of the video)
* put a close button on the banner after it's initially shown?  but then what to do about presenting a log out button?

alundgard:
- I think this makes the most sense: "shrink the video so that it fits in with the banner without chopping off the controls or laying the banner over the video." Player controls should remain visible.
- Since we're going with the "simple" persistent banner for now, adding a banner "close/dismiss/X" button after login could be a good compromise design. I'll think more about this possibility, and check with the POs. Thanks!

app/assets/stylesheets/media.scss
```diff
@@ -275,9 +275,8 @@
     flex-shrink: 0;
   }
 
-  // First selector is for embargo. Last selector is other auth restrictions
-  .authLinkWrapper + .media-component-body, [data-controller="auth-restriction"]:has(.authLinkWrapper:not([hidden])) + .media-component-body {
-    max-height: calc(100vh - 88px); // Scroll height for metadata when the auth wrapper is displaying
+  [data-controller="auth-restriction"]:not([hidden]) + .media-component-body {
```
jmartin-sul, line 278: without the selector simplification here and the associated HTML and JS updates (motivation described in more detail the last commit message), i found that the video player wouldn't expand to fill the space left by the dismissed banner.  also, the video player would be a bit jumpy on page load, and would be shrunk as if the banner was already displayed, even if the banner had yet to be shown because a response from stacks had yet to be received.

the fix associated with this CSS tweak seems to smooth things out so that the video player consistently shrinks when auth message is displayed, and expands when it's not.

example of the incorrectly unexpanded player before the fix in the last commit:
Screen Shot 2023-11-29 at 10 11 23 PM

app/assets/stylesheets/media.scss
```diff
@@ -275,9 +275,8 @@
...
+    max-height: calc(100vh - 100px); // Scroll height for metadata when the auth wrapper is displaying
```
jmartin-sul, line 279: if i left it at `88px`, the bottom of the control bar was still a bit cut off, e.g.

Screen Shot 2023-11-29 at 10 00 02 PM

vs after this fix:
Screen Shot 2023-11-29 at 10 00 55 PM

app/assets/stylesheets/media.scss
```diff
@@ -331,6 +330,14 @@
...
       text-transform: uppercase;
     }
+
+    button.dismissButton {
+      padding: 0rem;
+
+      svg {
+        font-size: 0.85rem;
+      }
+    }
```
jmartin-sul, line 340: this styling was just the result of a bit of tweaking till i got something that i thought looked slightly better than the default.  very open to suggestions here.

app/javascript/src/controllers/media_tag_controller.js
```diff
@@ -25,8 +25,9 @@ export default class extends Controller {
   afterValidate(result, completeCallback) {
     if (result.authResponse.status === 'success') {
       this.initializeVideoJSPlayer()
-      const event = new CustomEvent('auth-success')
-      window.dispatchEvent(event)
+      if (result.authResponse.access_restrictions.stanford_restricted === true)
```
jmartin-sul, line 28: this is what takes advantage of https://github.com/sul-dlss/stacks/pull/1049 to display the message that the user is logged in on a fresh page load, where they visit something after having already logged in on another page.

app/javascript/controllers/auth_restriction_controller.js
```diff
@@ -31,26 +31,30 @@ export default class extends Controller {
     this.stanfordRestrictionDismissButtonTarget.hidden = false
   }
 
-  dismissStanfordRestriction() {
-    this.stanfordRestrictionTarget.hidden = true
+  hideAuthRestrictionMessages() {
+    this.authRestrictionMessagesContainerTarget.hidden = true
```
jcoyne, line 35: Could this be `this.element.hide = true`?  I don't think we need a special target for the container itself.

jmartin-sul: fixed

app/assets/stylesheets/media.scss
jmartin-sul: ok, after a bunch of uncertainty, i did what i thought was the safest thing to move this PR along without introducing a regression: i put the old selector back.  but i made two small tweaks: i flipped the order, because that felt more natural given the order of the DOM elements in `media_with_companion_windows_component.erb`, and i updated the comment because i'm pretty sure the selector i added back is for citation-only.

per stereotypically wordy slack reply, there's a corner case with two banners that neither the old CSS nor the slightly re-jiggered CSS seem to account for.  but i don't think this PR makes that worse, so maybe we can circle back on that in a follow up if need be.

aaron-collier (approved):

Issue #1753, opened 2023-11-16
Create mapping of BCP 47 tags to label strings for file languages


mjgiarlo:
We use BCP 47 tags in file languages, and when a resource has e.g. multiple captions in different languages, we want to render a user-friendly label like `English captions` or `Russian captions`. To accomplish this, given we have no reliably way to translate BCP 47 tags into label strings, we'll roll with a mapping for the most common language tags and their labels, such as:

```
en: English
ru: Russian
```

This is for @dinahhandel and @andrewjbtw

hudajkhan:
I think this also relates to #1588 since the captions options are identified by BCP 47 codes in the PURL and the labels for these need to be displayed in the captions options in the video player.

dinahhandel:
Here's what we should start with when it comes to language mappings:

en: English
ru: Russian
de: German
et: Estonian
lv: Latvian
es: Spanish

These are the only mappings we currently need.

dinahhandel:
Is this ticket still relevant, given #1765?

mjgiarlo:
@dinahhandel I think we can close it with the comment above.

PR #1765, opened 2023-11-17
handle multiple caption files

### LINKED ISSUE #1588: Viewer should be able to choose captions in different languages
We need a test object for this.

No longer blocked by #1498 because even without the full settings interface, the user is presented with a choice of language.

hudajkhan:
What this pull request does:
- Allow for identification of multiple caption files associated with a resource
- Enable creation of video track elements for all the caption files with the different languages specified
- If no language code is specified in the PURL XML for the caption file, the language will be designated as English by default
- Adding language to the attributes that can be retrieved for a resource file
- For display in the captions list, the languages list is sorted by caption language label

What this pull request does not do:
- Incorporate a complete BCP 47 language code mapping.  For now, the mappings rely on the list identified here: https://github.com/sul-dlss/sul-embed/issues/1753#issuecomment-1814967459 

Closes #1588


app/components/embed/media_tag_component.rb
```diff
@@ -87,7 +87,27 @@ def enabled_streaming_sources
...
+        file.vtt.sort_by { |cfile| caption_language(cfile.language) }.map do |caption_file|
+          lang_code = caption_file.language || 'en'
+          lang_label = caption_language(lang_code)
+          tag.track(src: caption_file.file_url, kind: 'captions', srclang: lang_code, label: lang_label)
+        end
+      )
+    end
+
+    def caption_language(language_code)
+      case language_code
```
jcoyne, line 103: How about `LANGUAGE_LABELS.fetch(language_code, 'Caption')` instead of the `case`?

app/models/embed/purl/resource_file.rb
```diff
@@ -62,14 +62,20 @@ def thumbnail?
         Settings.resource_types_that_contain_thumbnails.include?(resource.type)
       end
 
+      # Returns array of VTT files
       def vtt
```
jcoyne, line 66: let's change the method name to `vtts` so it appears as plural  or maybe `text_tracks` is better.

app/components/embed/media_tag_component.rb
```diff
@@ -87,7 +87,27 @@ def enabled_streaming_sources
     def transcript
       return unless render_captions?
 
-      tag.track(src: file.vtt.file_url, kind: 'captions', srclang: 'en', label: 'English')
+      # A video clip may have multiple caption files in different languages.
+      # We want to enable the user to select from any of these options.
+      # We also want the different language options to be listed alphabetically.
+      safe_join(
+        file.vtt.sort_by { |cfile| caption_language(cfile.language) }.map do |caption_file|
```
jcoyne, line 94: How about extracting this first part to a method: `sorted_text_tracks`?

app/components/embed/media_tag_component.rb
hudajkhan: This is a filler method so it's just intended to be temporary. Mike's working on a related issue so I'll wait to use his approach for getting labels based on language code.

app/models/embed/purl/resource_file.rb
hudajkhan: sounds good. I may say "captions" or similar and am considering looking for the "role" attribute instead of or in addition to mime type.

app/models/embed/purl/resource_file.rb
```diff
@@ -62,14 +62,20 @@ def thumbnail?
...
+        resource.files.select(&:vtt?)
       end
 
       def vtt?
         mimetype == 'text/vtt'
       end
 
+      # Returns language if specified by language tag
+      def language
+        @file.attributes['language'].try(:value)
```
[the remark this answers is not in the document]
hudajkhan: The intent here is to check if the language attribute exists at all, so I can change this to be more specific and not just see if the value exists.

app/components/embed/media_tag_component.rb
hudajkhan: Sounds good

app/models/embed/purl/resource_file.rb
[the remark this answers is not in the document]
hudajkhan: With the latest refactoring, this will go away entirely anyway

app/components/embed/media_tag_component.rb
hudajkhan: I am going to rewrite the filler method and leave a comment there that it should follow the caption code to label process from @mjgiarlo 's upcoming changes.

mjgiarlo (approved):

mjgiarlo:
LGTM, thanks for working on this!

PR #1771, opened 2023-11-17
Decouple ResourceFile from Resource 

jcoyne:
This will allow us to decouple the file from the resource

### DIFF
--- a/app/models/embed/purl/resource_file.rb
+++ b/app/models/embed/purl/resource_file.rb
@@ -3,21 +3,22 @@
 module Embed
   class Purl
     class ResourceFile
-      # @param [Embed::Purl::Resource] resource
+      # @param [String] druid the identifier of the resource this file belongs to.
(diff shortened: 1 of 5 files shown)


app/models/embed/purl/resource_file.rb
```diff
@@ -3,21 +3,22 @@
 module Embed
   class Purl
     class ResourceFile
-      # @param [Embed::Purl::Resource] resource
+      # @param [String] druid the identifier of the resource this file belongs to.
+      # @param [String] description the label for this file
```
mjgiarlo, line 7: Should we consider calling this `label`?

mjgiarlo (approved):

app/models/embed/purl/resource_file.rb
jcoyne: I'll get to that in the next PR.

PR #1775, opened 2023-11-17
Extract xml deserialization from the ResourceFile model

jcoyne:


app/models/embed/purl/resource_file.rb
```diff
@@ -3,27 +3,20 @@
...
+        yield(self) if block_given?
       end
 
-      attr_accessor :index
-
-      def label
-        @description
+      def attributes=(hash)
+        hash.each do |key, value|
+          send("#{key}=", value)
```
mjgiarlo, line 13: Maybe a bit paranoid, but should we consider changing this to `#public_send`?

mjgiarlo (approved):

app/models/embed/purl/resource_file.rb
jcoyne: Sure.

PR #1776, opened 2023-11-17
Use a delegator for the hierarchical file

jcoyne:
The hierarchy feature requires an index property, that is not relevant to ResourceFile itself, so a delegator can keep this concern separate

### DIFF
--- a/app/models/embed/hierarchical_contents.rb
+++ b/app/models/embed/hierarchical_contents.rb
@@ -22,7 +22,7 @@ module Embed
     attr_reader :resources
 
     def file_contents
-      resources.map(&:files).flatten
+      resources.map(&:files).flatten.map { |file| HierarchicalFile.new(file) }
     end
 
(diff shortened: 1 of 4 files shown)


app/models/embed/hierarchical_contents.rb
```diff
@@ -22,7 +22,7 @@ def contents
     attr_reader :resources
 
     def file_contents
-      resources.map(&:files).flatten
+      resources.map(&:files).flatten.map { |file| HierarchicalFile.new(file) }
```
mjgiarlo, line 25: Could we consider `flat_map` here?

mjgiarlo (approved):

app/models/embed/hierarchical_contents.rb
jcoyne: You could for `flat_map(&:files)`, but that isn't really the part I'm changing here.

PR #1778, opened 2023-11-17
Extract a resource deserializer

jcoyne:
So that the models don't have to know about their serialization

### DIFF
--- a/app/models/embed/purl/resource_xml_deserializer.rb
+++ b/app/models/embed/purl/resource_xml_deserializer.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module Embed
+  class Purl
+    class ResourceXmlDeserializer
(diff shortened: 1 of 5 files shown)


app/models/embed/purl/resource_xml_deserializer.rb
```diff
@@ -0,0 +1,29 @@
...
+        @druid = druid
+        @resource = resource
+        @rights = rights
+      end
+
+      def deserialize
+        description = @resource.xpath('./label').text
+        files = @resource.xpath('./file').map do |file|
+          FileXmlDeserializer.new(@druid, description, file, @rights).deserialize
+        end
```
mjgiarlo, line 19: These could be private methods... but I'd say that's a relatively low-value refactoring.

mjgiarlo (approved):

PR #1796, opened 2023-11-21
Separate the legacy media-tag from the new media-tag component

jcoyne:
This will make it possible for us to add the UV preview only to the media tag only in the new player.  It also allows us to simplify the MediaTagComponent for the new player as we don't have to maintain as much complexity there.


app/components/embed/legacy_media/media_tag_component.rb
```diff
@@ -0,0 +1,203 @@
+# frozen_string_literal: true
```
jcoyne, line 1: This a copy of the Embed::MediaTagComponent.

spec/components/embed/legacy_media/media_tag_component_spec.rb
```diff
@@ -0,0 +1,253 @@
+# frozen_string_literal: true
```
jcoyne, line 1: This a copy of the Embed::MediaTagComponent test

mjgiarlo (approved):

PR #1799, opened 2023-11-21
Add BCP 47 resolution

mjgiarlo:
This commit adds the ability to parse BCP 47 registries into records and look up labels given BCP 47 tag strings.


lib/bcp47.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Bcp47
+  require 'bcp47/parser'
+  require 'bcp47/record'
+  require 'bcp47/registry'
+end
```
mjgiarlo, line 7: Sticking this in `lib/` since it seems ripe for extraction to a new BCP 47 gem. The current gems for BCP 47 are outdated and unmaintained, as best I can tell.

lib/bcp47/registry.rb
```diff
@@ -0,0 +1,65 @@
...
+require 'faraday'
+
+module Bcp47
+  class Registry
+    CODE_DELIMITER = '-'
+    LABEL_DELIMITER = ', '
+    LANGUAGE_SUBTAG_REGISTRY_URL = 'https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry'
+
+    def self.records
+      @records ||= Parser.new(Faraday.get(LANGUAGE_SUBTAG_REGISTRY_URL).body).records
```
jcoyne, line 15: Can we split out the download part?  It seems better if we don't have to depend on a remote service.  Perhaps this could be a rake task to update the file that you keep in the repo.

jcoyne (changes requested):

lib/bcp47/registry.rb
mjgiarlo: Yes, that sounds good. I'll stick the file somewhere in `app`, I guess, and toss a simple rake task at the PR

mjgiarlo: @jcoyne Done.

jcoyne:
Needs a rebase.

mjgiarlo:
> Needs a rebase.

@jcoyne done, thanks!

.rubocop_todo.yml
```diff
@@ -18,10 +18,11 @@ Capybara/SpecificFinders:
   Exclude:
     - 'spec/features/embed_this_panel_spec.rb'
 
-# Offense count: 1
+# Offense count: 2
 # Configuration parameters: AllowedMethods, AllowedPatterns, CountRepeatedAttributes, Max.
 Metrics/AbcSize:
   Exclude:
+    - 'lib/bcp47/registry.rb'
```
jcoyne, line 25: Would you please move these to inline exclusions rather than adding to the TODO?

jcoyne (changes requested):

.rubocop_todo.yml
mjgiarlo: Sure.

jcoyne (approved):

PR #1810, opened 2023-11-21
Disambiguate file labels in download modal

### LINKED ISSUE #1527: Disambiguate "Video file" in download list of media objects

mjgiarlo:
Fixes #1527

This commit is primarily concerned with changing the generic "Video file" or "Audio file" file labels in the download modal. It includes:

* Concentrate caption language logic in the `ResourceFile` model
* In a few spots, be clearer about whether we're dealing with transcripts or captions (VTT)
* Move caption file sorting logic into `Resource#caption_files`
* Move logic about whether a `ResourceFile` is downloadable into that models
* Change references to `.webvtt` files to `.vtt` files
* Add `caption` role to `ResourceFile` factory
* Simplify ViewComponent previews by extracting the common bits, meaning less boilerplate copypasta for new previews

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -47,13 +47,11 @@ module Embed
     end
 
     def all_resource_files
-      contents.map(&:files).flatten
+      contents.flat_map(&:files)
     end
 
     def downloadable_files
-      all_resource_files.select do |rf|
-        rf.world_downloadable? || rf.stanford_only_downloadable?
-      end
+      all_resource_files.select(&:downloadable?)
     end
 
     def purl_url
--- a/app/models/embed/purl/resource_file.rb
+++ b/app/models/embed/purl/resource_file.rb
@@ -3,6 +3,18 @@
 module Embed
   class Purl
     class ResourceFile
+      NON_DOWNLOADABLE_ROLES = %w[thumbnail].freeze
+      # Hardcoding some language code to label mappings, based on the mappings we currently need.
+      # This approach should be revisited once we have more robust BCP 47 code to label mapping integrated.
+      FILE_LANGUAGE_CAPTION_LABELS = {
+        'en' => 'English',
+        'ru' => 'Russian',
+        'de' => 'German',
+        'et' => 'Estonian',
+        'lv' => 'Latvian',
+        'es' => 'Spanish'
+      }.freeze
+
       def initialize(attributes = {})
         self.attributes = attributes
         yield(self) if block_given?
@@ -14,7 +26,7 @@ module Embed
         end
       end
 
-      attr_accessor :druid, :label, :filename, :mimetype, :size, :duration, :language,
+      attr_accessor :druid, :label, :filename, :mimetype, :size, :duration, :language, :role,
                     :world_downloadable, :stanford_only, :location_restricted, :stanford_only_downloadable
 
       alias title filename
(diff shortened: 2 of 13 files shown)


app/models/embed/purl/resource_file.rb
```diff
@@ -40,15 +52,34 @@ def stacks_url
       end
 
       def label_or_filename
+        return caption_label if caption?
+
         label.presence || filename
       end
```
mjgiarlo, line 58: @jcoyne I know we discussed putting the caption label logic in the view layer, but the RF instance already seems to has some logic around what label to present to the world (in `#label_or_filename`). I could move the caption label logic *back* into the component, but then we're overriding what the RF asserts as its label or filename which strikes me as something we don't do elsewhere. Happy to be convinced, though, so we can get this work moving.

hudajkhan: This seems to fine to me.

hudajkhan (approved):
These changes make sense to me. I'll also try to test them locally (somehow) but for the moment, am basing this approval on reading over the changes.

app/models/embed/purl/resource_file.rb
```diff
@@ -40,15 +52,34 @@ def stacks_url
...
 
+      def language_label
+        FILE_LANGUAGE_CAPTION_LABELS[language_code]
```
jcoyne, line 61: should there be a default here, in case we find a lang we don't know about?

jcoyne (changes requested):

app/models/embed/purl/resource_file.rb
hudajkhan: In the earlier approach, if a language code exists but is not recognized in the mapping, the label returned is "Caption"

hudajkhan: "Caption" is also returned by the media tag component in this case

hudajkhan: The other behavior in the former approach was that if there was no language code, "en" would be returned. I see that in the language_code method on line 73 below so that seems fine.

mjgiarlo: @jcoyne Yes, because `language_code` defaults to `en`, the effective default for `language_label` is `English`. I believe Dinah has said before that English as the default caption language is preferred.

app/models/embed/purl.rb
```diff
@@ -47,13 +47,11 @@ def size
     end
 
     def all_resource_files
-      contents.map(&:files).flatten
+      contents.flat_map(&:files)
```
mjgiarlo, line 50: Do this in one step instead of two.

app/models/embed/purl.rb
```diff
@@ -47,13 +47,11 @@ def size
...
     end
 
     def downloadable_files
-      all_resource_files.select do |rf|
-        rf.world_downloadable? || rf.stanford_only_downloadable?
-      end
+      all_resource_files.select(&:downloadable?)
```
mjgiarlo, line 54: Let the RF be home of the conditions for downloadable or non-downloadable.

app/models/embed/purl/resource_file.rb
mjgiarlo: @jcoyne the last couple times we talked about it, Dinah opted for "Caption file" when the language is unknown, IIRC. I can patch in that change.

mjgiarlo: @jcoyne Made the change. Ready for another look.

jcoyne (approved):

PR #1821, opened 2023-11-22
Enhance rights panel in left companion window of media player

### LINKED ISSUE #1797: Media viewer rights panel should show an attribution to Stanford libraries
### Expected behavior

Following the example of Mirador, the media viewer rights panel should always show an attribution "Provided by Stanford Libraries" and include the Stanford libraries logo.

Mirador example:
>Screenshot 2023-11-21 at 8 57 00 AM
https://sul-purl-stage.stanford.edu/hk569bc9824

### Actual behavior

Currently, no attribution is shown and the result is that if rights statements have not been supplied, the rights panel is blank. Example: https://sul-purl-stage.stanford.edu/dp213qm6802

mjgiarlo:
Fixes #1797

This commit makes two changes to the rights panel in the LCW of the media
player, both of which bring look & feel parity with the Mirador viewer:

* Add the Stanford Libraries logo to the bottom of the panel
* When an object lacks rights information (use statement, copyright statement,
license), render a default attribution

Screenshot from 2023-11-22 12-44-10

### DIFF
--- a/app/assets/stylesheets/media.scss
+++ b/app/assets/stylesheets/media.scss
@@ -377,7 +377,18 @@
       font-family: "Roboto", "Helvetica", "Arial", sans-serif;
     }
 
+    .logoContainer {
+      padding: 1rem;
+      border-bottom: .5px solid rgba(0, 0, 0, 0.25);
+    }
+
+    .attributionLogo {
+      max-width: 100%;
+    }
+
     .metadataList, .rightsList {
+      border-bottom: .5px solid rgba(0, 0, 0, 0.25);
+
       dt {
         font-size: 0.878rem;
         font-weight: 500;
--- a/app/components/rights_component.html.erb
+++ b/app/components/rights_component.html.erb
@@ -9,8 +9,13 @@
   <% end %>
   <% if license.present? %>
     <dt>License</dt>
-    <dd>
-      <%= license %>
-    </dd>
+    <dd><%= license %></dd>
+  <% end %>
+  <% if no_rights_information_present? %>
+    <dt>Attribution</dt>
+    <dd><%= default_attribution %></dd>
   <% end %>
 </dl>
+<div class="logoContainer">
+  <%= image_tag stanford_libraries_logo_url, alt: "Stanford Libraries logo", role: "presentation", class: "attributionLogo" %>
+</div>
--- a/app/components/rights_component.rb
+++ b/app/components/rights_component.rb
@@ -9,4 +9,20 @@ class RightsComponent < ViewComponent::Base
 
   delegate :purl_object, to: :viewer
   delegate :use_and_reproduction, :copyright, :license, to: :purl_object
+
+  def no_rights_information_present?
+    use_and_reproduction.blank? &&
+      copyright.blank? &&
+      license.blank?
+  end
+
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/media.scss
```diff
@@ -374,7 +374,21 @@
       font-family: "Roboto", "Helvetica", "Arial", sans-serif;
     }
 
+    .logoContainer {
+      padding-top: 16px;
```
jcoyne, line 378: can we use relative units for padding (rems) as we do elsewhere in this file?

app/components/rights_component.html.erb
```diff
@@ -9,8 +9,13 @@
...
-    </dd>
+    <dd><%= license %></dd>
+  <% end %>
+  <% if no_rights_information_present? %>
+    <dt>Attribution</dt>
+    <dd><%= default_attribution %></dd>
   <% end %>
 </dl>
+<div class="logoContainer">
+  <img src="<%= stanford_libraries_logo_url %>" alt="Stanford Libraries logo" role="presentation" class="attributionLogo">
```
jcoyne, line 20: can we use image_tag here?

jcoyne (changes requested):

app/assets/stylesheets/media.scss
mjgiarlo: sure, will do

app/components/rights_component.html.erb
mjgiarlo: 👍🏻

mjgiarlo: Done.

app/assets/stylesheets/media.scss
mjgiarlo: Done.

app/components/rights_component.html.erb
```diff
@@ -9,8 +9,13 @@
...
-    </dd>
+    <dd><%= license %></dd>
+  <% end %>
+  <% if no_rights_information_present? %>
+    <dt>Attribution</dt>
+    <dd><%= default_attribution %></dd>
   <% end %>
 </dl>
+<div class="logoContainer">
+  <%= image_tag stanford_libraries_logo_url, alt: "Stanford Libraries logo", role: "presentation", class: "attributionLogo" %>
```
jcoyne, line 20: Can we link to `` instead?  This is what mirador does and it would prevent us from having to ship another copy of this rather large file. (chances are this one is already cached).

jcoyne (changes requested):

app/components/rights_component.html.erb
mjgiarlo: @jcoyne I considered it, but I have no idea how stable that ☝🏻 link is. I guess as long as we're confident that druid will remain published and that filename won't change, it's fine. It at least increases parity among the players?

jcoyne (approved):

app/components/rights_component.html.erb
andrewjbtw: I've never looked at it before but that druid is "SUL Logo 2015", which implies it might have been created for this purpose: https://purl.stanford.edu/wy534zh7137

PR #1832, opened 2023-11-27
Use the browser PDF viewer via object instead of pdf.js

peetucket:
Based on #1418, rebased on main, updated as needed to fix conflicts, and fixed issues with download/information panels being broken.

Part of #1834 

Uses the built-in browser PDF viewer (which is different for each browser) instead of the [PDF.js library ](https://mozilla.github.io/pdf.js/) which would provide the same experience across all browsers but which requires coding, creating your own UI and maintenance of this code.

~~Experiment to see if this approach is acceptable/makes sense.~~

After review, agreed this approach makes sense.  Instead of completely removing the current PDF viewer implementation, this renames the old component to `LegacyPdfComponent` and adds a feature flag in settings that allows us to toggle it back if needed.  It also adds some notes to the code about what to remove when no longer needed.

If accepted and merged, the new browser based PDF viewer would immediately become the default.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -62,7 +62,7 @@ analytics_debug: true
 min_files_to_search_default: 10
 enabled_features:
   transcripts: false
-  legacy_pdf_viewer: false
+  legacy_pdf_viewer: true # true shows the PDF.js based PDF viewer, false uses <object> tag for browser PDF rendering
 language_subtag_registry:
   path: vendor/data/language-subtag-registry
   url: 'https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry'


peetucket:
### Mac:

**Chrome:**

Screen Shot 2023-11-27 at 2 24 14 PM

**Firefox:**

Screen Shot 2023-11-27 at 2 24 23 PM

**Microsoft Edge:**

Screen Shot 2023-11-27 at 2 25 26 PM

app/components/embed/pdf_component.html.erb
```diff
@@ -2,27 +2,8 @@
...
-      <canvas></canvas>
-      <button
-        aria-label="Exit full screen"
-        class="sul-embed-btn sul-i-close sul-close-fullscreen-button"
-        id="close-full-screen-button">
-      </button>
-    </div>
     <%= javascript_pack_tag('pdf_viewer') %>
+    <embed src="<%= viewer.pdf_files.first %>" style="height: 100vh; width: 100%">
   </div>
```
peetucket, line 7: replace with embed tag

app/javascript/packs/pdf_viewer.js
```diff
@@ -3,13 +3,11 @@
...
 import { PopupPanels } from '../src/modules/popup_panels.js';
 import CommonViewerBehavior from '../src/modules/common_viewer_behavior.js';
-import PDFViewer from '../src/modules/pdf_viewer.js';
 import Fullscreen from '../src/modules/fullscreen.js';
 
 $(document).ready(function() {
   CommonViewerBehavior.initializeViewer();
   PopupPanels.init();
   EmbedThis.init();
-  PDFViewer.init();
```
peetucket, line 13: remove pdf.js import and init

app/javascript/src/modules/pdf_viewer.js
peetucket, line 1: remove our custom pdf related javascript

package.json
```diff
@@ -40,7 +40,6 @@
     "mirador-dl-plugin": "^0.13.0",
     "mirador-image-tools": "^0.8.0",
     "mirador-share-plugin": "^0.11.0",
-    "pdfjs-dist": "^4",
     "prop-types": "^15.7.2",
```
peetucket, line 43: remove pdf.js package

peetucket:
Note: search within PDF only seems to be supported in Firefox and Edge, I don't see it in Chrome or Safari.

The object tag also works fine and allows for a fallback if the browser doesn't support inline PDFs (you can include a link to download the PDF in it's place.

spec/features/pdf_viewer_spec.rb
```diff
@@ -12,11 +12,7 @@
   end
 
   it 'renders the PDF viewer for documents' do
-    expect(page).to have_css('.sul-embed-pdf #pdf-viewer')
-  end
-
-  it 'renders a loading spinner' do
-    expect(page).to have_css('.loading-spinner')
+    expect(page).to have_css('.sul-embed-pdf')
```
peetucket, line 15: we don't have a loading spinner anymore

config/settings.yml
```diff
@@ -62,6 +62,7 @@ analytics_debug: true
 min_files_to_search_default: 10
 enabled_features:
   transcripts: false
+  legacy_pdf_viewer: false
```
jcoyne, line 65: So, by default we're going to use the new implementation in all environments?  The feature here is to go back to the old one?

app/javascript/packs/pdf_viewer.js
jcoyne: I'm not sure you can have the legacy pdf viewer work if you modify this file.  I think you either need to remove the legacy pdf component or make a new pack for the new pdf viewer.

jcoyne (changes requested):

app/javascript/packs/pdf_viewer.js
peetucket: Indeed it would not.  That is why I moved the original version of this file to `app/javascript/packs/legacy_pdf_viewer.js` (see above) and reference that pack in the `LegacyPdfComponent`.  If/when we remove the Legacy component, we can just delete the entire legacy pack file (as described in the comments in `app/components/embed/legacy_pdf_component.rb`).

config/settings.yml
andrewjbtw: That's not quite what I meant. I meant to move to the browser-based on stage and UAT but not yet to prod.

peetucket: Ahh, ok. So I guess (1) keep this PR but make Legacy the default for now via the feature flag, (2) update this PR to keep the current component names the same and create a "NewPdfComponent" that can be turned on with the feature flag (similar to media).  Net effect is the same, just so naming semantics that come down to the likelihood that we will switch to the new method sooner rather than later.

PR #1845, opened 2023-11-28
Reimplement file viewer as a tree grid.

justinlittman:
closes #801


app/javascript/controllers/tree_controller.js
```diff
@@ -0,0 +1,230 @@
...
+    const count = this.countTarget.textContent = this.tableTarget.querySelectorAll('tbody > tr[data-tree-role="leaf"]:not(.search-hidden)').length
+    if (count == 1) {
+      this.countTarget.textContent = '1 file'
+    } else {
+      this.countTarget.textContent = `${count} files`
+    }
+  }
+
+  restripe() {
+    this.visibleRows.forEach((rowElement, index) => {
```
jcoyne, line 222: I believe this can be done in plain css, right?

jcoyne (changes requested):

app/javascript/controllers/tree_controller.js
justinlittman: Unfortunately, I don't think CSS striping works in this case since it needs to adjust the striping based on what is visible.

app/javascript/controllers/tree_controller.js
```diff
@@ -0,0 +1,230 @@
...
+    }
+    return null
+  }
+
+  isVisible(element) {
+    return !this.isHidden(element)
+  }
+
+  isHidden(element) {
+    return element.classList.contains('hidden') || element.classList.contains('search-hidden')
```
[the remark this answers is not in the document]
justinlittman: @alundgard @dbranchini What say you?

dbranchini:
I'm happy with it.

app/javascript/controllers/tree_controller.js
```diff
@@ -0,0 +1,234 @@
...
+  
+    }
+  }
+
+  isBranch(element) {
+    return element.dataset.treeRole === 'branch'
+  }
+
+  isClosedBranch(element) {
+    return element.dataset.treeRole === 'branch' && element.getAttribute('aria-expanded') !== 'true'
```
jcoyne, line 160:
```suggestion
    return this.isBranch(element) && element.getAttribute('aria-expanded') !== 'true'
```

app/javascript/controllers/tree_controller.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  isOpenBranch(element) {
+    return element.dataset.treeRole === 'branch' && element.getAttribute('aria-expanded') === 'true'
```
jcoyne, line 164:
```suggestion
    return this.isBranch(element) && element.getAttribute('aria-expanded') === 'true'
```

app/javascript/file.js
```diff
@@ -0,0 +1,15 @@
+import EmbedThis from 'src/modules/embed_this';
+import PopupPanels from 'src/modules/popup_panels';
+import { Application } from "@hotwired/stimulus"
+import TreeController from "controllers/tree_controller"
+import DownloadAllController from "controllers/download_all_controller"
+
+window.Stimulus = Application.start()
+Stimulus.register("tree", TreeController)
```
jcoyne, line 8: Currently these controllers are going to be in the media bundle as well because it loads everything in `controllers` (see `controllers/index.js`).  Do we want to adjust that so these new controllers are not required for media?

jcoyne (changes requested):

app/javascript/controllers/tree_controller.js
justinlittman: `tr:visible:nth-child(odd)` and `[visible]:nth-child(odd)` don't work.

`tr:nth-child(odd)` stripes, but doesn't take into account visibility.

app/javascript/file.js
justinlittman: Adjusted.

app/javascript/controllers/tree_controller.js
jcoyne: sorry, I meant to say hidden rather than visible

justinlittman: @jcoyne None of these worked:
```
[visible]:nth-child(odd)  {
  background-color: $color-beige-10;
}

tr[visible]:nth-child(odd)  {
  background-color: $color-beige-10;
}

tr:visible:nth-child(odd)  {
  background-color: $color-beige-10;
}
```

jcoyne: @justinlittman yes, you used "visibile", but we want `not(.hidden)`

app/javascript/file_controllers/tree_controller.js
```diff
@@ -0,0 +1,234 @@
...
+
+  navigate(event) {
+    switch(true) {
+      case event.key === 'Enter' && this.isBranch(event.target):
+        event.preventDefault()
+        this.toggle(event)
+        break;
+      case event.key === 'Enter' && this.isLeaf(event.target):      
+        const downloadElement = this.downloadElement(event.target)
+        if(!downloadElement) break;
```
jcoyne, line 104: Can you add a comment here about what this is for.

app/javascript/file_controllers/tree_controller.js
```diff
@@ -0,0 +1,234 @@
...
+      case event.key === 'ArrowRight' && this.isOpenBranch(event.target):
+        event.preventDefault()
+        event.target.nextElementSibling.focus()
+        break;
+      case event.key === 'ArrowLeft' && this.isOpenBranch(event.target):
+        event.preventDefault()
+        this.toggle(event)
+        break;
+      case event.key === 'ArrowLeft' && event.target.getAttribute('aria-level') > 1:
+        const parentBranchRowElement = this.parentBranchRowElement(event.target)
```
jcoyne, line 122: Can you please add a comment here about what this is doing?

app/javascript/file_controllers/tree_controller.js
```diff
@@ -0,0 +1,234 @@
...
+        if(!parentBranchRowElement) break;
+
+        event.preventDefault()
+        parentBranchRowElement.focus()
+        break;
+      case event.key === 'ArrowUp':
+        const previousRowElement = this.previousBranchRowElement(event.target)
```
jcoyne, line 129: And here

app/javascript/file_controllers/tree_controller.js
```diff
@@ -0,0 +1,234 @@
...
+        if(!previousRowElement) break;
+        
+        event.preventDefault()
+        previousRowElement.focus()
+        break;
+      case event.key === 'ArrowDown':
+        const nextRowElement = this.nextBranchRowElement(event.target)
```
jcoyne, line 136: And here?

app/javascript/file_controllers/tree_controller.js
```diff
@@ -0,0 +1,234 @@
...
+        if(!nextRowElement) break;
+        
+        event.preventDefault()
+        nextRowElement.focus()
+        break;
+      case event.key === 'Home':
+        event.preventDefault()
+        this.tableTarget.querySelector('tbody > tr').focus()
```
jcoyne, line 144: Can you make this tr a stimulus target?

app/javascript/controllers/tree_controller.js
justinlittman: I've tried `not(.hidden)` as well.

app/javascript/file_controllers/tree_controller.js
```diff
@@ -0,0 +1,234 @@
...
+        break;
+      case event.key === 'End':
+        event.preventDefault()
+        const rowElements =
```
jcoyne, line 148: Why do we need this assignment?

jcoyne (changes requested):

app/javascript/file_controllers/tree_controller.js
justinlittman: Comments added here and below.

app/javascript/file_controllers/tree_controller.js
justinlittman: I switched this to use `this.visibleRows` instead.

jcoyne (approved):

PR #1866, opened 2023-11-29
Convert EmbedThis to a module

jcoyne:


app/javascript/src/modules/embed_this.js
```diff
@@ -1,42 +1,39 @@
...
-              urlAttr   = '&' + $(this).data('embed-attr') + '=true';
-          if(inputType === 'checkbox'){
-            if(checked) {
-              textarea.text(textarea.text().replace(urlAttr, ''));
-            }else{
-              textarea.text(textarea.text().replace(src, src + urlAttr));
-            }
+const EmbedThis = (function() {
+  let formContainer;
+  return {
```
jcoyne, line 5: This is mostly an indent change. It may be easier to review if you select "Hide whitespace"

mjgiarlo (approved):

PR #1870, opened 2023-11-29
Switch to model-viewer

### LINKED ISSUE #1486: virtex3d is unmaintained
We should switch to something that is maintained. 

After some prototyping and experimentation (see below) and consultation with @calavano and @andrewjbtw we can move forward using [model-viewer](https://modelviewer.dev/). We are not going to modify assemblyWF at this time, and will migrate the existing OBJ files to GLB (GLTF), and update them.

Maybe these should be separate tickets but it seems easier just to list the sub-tasks here:

- [x] Download existing 3D items that accessioned from preservation-catalog
- [x] Convert the OBJ files to GLB (GLTF)
- [ ] Accession the new files
- [x] Update sul-embed to use model-viewer instead of virtex3d
- [ ] Is keyboard-only navigation is possible for zoom in/out?
- [ ] Does model-viewer support full descriptions for stage views in place of "viewing stage left", etc? If it does, could additional metadata be supplied during accessioning?

edsu:
Since virtex3d is no longer supported we are switching to a 3D object viewer that is: Google's [model-viewer](https://modelviewer.dev/). Not much really changes, but because "virtex3d" was cooked into lots of filenames and classes there appear to be a lot more changes than there are.

This PR needs to be held until new GLTF files that were created from existing OBJ files are accessioned into SDR. The files are ready to go, I just need to put together a PreAssembly job to update them in prod.

I have tested in stage by deploying this branch to *embed-stage* and looking at [druid:br443tk6704](https://argo-stage.stanford.edu/view/druid:br443tk6704) which has had the GLTF file accessioned.

Fixes #1486


app/assets/stylesheets/virtex_3d.scss
```diff
@@ -24,84 +21,9 @@ $loader-black-bg: 'data:image/gif;base64,R0lGODlhFAAUAMQSAGxsbFpaWnh4eEZGRhwcHD8
...
-      cursor: grabbing !important;
-      cursor: -moz-grabbing !important;
-      cursor: -webkit-grabbing !important;
-    }
-    // scss-lint:enable ImportantRule
-
-    .viewport {
-      position: relative;
-    }
-
```
edsu, line 42: The loading styles here were hooking into something virtex3d specific.

app/assets/stylesheets/virtex_3d.scss
```diff
@@ -24,84 +21,9 @@ $loader-black-bg: 'data:image/gif;base64,R0lGODlhFAAUAMQSAGxsbFpaWnh4eEZGRhwcHD8
...
-      }
-    }
-    // scss-lint:enable SelectorDepth SelectorFormat
+  model-viewer {
+    background-color: black;
+    width: 100vw;
+    height: calc(100vh - 70px);
   }
 }
-
```
edsu, line 84: These appeared to be unused.

app/components/embed/virtex3d_component.html.erb
```diff
@@ -5,11 +5,8 @@
       <button aria-label="Zoom in" class="zoom-in">+</button>
       <button aria-label="Zoom out" class="zoom-out">-</button>
     </div>
-    <div
-      id="virtex-3d-viewer"
-      class="virtex"
-      data-three-dimensional-file="<%= viewer.three_dimensional_files.first %>"
-    >
+    <div id="3d-viewer">
+      <model-viewer auto-rotate camera-controls ar-modes="webxr scene-viewer quick-look" src="<%= viewer.three_dimensional_files.first %>"></model-viewer>
```
edsu, line 9: The is the new model-viewer component!

app/javascript/src/modules/virtex_3d_viewer.js
```diff
@@ -1,47 +1,17 @@
 'use strict';
```
edsu, line 2: Lots of the code below can go away since the model-viewer component declaratively sets up the viewer. It's hard to see from the diff but I also used JavaScript directly to wire up the zooming buttons, rather than relying on jQuery.

config/application.rb
```diff
@@ -29,7 +30,11 @@ class Application < Rails::Application
     config.autoload_lib(ignore: %w(assets tasks))
 
     config.view_component.default_preview_layout = "component_preview"
-    
+
+    # allow model-viewer web component to be used in views (and not stripped)
+    config.action_view.sanitized_allowed_tags = Rails::Html::SafeListSanitizer.allowed_tags.add('model-viewer')
```
edsu, line 35: I needed to add `model-viewer` and its attributes to the list of things that Rails will allow in templates.

yarn.lock
```diff
@@ -1035,18 +1035,18 @@
   resolved "https://registry.yarnpkg.com/@edsilv/http-status-codes/-/http-status-codes-1.0.3.tgz#a1b767c552ac43f2983ab2b9ee20e9c73b79960b"
   integrity sha512-HLK2FS5sZqxPqD53D6hhZxC6C8THTVwlyZDZ7J0iWsrB8JmMA69m/CQuNKZc1kki9WSVeck2fXna26NL0SE7cg==
 
-"@edsilv/key-codes@0.0.9":
-  version "0.0.9"
-  resolved "https://registry.yarnpkg.com/@edsilv/key-codes/-/key-codes-0.0.9.tgz#70b411e8812887c14954a2b809cb05896e0cc853"
-  integrity sha512-phdZjRKoa6EqAAfnraWZrwfAroy9cY68EM+WRUU6axq1xS8u3KMrLhyR6Hw/N+mL7ZZNveyfNGr5qg2O0LO3EQ==
-  dependencies:
-    "@types/node" "^7.0.5"
-
```
edsu, line 1044: 👋

app/javascript/packs/virtex_3d_viewer.js
```diff
@@ -6,6 +6,8 @@ import { PopupPanels } from '../src/modules/popup_panels.js';
 import Fullscreen from '../src/modules/fullscreen.js';
 import Virtex3DViewer from '../src/modules/virtex_3d_viewer.js';
```
jcoyne, line 7: Can this be removed?

config/application.rb
jcoyne: We aren't using the sanitize helper, so I'm not sure this is doing anything.   Rails does not sanitize templates. You can put anything in them, even if it's not valid.

edsu: At least in development the `<model-viewer>` component disappeared entirely from the delivered HTML until I added that. I can double check that's the case though.

edsu: Also, FWIW, Rails uses the sanitize helper.

edsu: Actually, you're right @jcoyne. I don't know what I was doing wrong, but after removing this config it still works. Thanks for catching this!

app/javascript/src/modules/model_viewer.js
```diff
@@ -0,0 +1,17 @@
...
+      e.preventDefault();
+      modelViewer.zoom(1);
+    });
+
+    document.querySelector('.zoom-out').addEventListener('click', function(e) {
+      e.preventDefault();
+      modelViewer.zoom(-1);
+    });
+  }
+}
```
edsu, line 17: The original JavaScript has been simplified and jqueryisms have been removed.

spec/lib/embed/viewer/model_viewer_spec.rb
```diff
@@ -20,7 +20,7 @@
   describe '#three_dimensional_files' do
     it 'returns the full file URL for the PDFs in an object' do
       expect(viewer.three_dimensional_files.length).to eq 1
-      expect(viewer.three_dimensional_files.first).to match(%r{/file/druid:abc123/obj1\.obj$})
+      expect(viewer.three_dimensional_files.first).to match(%r{/file/druid:abc123/obj1\.glb$})
```
edsu, line 23: The viewer needs binary GLTF (glb) files now.

jcoyne:
@edsu can you specify what the "[HOLD]" is for?

jcoyne (approved):
👏

edsu:
> @edsu can you specify what the "[HOLD]" is for?

Sure, I added a note to the description.

jcoyne (approved):

PR #1875, opened 2023-11-29
[DRAFT] Optionally set the content of the a11y modal if transcript files are present

aaron-collier:
Needs testing in stage.


app/models/embed/purl/resource.rb
```diff
@@ -51,6 +51,11 @@ def thumbnail
       def caption_files
         files.select(&:caption?).sort_by(&:language_code)
       end
+
+      # @return [Array<ResourceFile>]
+      def transcript_files
```
jcoyne, line 56: Is this method used anywhere?

app/components/embed/media_with_companion_windows_component.html.erb
```diff
@@ -161,11 +161,13 @@
 
   <dialog class="modal-container" data-media-target="accessibility" data-action="click->media#handleBackdropClicks" id="accessibility">
     <header>Accessibility</header>
-    <p>To request a transcript or other accommodation for this media item,
-      please send a request along with a link to this item to
-      <a href="mailto:purl-feedback@lists.stanford.edu">purl-feedback@lists.stanford.edu</a>.</p>
-
-    <p>A transcript may be available for download. Click the share & download icon to see if a download is available.</p>
+    <% if transcript_files.present? %>
```
jcoyne, line 164: Do we need to make sure the transcript is downloadable by the user?

app/models/embed/purl/resource.rb
aaron-collier: @jcoyne: Nope - that was part of my first and will be removed before I undraft.

app/models/embed/purl.rb
```diff
@@ -52,6 +52,10 @@ def downloadable_files
       all_resource_files.select(&:downloadable?)
     end
 
+    def transcript_files
```
jcoyne, line 55: How about renaming to `downloadable_transcript_files?` and then use `any?` instead of `select`

jcoyne (changes requested):

app/components/embed/media_with_companion_windows_component.html.erb
jcoyne: I mean to say, the transcript could be "stanford only" but the user may not have that access.  Might need design for that.

aaron-collier: It has in `models/embed/purl.rb` - which I believe is confirmed in the test, but if you think more is necessary, I'm open to suggestion.

aaron-collier: I would need some pairing to think through that.

jcoyne (approved):

Issue #1761, opened 2023-11-16
image indicating audio should display prior to user login


dinahhandel:
### Expected behavior
There is an image used that indicates something is an audio media object. This image should display prior to the user logging in.

### Actual behavior
The current screen is blank/white and shows a scrub bar at the bottom.
![Screenshot 2023-11-16 at 3 57 39 PM](https://github.com/sul-dlss/sul-embed/assets/9371978/f642dffd-385c-41cf-96ec-02536e9368b4)

### Steps to reproduce the behavior
If already authenticated in, open in incognito window: https://sul-purl-stage.stanford.edu/yh788wh7909

mjgiarlo:
This is specifically to display the default poster image for audio for audio-type objects that require user login. Because VideoJS handles posters currently, we recommend drawing a `<div>` or the like atop the audio element and hiding it after successful authentication.

mjgiarlo:
We could try converting the default audio poster from an SVG to a PNG. That may be the easiest thing to do.

mjgiarlo:
video/audio, restricted/world, has poster/lacks poster, pre-login/post-login, examples

<table>
<th>
 <td>Type
 <td>Access
 <td>Poster
 <td>Example URL
 <td>Pre-login display
 <td>Post-login display
<tr>
 <td>Video
 <td>Open
 <td>Has Poster
 <td>Object's poster
 <td>Object's poster
 <td>
<tr>
 <td>Video
 <td>Open
 <td>Lacks Poster
 <td>Default waveform poster
 <td>Default waveform poster
 <td>
<tr>
 <td>Video
 <td>Restricted
 <td>Has Poster
 <td>Locked poster
 <td>Object's poster
 <td>
<tr>
 <td>Video
 <td>Restricted
 <td>Lacks Poster
 <td>Locked poster
 <td>Default waveform poster
 <td>
<tr>
 <td>Audio
 <td>Open
 <td>Has Poster
 <td>Object's poster
 <td>Object's poster
 <td>
<tr>
 <td>Audio
 <td>Open
 <td>Lacks Poster
 <td>Default waveform poster
 <td>Default waveform poster
 <td>
<tr>
 <td>Audio
 <td>Restricted
 <td>Has Poster
 <td>Locked poster
 <td>Object's poster
 <td>
<tr>
 <td>Audio
 <td>Restricted
 <td>Lacks Poster
 <td>Locked poster
 <td>Default waveform poster
 <td>
</table>

[...]

PR #1887, opened 2023-11-30
Convert waveform-audio-poster.svg to .png

aaron-collier:
Part of #1761 

This fixes the display of the default audio image (when logged in):

Screenshot 2023-11-30 at 3 47 07 PM

A follow up PR will work on the logged out image.


jcoyne:
I thought the bug was only for users who did not have access. What bug is occurring when they are logged in?

aaron-collier:
No image is displayed at all when the audio is played, at least as far as I could tell.

jcoyne:
Screenshot 2023-11-30 at 8 18 11 PM
It (the svg) appears on stage for me when the audio is playing.

PR #1915, opened 2023-12-05
Allow clipboard-read in iframe

### LINKED ISSUE #1913: Copy icons in share modal don't work in chrome
### Expected behavior
When a user clicks the copy icon in the share modal, the text should be copied to the clipboard.

### Actual behavior
Using Chrome, when I click the copy icon, no text is copied to the clipboard. 

https://github.com/sul-dlss/sul-embed/assets/9371978/df661929-f204-4fbb-a98b-629e9ceb8546

Using Chrome, open the share & download modal and click the icon to copy the PURL link or the embed text. 

### Browser / Environment
Chrome Version 119.0.6045.159 (Official Build) (x86_64)

aaron-collier:
Fixes #1913 by allowing reading of the clipboard.

### DIFF
--- a/app/components/iframe_component.rb
+++ b/app/components/iframe_component.rb
@@ -25,6 +25,7 @@ class IframeComponent < ViewComponent::Base
 
   def call
     tag.iframe(src:, height: height_style, width: width_style, title: iframe_title,
-               frameborder: 0, marginwidth: 0, marginheight: 0, scrolling: 'no', allowfullscreen: true)
+               frameborder: 0, marginwidth: 0, marginheight: 0, scrolling: 'no', allowfullscreen: true,
+               allow: 'clipboard-read')
   end
 end


app/components/iframe_component.rb
```diff
@@ -25,6 +25,7 @@ def src
 
   def call
     tag.iframe(src:, height: height_style, width: width_style, title: iframe_title,
-               frameborder: 0, marginwidth: 0, marginheight: 0, scrolling: 'no', allowfullscreen: true)
+               frameborder: 0, marginwidth: 0, marginheight: 0, scrolling: 'no', allowfullscreen: true,
+               allow: 'clipboard-read')
```
jcoyne, line 29: don't we want to write to the clipboard?

aaron-collier: Yes, `clipboard-write` works as well, which I believe you are correct that it is preferred.

jcoyne (approved):

PR #1925, opened 2023-12-06
Adds conditional caching for embed get and iframe endpoints.

justinlittman:
Does NOT add for iiif endpoint.

### DIFF
--- a/spec/models/embed/purl_json_loader_spec.rb
+++ b/spec/models/embed/purl_json_loader_spec.rb
@@ -5,100 +5,132 @@ require 'rails_helper'
 RSpec.describe Embed::PurlJsonLoader do
   include PurlFixtures
 
-  subject(:data) { described_class.load('12345') }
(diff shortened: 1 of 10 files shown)


spec/models/embed/purl_json_loader_spec.rb
```diff
@@ -5,34 +5,56 @@
 RSpec.describe Embed::PurlJsonLoader do
   include PurlFixtures
 
-  subject(:data) { described_class.load('12345') }
+  subject(:loader) { described_class.new('12345') }
+
+  let(:data) { loader.load }
```
jcoyne, line 10: The result of load is the subject of these tests. So can you please keep this the `subject`?  If you want to test loader.etag, make a different `description` block.  and set the subject to `loader.etag`

justinlittman: Fixed.

PR #1937, opened 2023-12-07
Refactor locked-media-poster to set hidden

aaron-collier:
... and use `aria-hidden`

### DIFF
--- a/app/components/embed/media_with_companion_windows_component.html.erb
+++ b/app/components/embed/media_with_companion_windows_component.html.erb
@@ -105,7 +105,7 @@
       </div>
     </aside>
     <div class='sul-embed-body sul-embed-media' id='sul-embed-body'>
-      <div data-controller="locked-media-poster" data-action="auth-success@window->locked-media-poster#hide auth-denied@window->locked-media-poster#show" tabindex="-1" role="presentation" style="display: none; flex: 1 0 100%;">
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/locked_media_poster_controller.js
```diff
@@ -3,16 +3,19 @@ import { Controller } from "@hotwired/stimulus"
...
   }
 
   hide() {
-    this.element.style.display = 'none'
+    this.element.hidden = true
+    this.element.setAttribute('aria-hidden', 'true');
   }
 
   hasPoster() {
-    const tags = document.querySelectorAll('video,audio')
```
aaron-collier, line 15: Removing the `audio` selector here as a specific audio poster component is added in #1912

app/components/embed/media_with_companion_windows_component.html.erb
```diff
@@ -105,7 +105,7 @@
       </div>
     </aside>
     <div class='sul-embed-body sul-embed-media' id='sul-embed-body'>
-      <div data-controller="locked-media-poster" data-action="auth-success@window->locked-media-poster#hide auth-denied@window->locked-media-poster#show" tabindex="-1" role="presentation" style="display: none; flex: 1 0 100%;">
+      <div data-controller="locked-media-poster" data-action="auth-success@window->locked-media-poster#hide auth-denied@window->locked-media-poster#show" tabindex="-1" role="presentation" style="flex: 1 0 100%;" aria-hidden="true" hidden>
```
jcoyne, line 108: do we need aria-hidden if we have the hidden attribute? I don't believe we do.

jcoyne (changes requested):

jcoyne (approved):

Issue #1919, opened 2023-12-05
PDF Viewer Auth Flow: Implement sul-embed UI


peetucket:
Part of #1757 - auth flow for PDF viewer

Add files controllers and tags, etc. to UI

See bullet points 3 and 6 here: https://docs.google.com/document/d/17nhSWDp28CocKsBdcZvtqAOMiNbLnVspBRXENSr6m84/edit#heading=h.fknftcv8cgfk

https://docs.google.com/document/d/1IIPYw4CVIaNCjPjtY427JxcPQngd5ZxgqhQLfhxVbi4/edit#heading=h.fuuyr6c6x300

https://iiif.io/api/auth/2.0/#71-authorization-flow-algorithm

Note that auth service should also indicate if a user is denied due to location restrictions, and this same logic can be used to show messaging around location restrictions, exactly as is done for media.

- [x] Use auth service from #1067 (IIIF probe service)
~~Adjust the viewers (each the PDF viewer) to add target tags to each file that needs to have auth checked.
We need to have the equivalent of the media_tag_controller for files, extracting just the parts that iterate through the existing target tagged files and running auth checks to the new stacks endpoint for each.  It’s possible that most of the media_tag_controller can be generalized, with only the media specific methods (e.g. having to do with media player and thumbnails) left in the media_tag_controller.~~
~~Generalize the auth_restriction_controller … could be as simple as moving out of the media namespace and updating references.~~
~~Add hidden divs with the relevant to the other sul-embed viewers (e.g. pulled from https://github.com/sul-dlss/sul-embed/blob/main/app/components/embed/media_with_companion_windows_component.html.erb#L17-L50).  This could probably be generalized into a new view component to avoid duplication.  Update references to new javascript controllers that are generalized.~~

dbranchini:
Related #1939 .

lwrubel:
See https://github.com/sul-dlss/sul-embed/issues/1757 for additional info.

PR #1971, opened 2023-12-12
Show better messaging for PDFs that are no download/citation only/embargoed

### LINKED ISSUE #1939: Improve UX for Stanford no-download PDFs (i.e., provide some kind of message saying the document isn't viewable)
These are the two sample Purls I'm using in stage, but this issue is mainly about the Stanford no-download one:

- [Stanford-only](https://sul-purl-stage.stanford.edu/xj550hm9593) 
  - we would like this to work with some kind of auth flow, which I think is ticketed elsewhere
- [Stanford-only but no download](https://sul-purl-stage.stanford.edu/gd142xn9905) 
  - we know this won't work and would like that to be more up front about that, rather then sending people through auth only to tell them they can't get the file anyway
  - (as an aside) in general, we should have an indication that a "no-download" PDF won't work

**Suboptimal scenario 1:**

When viewing the Stanford no-download purl in a firefox private window earlier today, I got prompted to auth within the embed viewer itself. I took this screenshot after going through Duo but before clicking through to the next "page":

> Screenshot 2023-12-07 at 8 40 44 AM

I was then shown a Firefox error message:

> Screenshot 2023-12-07 at 8 41 00 AM

Finally, I clicked on the blue "open site in new window" and got a page that said "Access forbidden."

**Suboptimal scenario 2**

When already logged in as a Stanford user, I get a message telling me I should download the file:

> Screenshot 2023-12-07 at 8 44 13 AM

The problem of course is that the file is no-download, so that will never work.

**What is to be done?**

Ideally, people would not accession PDFs with no-download rights (world or Stanford) and we wouldn't encounter this issue. But we do have some content with such rights and we should show a message that at least doesn't imply the content is downloadable, only to deny the access when they run out of links to click on.

I believe there's a design for an access restriction banner that says "There was an issue viewing this item" and directs people to look at the Access conditions to see if there's any more information. There's a chance that there won't be helpful information in the Access conditions, but that's where there could be information about restrictions. We could alternatively ask people to contact purl-feedback for help. If we don't go with a banner we could at least put a message in the viewer box.

**What happens currently**

In prod, we show an error in this situation. It's not ideal but better than going down an authentication path that ends in disappointment.

https://docs.google.com/document/d/1IIPYw4CVIaNCjPjtY427JxcPQngd5ZxgqhQLfhxVbi4/edit#heading=h.fuuyr6c6x300

peetucket:
Fixes #1939 - improve the UX experience for no-download PDFs.

Screen Shot 2024-01-05 at 11 31 40 AM

Notes:
- This PR adds embargo and no citation checking and messaging (not specifically identified in the ticket, but identified in the messaging to be shown)
- The "no download" for PDFs is applicable to anything (not just stanford only no download).  If the PDF has "no download" set, it won't work for anyone.
- Stanford only downloadable content (i.e. where users could see it if authed) has it's own workflow and will be dealt with separately in #2032 
- Location restricted content already shows a message on the PURL page, eg. https://purl.stanford.edu/bd742gh0511

Screen Shot 2024-01-05 at 11 25 15 AM

Todo:
- [x] put message text in locale file
- [x] verify message text with what is in the ticket
- [x] tests
- [x] add messaging into legacy PDF viewer

### DIFF
--- a/app/components/embed/legacy_pdf_component.html.erb
+++ b/app/components/embed/legacy_pdf_component.html.erb
@@ -2,28 +2,48 @@
   <%= render Embed::HeaderComponent.new viewer: viewer %>
   <% # removing 5 pixels from the body height to account for being embedded %>
   <div class='sul-embed-body sul-embed-pdf'>
-    <button aria-label="Zoom in" class="button zoom-in">+</button>
-    <button aria-label="Zoom out" class="button zoom-out">-</button>
-    <button aria-label="Previous page" class="button prev-page">
-      <i class="sul-i-arrow-left-8"></i>
-    </button>
-    <button aria-label="Next page" class="button next-page">
-      <i class="sul-i-arrow-right-8"></i>
-    </button>
-    <div id="pdf-viewer" class="pdf-viewer"
-      data-pdf-url="<%= viewer.pdf_files.first %>"
-      data-location-restricted="<%= viewer.all_documents_location_restricted? %>"
-    >
-      <div class="loading-spinner"></div>
-      <canvas></canvas>
-      <button
-        aria-label="Exit full screen"
-        class="btn sul-i-close sul-close-fullscreen-button"
-        id="close-full-screen-button">
-      </button>
-    </div>
-    <% # NOTE: Remove legacy_pdf_viewer.js file when this component is removed %>
-    <%= javascript_pack_tag('legacy_pdf_viewer') %>
+    <% if viewer.available? %>
+        <button aria-label="Zoom in" class="button zoom-in">+</button>
+        <button aria-label="Zoom out" class="button zoom-out">-</button>
+        <button aria-label="Previous page" class="button prev-page">
+          <i class="sul-i-arrow-left-8"></i>
+        </button>
+        <button aria-label="Next page" class="button next-page">
+          <i class="sul-i-arrow-right-8"></i>
+        </button>
+        <div id="pdf-viewer" class="pdf-viewer"
+          data-pdf-url="<%= viewer.pdf_files.first %>"
+          data-location-restricted="<%= viewer.all_documents_location_restricted? %>"
+        >
+          <div class="loading-spinner"></div>
+          <canvas></canvas>
+          <button
+            aria-label="Exit full screen"
+            class="btn sul-i-close sul-close-fullscreen-button"
+            id="close-full-screen-button">
+          </button>
+          <button aria-label="Next page" class="button next-page">
+            <i class="sul-i-arrow-right-8"></i>
+          </button>
+          <div id="pdf-viewer" class="pdf-viewer"
+            data-pdf-url="<%= viewer.pdf_files.first %>"
+            data-location-restricted="<%= viewer.all_documents_location_restricted? %>"
+          >
+            <div class="loading-spinner"></div>
+            <canvas></canvas>
+            <button
+              aria-label="Exit full screen"
+              class="sul-embed-btn sul-i-close sul-close-fullscreen-button"
+              id="close-full-screen-button">
+            </button>
+          </div>
+          <% # NOTE: Remove legacy_pdf_viewer.js file when this component is removed %>
+          <%= javascript_pack_tag('legacy_pdf_viewer') %>
+      <% else %>
+          <div style="padding: 10%;">
+            <%= viewer.restriction_message %>
+          </div>
+      <% end %>
   </div>
   <%= render 'embed/metadata_panel', viewer: viewer %>
   <%= render 'embed/embed_this/generic', viewer: viewer %>
--- a/app/components/embed/legacy_pdf_component.html.erb
+++ b/app/components/embed/legacy_pdf_component.html.erb
@@ -41,7 +41,7 @@
           <%= javascript_pack_tag('legacy_pdf_viewer') %>
       <% else %>
           <div style="padding: 10%;">
-            <%= viewer.restriction_message %>
+            <strong><%= I18n.t('restrictions.not_accessibile') %></strong>
           </div>
       <% end %>
   </div>
--- a/app/components/embed/pdf_component.html.erb
+++ b/app/components/embed/pdf_component.html.erb
@@ -3,7 +3,7 @@
   <% # removing 5 pixels from the body height to account for being embedded %>
   <div class='sul-embed-body sul-embed-pdf'>
     <object data="<%= viewer.pdf_files.first %>" type="application/pdf" style="height: 100vh; width: 100%">
-      <p>Your browser does not support viewing PDFs.  Please <a href="<%= viewer.pdf_files.first %>">download the file</a> to view it.</p>
+      <p>The PDF could not be displayed - either your browser does not support viewing PDFs or access is restricted.  Please try <a target="_blank" href="<%= viewer.pdf_files.first %>">download the file</a> to view it.</p>
     </object>
   </div>
   <%= render 'embed/metadata_panel', viewer: viewer %>
--- a/app/components/embed/pdf_component.html.erb
+++ b/app/components/embed/pdf_component.html.erb
@@ -5,12 +5,14 @@
     <% if viewer.available? %>
       <object data="<%= viewer.pdf_files.first %>" type="application/pdf" style="height: 100vh; width: 100%">
         <div style="padding: 10%;">
-          Your browser does not support viewing PDFs.  Please <a target="_blank" href="<%= viewer.pdf_files.first %>">download the file</a> to view it.
+          <strong>
+            <%= I18n.t('restrictions.not_supported_html', download_link: link_to(I18n.t('restrictions.download_the_file'), viewer.pdf_files.first, target: '_blank')).html_safe %>
+          </strong>
         </div>
       </object>
     <% else %>
        <div style="padding: 10%;">
-          <%= viewer.restriction_message %>
+          <strong><%= I18n.t('restrictions.not_accessibile') %></strong>
        </div>
     <% end %>
   </div>
--- a/app/components/embed/pdf_component.html.erb
+++ b/app/components/embed/pdf_component.html.erb
@@ -2,9 +2,17 @@
   <%= render Embed::HeaderComponent.new viewer: viewer %>
   <% # removing 5 pixels from the body height to account for being embedded %>
   <div class='sul-embed-body sul-embed-pdf'>
-    <object data="<%= viewer.pdf_files.first %>" type="application/pdf" style="height: 100vh; width: 100%">
-      <p>The PDF could not be displayed - either your browser does not support viewing PDFs or access is restricted.  Please try <a target="_blank" href="<%= viewer.pdf_files.first %>">download the file</a> to view it.</p>
-    </object>
+    <% if viewer.available? %>
+      <object data="<%= viewer.pdf_files.first %>" type="application/pdf" style="height: 100vh; width: 100%">
+        <div style="padding: 10%;">
+          Your browser does not support viewing PDFs.  Please <a target="_blank" href="<%= viewer.pdf_files.first %>">download the file</a> to view it.
+        </div>
+      </object>
+    <% else %>
+       <div style="padding: 10%;">
+          <%= viewer.restriction_message %>
+       </div>
+    <% end %>
   </div>
   <%= render 'embed/metadata_panel', viewer: viewer %>
   <%= render 'embed/embed_this/generic', viewer: viewer %>
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -17,13 +17,13 @@ module Embed
     attr_accessor :druid, :type, :title, :use_and_reproduction, :copyright, :contents, :collections,
                   :license, :bounding_box, :embargo_release_date, :archived_site_url, :external_url,
                   :embargoed, :citation_only, :stanford_only_unrestricted, :public, :controlled_digital_lending,
-                  :stanford_only_no_download, :etag, :last_modified
+                  :no_download, :etag, :last_modified
 
     alias embargoed? embargoed
     alias stanford_only_unrestricted? stanford_only_unrestricted
     alias public? public
     alias controlled_digital_lending? controlled_digital_lending
-    alias stanford_only_no_download? stanford_only_no_download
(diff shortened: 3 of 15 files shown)


app/views/pages/sandbox.html.erb
```diff
@@ -198,6 +198,15 @@
...
+      <li>
+        pdf (Stanford Only) - bb461xx1037
+      </li>
+      <li>
+        pdf (Citation Only) - bz673hm0344
+      </li>
+      <li>
+        pdf (emgargoed) - bb990sg8412
+      </li>
       <li>
```
peetucket, line 210: some more examples for testing different rights

app/components/embed/pdf_component.html.erb
```diff
@@ -3,7 +3,7 @@
   <% # removing 5 pixels from the body height to account for being embedded %>
   <div class='sul-embed-body sul-embed-pdf'>
     <object data="<%= viewer.pdf_files.first %>" type="application/pdf" style="height: 100vh; width: 100%">
-      <p>Your browser does not support viewing PDFs.  Please <a href="<%= viewer.pdf_files.first %>">download the file</a> to view it.</p>
+      <p>Your browser does not support viewing PDFs.  Please <a target="_blank" href="<%= viewer.pdf_files.first %>">download the file</a> to view it.</p>
     </object>
```
peetucket, line 7: if the user needs to auth to view the PDF, this at least sends them to a new window

app/models/embed/purl_xml_loader.rb
```diff
@@ -52,6 +53,10 @@ def stanford_only_unrestricted
       rights.stanford_only_unrestricted?
     end
 
+    def stanford_only_no_download
+      !public && !rights.stanford_only_downloadable?
+    end
+
     def controlled_digital_lending
```
peetucket, line 60: Public XML: A new restriction type we are catching (currently only for the PDF viewer, but could be elsewhere).

app/models/embed/purl_json_loader.rb
```diff
@@ -56,6 +57,10 @@ def stanford_only_unrestricted
       json.dig('access', 'download') == 'stanford'
     end
 
+    def stanford_only_no_download
+      json.dig('access', 'download') == 'none' && json.dig('access', 'view') == 'stanford'
+    end
+
```
peetucket, line 63: Cocina JSON: A new restriction type we are catching (currently only for the PDF viewer, but could be elsewhere).

app/components/embed/pdf_component.html.erb
```diff
@@ -2,9 +2,17 @@
...
+        <div style="padding: 10%;">
+          Your browser does not support viewing PDFs.  Please <a target="_blank" href="<%= viewer.pdf_files.first %>">download the file</a> to view it.
+        </div>
+      </object>
+    <% else %>
+       <div style="padding: 10%;">
+          <%= viewer.restriction_message %>
+       </div>
+    <% end %>
   </div>
```
peetucket, line 16: Show restriction messaging in place of PDF file (if there are restrictions that prevent viewing of content).  Stanford only restrictions will be dealt with in a separate PR since users will be able to login and to do something about it, while these restrictions prevent viewing in all cases.

peetucket:
**Citation Only or No Download (regardless of Stanford only or world access)**

Screen Shot 2023-12-13 at 11 37 23 AM

**Embargoed**

Screen Shot 2023-12-13 at 11 37 13 AM

**Stanford Only (but not authed yet)** 

New player:

Screen Shot 2023-12-13 at 11 37 42 AM

Old player:

Screen Shot 2023-12-13 at 11 40 22 AM

**Stanford Only (already authed)**

New player: works as expected

Old player: still does not work

peetucket:
On stage:

World viewable: https://sul-purl-stage.stanford.edu/br379jx5150 (anyone can see, as normal)
Stanford only viewable: https://sul-purl-stage.stanford.edu/xj550hm9593 (you can view it if authed since this is new player)
Stanford only but no download: https://sul-purl-stage.stanford.edu/gd142xn9905 (restriction message shown)
Citation only: https://sul-purl-stage.stanford.edu/vn873rh2139 (restriction message shown)
Embargoed: (cannot find example on stage, but tested in localhost sul-embed against prod and screen above)

app/models/embed/purl.rb
```diff
@@ -17,13 +17,14 @@ def attributes=(hash)
...
                   :embargoed, :citation_only, :stanford_only_unrestricted, :public, :controlled_digital_lending,
-                  :etag, :last_modified
+                  :no_download, :etag, :last_modified
 
     alias embargoed? embargoed
     alias citation_only? citation_only
     alias stanford_only_unrestricted? stanford_only_unrestricted
     alias public? public
     alias controlled_digital_lending? controlled_digital_lending
+    alias no_download? no_download
```
jcoyne, line 27: Rather than adding this, can we check the file properties instead?  I find the "download" property of an object to be confusing, but the "download" property of a file is much more easily understandable.

jcoyne (changes requested):

app/models/embed/purl.rb
peetucket: Yes, it is confusing.  I guess the question is, is it possible to have an object for which at the object level it says "download=none", but for specific files to have download allowed in some way?  If not (or if this situation happens but is to be treated as a metadata error and should act as **all** files are "download=none"), then looking at the object level is still appropriate.

BUT... if it is possible for an PDF object to **allow** download at the object level but **disallow** it for the first file, then this situation warrants checking the file first.  For PDFs anyway (where we only even try to show the first file in the object), this check may be useful to do.

peetucket: So I guess I would argue it may still be useful to have this method that verifies global no download at the object level available, but we need a second check (maybe only useful for PDFs) that also looks at the first file's rights...perhaps later we for other object types, we will need additional checks.  Unless we can conclusively say that the object level no download rights really have no meaning and never supercede what is set at the file level.

peetucket: Just added some logic to verify first PDF file download rule, but only for the case being handled here (PDF viewer only).

jcoyne: Stacks is looking at the file level. If we don't use the file level here it may not match with what stacks thinks.  The "download" on a object is just a default setting for new files added to the object.  The canonical setting is the one on the file.

peetucket: OK, removed object level no download code, only look at the first PDF file in this case.

app/models/embed/purl.rb
```diff
@@ -16,11 +16,12 @@ def attributes=(hash)
 
     attr_accessor :druid, :type, :title, :use_and_reproduction, :copyright, :contents, :collections,
                   :license, :bounding_box, :embargo_release_date, :archived_site_url, :external_url,
-                  :embargoed, :stanford_only_unrestricted, :public, :controlled_digital_lending,
+                  :embargoed, :citation_only, :stanford_only_unrestricted, :public, :controlled_digital_lending,
                   :etag, :last_modified
 
     alias embargoed? embargoed
     alias stanford_only_unrestricted? stanford_only_unrestricted
+    alias citation_only? citation_only
```
jcoyne, line 24: Is this needed?

jcoyne (changes requested):

app/models/embed/purl.rb
peetucket: Well, I want to check for citation_only below (here): https://github.com/sul-dlss/sul-embed/pull/1971/files#diff-299fb888af648917064acec2d633ac5ebfa31cdbb3794359a5de888ca27f371fR49

But now I realize I don't have the corresponding citation_only methods coming from both the XML and JSON PURL loaders.  I can easily add one for the JSON parsing, but the XML parsing comes from DorRightsAuth and while they do have a citation_only method, it seems to be used only by the indexer now: https://github.com/sul-dlss/dor-rights-auth/blob/main/lib/dor/rights_auth.rb#L76-L79

pondering how best to do this

jcoyne: Why do you need to check for citation only.  Can't you just check the download access on the files?  That will also be no-download if the object is citation only.

peetucket: Ahh, ok...so just don't bother with a citation only check.

dbranchini:
@peetucket , I'm confused about your embargo example titled Interview with John Mellonkopt. It's a strange set of objects, but the PDF is accessible now with other parts of it unavailable and embargoed until 2096. https://purl.stanford.edu/bb990sg8412

andrewjbtw:
There is an embargoed PDF in stage: https://argo-stage.stanford.edu/view/druid:xy495ms7961

dbranchini:
For the Stanford-only download (world view) example above (Minutes and supporting materials... ), could the message say, "Stanford users: log in to access all available features [LOG IN - Link or Button]" instead?

peetucket:
> @peetucket , I'm confused about your embargo example titled Interview with John Mellonkopt. It's a strange set of objects, but the PDF is accessible now with other parts of it unavailable and embargoed until 2096. https://purl.stanford.edu/bb990sg8412

Well I'm confused by that too.  It looks like the object level rights says citation only because of the embargo, but the file level rights are offering access to the PDF (hence why it is showing and why it is downloadable).  My current PR looks at object level embargo and so would show a restriction message in this case, which maybe isn't right?  I guess I don't understand the relationship between object levels and file level rights when they differ like this.  And perhaps I need to just ignore the object level rights (if they are just a default value for example) and do more work to go look at the file specific rights to decide what message to show.  In the case of PDF, it would just be the first PDF file to check.

peetucket:
> For the Stanford-only download (world view) example above (Minutes and supporting materials... ), could the message say, "Stanford users: log in to access all available features [LOG IN - Link or Button]" instead?

I had intended all of the Stanford only and location restricted stuff to be done in a separate piece of work as described in the other ticket (#1919), as it is more complicated.  Since the user can do something about these restriction, it is going to be a bunch of client side code as well.

andrewjbtw:
The item in [bb990sg8412](https://purl.stanford.edu/bb990sg8412) is strange enough that it would be difficult to plan how to handle it.

Usually, an embargo will restrict access to all files within an object. But it's entirely possible to set file-level rights that differ from object-level rights. In this case, it appears that they wanted the transcript out but not the audio recording (which isn't evens shelved, so the public doesn't even know about it).

peetucket:
> The item in [bb990sg8412](https://purl.stanford.edu/bb990sg8412) is strange enough that it would be difficult to plan how to handle it.
> 
> Usually, an embargo will restrict access to all files within an object. But it's entirely possible to set file-level rights that differ from object-level rights. In this case, it appears that they wanted the transcript out but not the audio recording (which isn't evens shelved, so the public doesn't even know about it).

Good to know.  But that means this PR will need to change and look at the first PDF file's embargo status, because as is, it would block access to all contents in cases like this (which I presume even though are rare, should still be supported).

peetucket:
The object is embargoed…but a file in the object is viewable. That is really confusing and makes me think we can’t even message this easily.

Since we already check for “no download” of the first PDF and show a message in that scenario (“no access…”), that might be what we do for the PDF viewer in these cases. Basically ignore the object level embargo status in sul-embed and depend on the file level no download rule.
________________________________

dbranchini:
> > For the Stanford-only download (world view) example above (Minutes and supporting materials... ), could the message say, "Stanford users: log in to access all available features [LOG IN - Link or Button]" instead?
> 
> I had intended all of the Stanford only and location restricted stuff to be done in a separate piece of work as described in the other ticket (#1919), as it is more complicated. Since the user can do something about these restrictions, it is going to be a bunch of client side code as well.

Sure, that makes sense and you did tell me that! When I saw the example above, I thought you might have included it in this work. Thank you!

dbranchini:
> The object is embargoed…but a file in the object is viewable. That is really confusing and makes me think we can’t even message this easily. Since we already check for “no download” of the first PDF and show a message in that scenario (“no access…”), that might be what we do for the PDF viewer in these cases. Basically ignore the object level embargo status in sul-embed and depend on the file level no download rule.

Is it safe to assume that's what's happening in production since I can view and download that PDF transcript there?

peetucket:
> Is it safe to assume that's what's happening in production since I can view and download that PDF transcript there?

Yes, I think that is what is happening with https://purl.stanford.edu/bb990sg8412 .  The file itself is marked as publicaly available in the rights metadata even though the object is embargoed and set to read=none.  I verified this since I was worried there was an issue with us releasing files that shouldn't be.  So it's not a bug, it's just unexpected (to me at least) that a file within an object can have fewer restrictions that the object itself.

spec/factories/purls.rb
```diff
@@ -30,6 +34,12 @@
       contents { [build(:resource, :document)] }
     end
 
+    trait :document_no_download do
+      type { 'document' }
+      title { 'Title of the object' }
+      contents { [build(:resource, :document_no_download)] }
+    end
+
     trait :was_seed do
```
peetucket, line 43: we need to ensure the public purls created for specs are marked as public and have a world downloadable file or else the PDF viewer will restrict viewing now

peetucket:
Updated PR to account for the fact that embargoed objects could have downloadable files.  This makes the logic pretty simple now: if the object's first PDF is downloadable in some way (world, stanford only, location, etc.), the player will be loaded.  If the file is not downloadable at all (typically because the object embagoed or citation only), the restriction message will be shown instead of the player.  

Note that there is separate ongoing work to deal with what happens if the player is shown, but the PDF is still not viewable (because it is Stanford only but the user is not logged in or because there is a location restriction and they are not in the location).

app/components/embed/legacy_pdf_component.html.erb
```diff
@@ -2,28 +2,48 @@
...
+            </button>
+          </div>
+          <% # NOTE: Remove legacy_pdf_viewer.js file when this component is removed %>
+          <%= javascript_pack_tag('legacy_pdf_viewer') %>
+      <% else %>
+          <div style="padding: 10%;">
+            <strong><%= I18n.t('restrictions.not_accessibile') %></strong>
+          </div>
+      <% end %>
   </div>
```
peetucket, line 47: these are for the legacy PDF viewer component, same changes below are for the new browser based PDF viewer component

jcoyne (approved):

PR #1975, opened 2023-12-12
Protect against null cues

mjgiarlo:


app/javascript/controllers/transcript_controller.js
```diff
@@ -26,17 +26,13 @@ export default class extends Controller {
     const tracks = this.player.textTracks_?.tracks_
     if (!tracks) return []
 
-    return tracks.filter(track => track.kind === 'captions')
+    return tracks.filter(track => track.kind === 'captions' && track?.cues?.cues_)
```
mjgiarlo, line 29: Filter out caption tracks without cues *early* so we don't get `null` or `undefined` values later.

app/javascript/controllers/transcript_controller.js
```diff
@@ -26,17 +26,13 @@ export default class extends Controller {
...
   }
 
   get cuesByLanguage() {
     const cues = {}
-    if (this.captionTracks.length == 0) return cues
```
mjgiarlo, line 34: `this.captionTracks` always returns an array, and we can safely `forEach` an empty array below, so I don't see a need for this check.

app/javascript/controllers/transcript_controller.js
```diff
@@ -26,17 +26,13 @@ export default class extends Controller {
...
 
     this.captionTracks.forEach(track => {
-      if (!track.cues)
-        return
```
mjgiarlo, line 38: We already filtered these out earlier.

app/javascript/controllers/transcript_controller.js
```diff
@@ -68,8 +64,6 @@ export default class extends Controller {
   }
 
   setupTranscriptLanguageSwitching() {
-    if (this.captionTracks.length == 0) return
```
mjgiarlo, line 71: `this.captionTracks` always returns an array, and we can safely `forEach` an empty array below, so I don't see a need for this check.

jcoyne (approved):

PR #1978, opened 2023-12-13
Track file and object downloads

### LINKED ISSUE #1935: Send download events to the SDR metrics service
See https://github.com/sul-dlss/stacks/issues/1048#issuecomment-1841839703 for information on why the stacks-based, server-side approach doesn't work (at least, not without substantial effort to implement a new API or a clever workaround). tl;dr we need to send events from the browser, at least for now.

This ticket is for implementing download tracking via javascript for all the places that we currently can in viewer interfaces:

|done?|viewer|link location|
|-|-|-|
|❌|iiif|mirador uses its own download component...might emit JS event?|
|✅|geo|`Embed::Download::FileListItemComponent`|
|✅|media|`Embed::Download::FileListItemComponent`|
|✅|pdf|`Embed::Download::FileListItemComponent`|
|✅|3d|`Embed::Download::FileListItemComponent`|
|✅|files|`Embed::Components::FileRowComponent` (file),<br/>`views/embed/footer/_file.html.erb` (object)|

thatbudakguy:
Closes #1935


app/components/embed/download/file_list_item_component.html.erb
```diff
@@ -1,5 +1,5 @@
 <li>
-  <a href='<%= url %>' target='_blank' rel='noopener noreferrer'>
+  <a href='<%= url %>' target='_blank' rel='noopener noreferrer' download>
```
thatbudakguy, line 2: this attribute is used so that we know which links to track downloads on

app/components/embed/file_row_component.html.erb
```diff
@@ -16,8 +16,10 @@
   <td role="gridcell"><%= file_size_text%></td>
   <td role="gridcell">    
     <% if download? %>
-      <a href="<%= file_url %>" target='_blank' rel='noopener noreferrer' download=''>
-        <i class='sul-i-download-3'></i> Download
+      <a href="<%= file_url %>" target='_blank' rel='noopener noreferrer' download>
+        <i class='sul-i-download-3'></i>
+        <span>Download</span>
+        <span class="sul-embed-sr-only"><%= title %></span>
```
thatbudakguy, line 22: without this, the accessible text for all the links was just "download", but they go to different places.

spec/features/metrics_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+RSpec.describe 'metrics tracking', :js do
+  before do
+    allow(Settings).to receive(:metrics_api_url).and_return('/test_metrics_api')
+    allow(Embed::Purl).to receive(:find).and_return(purl)
+    StubMetricsApi.reset!
+  end
+
+  after do
+    FileUtils.rm_rf('tmp/test_downloads')
+  end
```
thatbudakguy, line 14: without this, running the tests will occasionally download an actual file, because it happens extremely fast. the new capybara config ensures it goes into `tmp/` and then we delete it afterwards

spec/factories/files.rb
```diff
@@ -52,6 +52,13 @@
       role { 'transcription' }
     end
 
+    trait :model do
```
ndushay, line 55: to consider:  make this model_3d as well?   Model isn't enough for me to recognize it as 3d.

spec/rails_helper.rb
```diff
@@ -13,6 +13,15 @@
 require 'view_component/test_helpers'
 require 'view_component/system_test_helpers'
 
+Capybara.register_driver :selenium_chrome_headless do |app|
+  options = Selenium::WebDriver::Chrome::Options.new
+
+  options.add_argument('--headless')
+  options.add_preference(:download, default_directory: 'tmp/test_downloads')
```
ndushay, line 20: not sure it's worth it, but the directory name could be something in settings, or a constant avail to any test

ndushay (approved):
approving with a couple of quibbles for your consideration

spec/factories/files.rb
thatbudakguy: makes sense. updated

spec/rails_helper.rb
thatbudakguy: made it a constant, good call

PR #1995, opened 2023-12-13
Switch to using hidden attribute.

### LINKED ISSUE #1928: File viewer: "hidden" class vs hidden attribute
image

justinlittman:
closes #1928

### DIFF
--- a/app/assets/stylesheets/file.scss
+++ b/app/assets/stylesheets/file.scss
@@ -75,7 +75,7 @@ body {
     padding-left: 15px;
   }
 
-  tbody > tr:nth-child(odd of :not(.hidden):not(.search-hidden)) {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/file.scss
```diff
@@ -101,9 +101,8 @@ body {
     vertical-align: top;
   }
 
-  /* Hide collapsed rows */
-  tr.hidden, tr.search-hidden {
-    display: none;
+  tr[hidden] {
```
jcoyne, line 104: I don't think this is needed because you never set the display style to anything else, right?

justinlittman: It doesn't work without it.

jcoyne (approved):

PR #2005, opened 2023-12-13
Handle case where the auth-service doesn't give access

jcoyne:
And the object isn't embargoed, stanford or location access.

This should also prevent double-headers.

### DIFF
--- a/app/components/embed/media_with_companion_windows_component.html.erb
+++ b/app/components/embed/media_with_companion_windows_component.html.erb
@@ -47,13 +47,13 @@
         <button data-auth-restriction-target="embargoLoginButton" data-action="media-tag#logIn">Log in</button>
       </div>
     </div>
-  </div>
-  <% if citation_only? %>
-    <div class="authLinkWrapper">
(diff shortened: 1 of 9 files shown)


app/javascript/controllers/media_tag_controller.js
```diff
@@ -45,8 +45,8 @@ export default class extends Controller {
         window.dispatchEvent(new CustomEvent('auth-stanford-restricted'))
       window.dispatchEvent(new CustomEvent('auth-success'))
     } else {
-      const event = new CustomEvent('auth-denied', { detail: result.authResponse })
-      window.dispatchEvent(event)
+        const event = new CustomEvent('auth-denied', { detail: result.authResponse })
+        window.dispatchEvent(event)
```
mjgiarlo, line 49: Revert extra spaces?

mjgiarlo (approved):

app/javascript/controllers/media_tag_controller.js
jcoyne: Those aren't extra.  Doesn't JS world use 4 spaces?

mjgiarlo: The rest of the file, and other JS files above, looks like 2 spaces. Smells like a JS file in a Ruby shop. :)

jcoyne: You're not Rubocop! You can't tell me what to do! 😆

PR #2009, opened 2023-12-13
Nested css is not fully supported

jcoyne:
Some of the older < 120 versions of chrome and < 17.2 of safari don't have full support yet.


hudajkhan (approved):
Tried out locally and looks better in Safari

PR #2010, opened 2023-12-13
Apply hover color to media content items

### LINKED ISSUE #1969: Add selection indicator for media content items
Currently, there is no visual indicator when selecting an item within the media contents.
Toggle style `box-shadow: 0px 0px 3px 2px #016895` on the selected item.
Note: `#016895` is [SU sky dark](https://identity.stanford.edu/design-elements/color/accent-colors/) and the gradient satisfies the [WCAG color contrast ratio](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html#change-of-contrast-gradients).

mjgiarlo:
Fixes #1969

This commit adds a visual style when the user hovers over media content items, letting them know which is currently being acted upon.

Screenshot from 2023-12-13 14-12-13

### DIFF
--- a/app/assets/stylesheets/media.scss
+++ b/app/assets/stylesheets/media.scss
@@ -85,7 +85,8 @@ body {
   --header-size: 62.65px; /* depends on the line-height of the header > nav > h2 */
   --modal-border-radius: 8px;
   --modal-max-width: 600px;
-  
+  --media-thumb-hover-color: #016895;
+
   border: 1px solid var(--border-color);
   display: flex;
   flex-direction: column;
@@ -339,7 +340,7 @@ body {
       display: flex;
       flex: 1 1 auto;
     }
-  
+
     .sul-embed-media-wrapper {
       display: inline-block;
       height: 100%;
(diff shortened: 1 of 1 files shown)


aaron-collier (approved):

hudajkhan (approved):
Works in Chrome

jcoyne (commented):
wasn't the point of the ticket for the style to be on the active item?

mjgiarlo:
@jcoyne Could be? Depends on what is meant by "selected."

@alundgard should the shadow box be on the item displayed in the main part of the viewer, or the item the user is hovering over in the companion window? I interpreted "selected" as the latter. If we really want the former, I can make that change, and I'd also be curious if we want any hover behaviors (regardless of what's currently displayed).

alundgard:
The former: the shadow box should be applied to the item currently displayed in the viewer. (My bad for not explaining this clearly in #1969.) The idea is to have an indicator that disambiguates between multiple visually-redundant items, as in [this example](https://sul-purl-uat.stanford.edu/bb142ws0723). I don't think we need any styling on hover; the `cursor: pointer` is enough of an indicator that the media item is clickable.

mjgiarlo:
@alundgard OK, I'll pick this back up, no problem. Thanks!

alundgard:
Thanks @mjgiarlo ! While you're in that vicinity, I'm also seeing some new visual cruft on the media items. Unsure if this is related or not. If not, I can make a separate ticket.
media-item-cruft

mjgiarlo:
@alundgard AFAICT, it's not related, but I could be mistaken. A new ticket would be grand.

PR #2024, opened 2023-12-14
Extract tab controller

jcoyne:
This allows us to add more tabs without having to modify the javascript.  It also decouples it from the other responsibilities of the companion_window_controller.

### DIFF
--- a/app/components/embed/companion_windows_component.html.erb
+++ b/app/components/embed/companion_windows_component.html.erb
@@ -56,27 +56,27 @@
   <div class="companion-window-component-body">
     <aside class="left-drawer collapse" data-companion-window-target="leftDrawer" id="left-drawer" aria-label="Sidebar" style="visibility: hidden">
       <nav class="vert-tabs" role="tablist">
-        <button data-action="click->companion-window#displayMetadata" data-companion-window-target="leftButton" class="active" aria-label="About this item" aria-controls="about" aria-selected="true" role="tab">
+        <button data-controller="tab" data-action="click->tab#switch" class="active" aria-label="About this item" aria-controls="about" aria-selected="true" role="tab">
           <svg class="MuiSvgIcon-root" focusable="false" viewBox="0 0 24 24" aria-hidden="true"><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm1 15h-2v-6h2v6zm0-8h-2V7h2v2z"></path></svg>
         </button>
-        <button data-action="click->companion-window#displayContents" data-companion-window-target="leftButton"  aria-label="Content" aria-controls="media-content" role="tab">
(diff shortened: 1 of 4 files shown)


app/javascript/controllers/tab_controller.js
```diff
@@ -0,0 +1,33 @@
...
+    this.#tabPanelFor(button).hidden = false
+  }
+
+  #deactivate(button) {
+    button.classList.remove("active")
+    button.setAttribute('aria-selected', false)
+    this.#tabPanelFor(button).hidden = true
+  }
+
+  #tabPanelFor(button) {
```
mjgiarlo, line 29: Can you say more about all the `#` characters in the function names?

app/components/embed/companion_windows_component.html.erb
```diff
@@ -56,27 +56,27 @@
...
     <aside class="left-drawer collapse" data-companion-window-target="leftDrawer" id="left-drawer" aria-label="Sidebar" style="visibility: hidden">
       <nav class="vert-tabs" role="tablist">
-        <button data-action="click->companion-window#displayMetadata" data-companion-window-target="leftButton" class="active" aria-label="About this item" aria-controls="about" aria-selected="true" role="tab">
+        <button data-controller="tab" data-action="click->tab#switch" class="active" aria-label="About this item" aria-controls="about" aria-selected="true" role="tab">
           <svg class="MuiSvgIcon-root" focusable="false" viewBox="0 0 24 24" aria-hidden="true"><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm1 15h-2v-6h2v6zm0-8h-2V7h2v2z"></path></svg>
         </button>
-        <button data-action="click->companion-window#displayContents" data-companion-window-target="leftButton"  aria-label="Content" aria-controls="media-content" role="tab">
+        <button data-controller="tab" data-action="click->tab#switch" data-tabs-target-param="contents" aria-label="Content" aria-controls="media-content" role="tab">
```
mjgiarlo, line 62: Only one button needs a target param?

app/javascript/controllers/tab_controller.js
jcoyne: That's how javascript denotes private methods: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

app/components/embed/companion_windows_component.html.erb
jcoyne: Actually none of them should have one. I'll remove this.

app/javascript/controllers/tab_controller.js
mjgiarlo: @jcoyne I'm not sure I've seen this before in any of our codebases. Cool!

mjgiarlo (approved):

PR #2025, opened 2023-12-14
changing event handling order and linking transcript load to data being ready for a track

### LINKED ISSUE #1968: [sul-embed/stage] TypeError: this.player is undefined
## Backtrace

```
  get captionTracks() {
>    const tracks = this.player.textTracks_?.tracks_
```

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49295/faults/102937867)

hudajkhan:
Background: 

- In the media_tag_controller file, when the videojs object is initialized, we listen for the "loadedmetadata" event. At that point, we fire the vent for "media-loaded".
- In the media component html.erb file, the "media-loaded" event triggers the "persistPlayer" method in the transcript_controller javascript file, which saves the player object in "this.player"
- When the user clicks on the companion window, the transcript "load" method is triggered (you can see this connection again in the html erb file)
- This error appears to be happening because the transcript controller "load" method can throw an error not recognizing "this.player" if the "loadedmetadata" event has not yet happened (and the resulting persistPlayer method has not been executed).
- Separately, based on analysis (noted below), the 'loadedmetadata' event happens before the tracks and cues are all necessarily available.

What this pull request does:

- Decouple the transcript loading from the companion window click.
- Load the transcript when the tracks are available. This event will be after "loadedmetadata" which would mean "this.player" should already be defined.
- The track cue information is not available until at least the first track has been loaded.  (See notes for various event timings below).   This is captured in a new event "media-data-loaded" and captured in "media_component.html.erb" in the data-action attribute on the transcript element.  When this event is fired, the transcript load method is called which will parse the cues for the various tracks and load them.  

Closes #1968

### DIFF
--- a/app/javascript/controllers/media_tag_controller.js
+++ b/app/javascript/controllers/media_tag_controller.js
@@ -73,6 +73,22 @@ export default class extends Controller {
           const event = new CustomEvent('time-update', { detail: timestamp })
           window.dispatchEvent(event)
         })
+       
+        // When at least one of the tracks has been loaded, trigger transcript loading
+        // Before the tracks are loaded, the cues will not be recognized and the transcript will
+        // not load properly
+        player.textTracks().on('addtrack', () => {
+          // Retrieve the track objects.  Unfortunately, the "textTracks()" method did not cooperate
+          const tracks = player.textTracks_?.tracks_
+          // If there is at least one track we can attach the "loadeddata" event to
+          if(tracks && tracks.length > 0) {
+            player.textTracks_.tracks_[0].on('loadeddata', () => {
+              // Trigger this event, which will then lead to the transcript player load method
+              const event = new CustomEvent('media-data-loaded', { detail: player })
+              window.dispatchEvent(event)
+            })
+          }
+        })
       })
       return player
     })
--- a/app/components/embed/media_with_companion_windows_component.html.erb
+++ b/app/components/embed/media_with_companion_windows_component.html.erb
@@ -1,7 +1,7 @@
 <div class="sul-embed-container media-component" id='sul-embed-object' hidden data-controller="media media-tag transcript" data-media-tag-iiif-manifest-value="<%= iiif_v3_manifest_url %>" data-action="media-seek@window->media-tag#seek">
   <header>
     <nav aria-label="Window navigation">
-      <button data-action="click->media#toggleLeft transcript#load" aria-label="Display sidebar" aria-controls="left-drawer">
+      <button data-action="click->media#toggleLeft" aria-label="Display sidebar" aria-controls="left-drawer">
         <span class="button-label"><svg class="MuiSvgIcon-root" focusable="false" viewBox="0 0 24 24" aria-hidden="true"><path d="M3 18h18v-2H3v2zm0-5h18v-2H3v2zm0-7v2h18V6H3z"></path></svg></span>
       </button>
       <h2 class=""><%= viewer.purl_object.title %></h2>
@@ -102,7 +102,7 @@
               </select>
             </div>
           </div>
-          <div class="transcript" data-action="media-loaded@window->transcript#persistPlayer time-update@window->transcript#highlightCue">
+          <div class="transcript" data-action="media-loaded@window->transcript#persistPlayer media-data-loaded@window->transcript#load time-update@window->transcript#highlightCue">
             <div data-transcript-target="outlet"></div>
           </div>
         </section>
--- a/app/javascript/controllers/transcript_controller.js
+++ b/app/javascript/controllers/transcript_controller.js
@@ -12,10 +12,14 @@ export default class extends Controller {
     this.player = evt.detail
   }
 
-  // We can't load right away, because the VTT tracks may not have been parsed yet. So we wait until this panel is revealed.
+  // We can't load right away, because the VTT tracks may not have been parsed yet. 
+  // We are going to load this panel when at least one track has been loaded.
+  // This function is triggered by the 'media-data-loaded' event which is triggered
+  // by the 'loadeddata' event on the first track.  
   load() {
-    if (this.loaded || !this.currentCues())
+    if (this.loaded || !this.currentCues()) 
       return
+
     this.revealButton()
     this.setupTranscriptLanguageSwitching()
     this.renderCues()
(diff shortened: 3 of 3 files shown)


app/javascript/controllers/media_tag_controller.js
```diff
@@ -73,6 +73,22 @@ export default class extends Controller {
           const event = new CustomEvent('time-update', { detail: timestamp })
           window.dispatchEvent(event)
         })
+
```
mjgiarlo, line 76: Wondering if `loadedmetadata` will reliably give us `textTracks()` if we attach hooks as part of the ready function. Something like:

```javascript
  initializeVideoJSPlayer() {
    this.players = this.mediaTagTargets.map((mediaTag) => {
      mediaTag.classList.add('video-js', 'vjs-default-skin')
      return videojs(mediaTag.id, {}, () => {
        this.on('loadedmetadata', () => {
          const event = new CustomEvent('media-loaded', { detail: player })
          window.dispatchEvent(event)

          // Stop the `loadedmetadata` event and don't bother listening for
          // `timeupdate` events until after `loadedmetadata` fires completes.
          // Why? If we don't, `timeupdate` will be triggered before the video
          // (and its captions) are fully present, and we use the custom
          // `time-update` event to scroll the transcript panel.
          this.off('loadedmetadata')
          this.on('timeupdate', () => {
            const timestamp = this.currentTime()
            const event = new CustomEvent('time-update', { detail: timestamp })
            window.dispatchEvent(event)
          })

          // TODO: Check what's up with this.textTracks() now? Is it reliably present?
          // If so, make sure there are tracks and emit the custom event?
        })
      })
    })
  }
```

mjgiarlo: It may make no difference at all...

hudajkhan: I tried out several approaches to understand the timing of various events better.  I will add these as comments below here as well on the original issue.  

The summary is the "ready" method passed to the videojs constructor (i.e. player = videojs(mediaTag.id, {}, () => { ...}) is actually called a little before the event handler attached to the "ready" event (i.e. player.on('ready')).  Since "on ready" still doesn't return the cues for the tracks, and that only seems to happen after a given track has been loaded, neither of these are the appropriate time to  trigger transcript loading.

hudajkhan:
I'll be adding these comments to the issue as well.
I tried to capture the time (hour, minute, second, millisecond) when different events seem to be triggered/handled.  Using the approach in the pull request, I captured the following: 

- player loadedmetadata 17-1-52-994 
- player loadeddata 17-1-52-995 
- textTracks() addtrack 17-1-52-997
- player ready 17-1-52-998
- first track loadeddata 17-1-53-21

The first part identifies the object on which we are capturing the event, e.g. "player loadedmetadata" corresponds to "player.on('loadedmetadata')".  For the player, the events that are triggered are 'loadedmetadata', followed by 'loadeddata', followed by 'ready'.  We also tried to capture when 'addtrack' was triggered on the textTrack list (i.e. textTracks()) and when the 'loadeddata' method was triggered on the first track object.  As can be seen, the first track loadeddata is the last to fire.  Also importantly, I tried seeing when the tracks and cues were available for retrieval, and cue retrieval only works after the first track is loaded.

hudajkhan:
Based on @mjgiarlo's suggestion above, I also tried out the following code setup:
`const player = videojs(mediaTag.id, {}, () => {` with the part in the {} representing the "ready" method for the player.  

The timings captured were:

- player loadedmetadata 10-4-0-119 
- player loadeddata 10-4-0-119 
- player ready method 10-4-0-119 
- player ready 10-4-0-123 

The "ready" method is called after the "loadeddata" event, but before the "ready" event handler (i.e. player.on('ready').  
I also checked to see which tracks/cues were available.  When the ready method is called, "player.textTracks_.tracks_.length" returns 0.  When the ready event handler is called (which happens later), this expression now returns 1 (in my example there was only one track) but the cues are still not available for processing as the code method to retrieve caption tracks still returns an empty array.

app/components/embed/companion_windows_component.html.erb
hudajkhan, line 1: Decoupled "transcript#load" from the companion window toggle, since the transcript will load when the first track is ready to parse

app/components/embed/companion_windows_component.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="sul-embed-container companion-window-component" id='sul-embed-object' hidden data-controller="companion-window <%= @stimulus_controller %>" data-media-tag-iiif-manifest-value="<%= iiif_v3_manifest_url %>" data-action="media-seek@window->media-tag#seek">
   <header aria-label="Embed main">
     <nav aria-label="Embed navigation">
-      <button data-action="click->companion-window#toggleLeft transcript#load" aria-label="Display sidebar" aria-controls="left-drawer">
+      <button data-action="click->companion-window#toggleLeft" aria-label="Display sidebar" aria-controls="left-drawer">
```
hudajkhan, line 4: Decoupled "transcript#load" from the companion window toggle, since the transcript will load when the first track is ready to parse

app/components/embed/media_component.html.erb
```diff
@@ -28,7 +28,7 @@
           </select>
         </div>
       </div>
-      <div class="transcript" data-action="media-loaded@window->transcript#persistPlayer time-update@window->transcript#highlightCue">
+      <div class="transcript" data-action="media-loaded@window->transcript#persistPlayer media-data-loaded@window->transcript#load time-update@window->transcript#highlightCue">
```
hudajkhan, line 31: media-data-loaded is a custom event that fires when the first track has a "loadeddata" event.  When this event fires, the transcript load function is executed

app/javascript/controllers/media_tag_controller.js
```diff
@@ -73,6 +73,22 @@ export default class extends Controller {
           const event = new CustomEvent('time-update', { detail: timestamp })
           window.dispatchEvent(event)
         })
+       
+        // When at least one of the tracks has been loaded, trigger transcript loading
+        // Before the tracks are loaded, the cues will not be recognized and the transcript will
+        // not load properly
+        player.textTracks().on('addtrack', () => {
```
hudajkhan, line 80: when videojs starts to parse/add tracks, we can check if the tracks exist, and if so, will trigger the "media-data-loaded" event

jcoyne (approved):

PR #2032, opened 2023-12-15
[OLD] IIIF Auth v2 interaction for PDFs

### LINKED ISSUE #1919: PDF Viewer Auth Flow: Implement sul-embed UI

jcoyne:
Fixes #1919 - Implement authorization flow for PDFs.

Using the new IIIF auth/token/probe services and setting it up so it can be re-used later for other content types

Todos, see code comments.

NOTE: use rebased branch over here: #2075

~~Also, figure out where the messages are coming from the access service in the IIIF response, as they are pretty generic right now ("Login" and "Stanford login")~~ In PURL at sul-dlss/purl#925

### DIFF
--- a/app/components/embed/login_component.html.erb
+++ b/app/components/embed/login_component.html.erb
@@ -1,0 +1,10 @@
+<div data-file-auth-target="loginPanel" hidden>
+    <div class="authLinkWrapper">
+        <svg class="MuiSvgIcon-root" focusable="false" aria-hidden="true" viewBox="0 0 24 24"><path d="m14.5 14.2 2.9 1.7-.8 1.3L13 15v-5h1.5v4.2zM22 14c0 4.41-3.59 8-8 8-2.02 0-3.86-.76-5.27-2H4c-1.15 0-2-.85-2-2V9c0-1.12.89-1.96 2-2v-.5C4 4.01 6.01 2 8.5 2c2.34 0 4.24 1.79 4.46 4.08.34-.05.69-.08 1.04-.08 4.41 0 8 3.59 8 8zM6 7h5v-.74C10.88 4.99 9.8 4 8.5 4 7.12 4 6 5.12 6 6.5V7zm14 7c0-3.31-2.69-6-6-6s-6 2.69-6 6 2.69 6 6 6 6-2.69 6-6z"></path></svg>
+        <p data-file-auth-target="loginMessage" class="loginMessage">Stanford users: log in to access all available features.</p>
+        <button data-file-auth-target="loginButton" data-action="file-auth#login"
+                 data-file-auth-messageId-param="" data-file-auth-url-param="">
+                Log in
+        </button>
+    </div>
+</div>
--- a/app/javascript/controllers/file_auth_controller.js
+++ b/app/javascript/controllers/file_auth_controller.js
@@ -1,0 +1,124 @@
+import { Controller } from "@hotwired/stimulus"
+// import validator from 'src/modules/validator'
+// import mediaTagTokenWriter from 'src/modules/media_tag_token_writer'
+// import buildThumbnail from 'src/modules/media_thumbnail_builder'
+
+export default class extends Controller {
+  static targets = ["container"]
+  static values = {
+    iiifManifest: String
+  }
+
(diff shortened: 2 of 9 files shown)


app/components/embed/login_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<div data-file-auth-target="loginPanel" hidden>
+    <div class="authLinkWrapper">
+        <svg class="MuiSvgIcon-root" focusable="false" aria-hidden="true" viewBox="0 0 24 24"><path d="m14.5 14.2 2.9 1.7-.8 1.3L13 15v-5h1.5v4.2zM22 14c0 4.41-3.59 8-8 8-2.02 0-3.86-.76-5.27-2H4c-1.15 0-2-.85-2-2V9c0-1.12.89-1.96 2-2v-.5C4 4.01 6.01 2 8.5 2c2.34 0 4.24 1.79 4.46 4.08.34-.05.69-.08 1.04-.08 4.41 0 8 3.59 8 8zM6 7h5v-.74C10.88 4.99 9.8 4 8.5 4 7.12 4 6 5.12 6 6.5V7zm14 7c0-3.31-2.69-6-6-6s-6 2.69-6 6 2.69 6 6 6 6-2.69 6-6z"></path></svg>
+        <p data-file-auth-target="loginMessage" class="loginMessage">Stanford users: log in to access all available features.</p>
+        <button data-file-auth-target="loginButton" data-action="file-auth#login"
+                 data-file-auth-messageId-param="" data-file-auth-url-param="">
+                Log in
+        </button>
```
peetucket, line 8: note: these labels are placeholders and are replaced with the values returned by the IIIf access service via the file auth js controller

app/javascript/controllers/file_auth_controller.js
```diff
@@ -0,0 +1,231 @@
...
+    document.body.appendChild(this.iframe)
+  }
+
+  // Open the login window in a new window and then poll to see if the auth credentials are now active.
+  loginNeeded(activeAccessService, messageId) {
+    console.log(activeAccessService)
+    this.loginPanelTarget.hidden = false
+    this.loginButtonTarget.innerHTML = activeAccessService.confirmLabel.en[0]
+    this.loginButtonTarget.setAttribute('data-file-auth-messageId-param', messageId)
+    this.loginButtonTarget.setAttribute('data-file-auth-url-param', activeAccessService.id)
```
peetucket, line 203: is there a more canonical way to set the attributes on the stimulus target?

peetucket:
NOTE: use rebased branch over here: #2075

PR #2046, opened 2023-12-19
Ensure transcript is loaded when tracks and cues are available

hudajkhan:
Background:
The code fires the "media-data-loaded" event which triggered the transcript load function. This custom event thus has to fire for the transcript button and the transcript itself to be available in the companion window. 

The event was tied to "addtrack" being fired on this.player.textTracks().  Although this event had fired when I was developing locally with an MP4 video, the event never fired when testing this code in stage on various browsers (Chrome, Firefox, Safari).   

I tried many different event combinations and finally landed on one that seemed to work. 

What this pull request does:

- Uses the "loadeddata" event to trigger the "load" function by firing the "media-data-loaded" event. 
- The html erb file related to transcript also needed the correct aria-control and data-controller attributes specified in order to work. 
- In addition, we retrieved the cues in a way that worked across browsers. (Safari doesn't recogize track.cues.cues_ while Firefox and Chrome don't think track.cues is iterable). 
- Also, the seek event was throwing an error saying this.player was undefined.  We needed to update the code there so it could retrieve the videojs player correctly. 

Originally, the code had various checks for if the captions were available and then if the cues were available for the first track before firing the "media-data-loaded" event.   Recently, the  captions and cues seemed to be consistently available when the videoplayer "loadeddata" event fires, so we removed the extra checks for now.


app/javascript/controllers/media_player_controller.js
```diff
@@ -20,27 +20,61 @@ export default class extends Controller {
...
+        if(cuesAvailable) {
+          // Trigger the media-data-loaded event which will then lead to the transcript load method
+          const event = new CustomEvent('media-data-loaded', { detail: this.player })
+          window.dispatchEvent(event)
+        } else {
+          // If the cues are not available, try iterating every half second a maximum of four times
+          let intervalTries = 0;
+          const intervalId = window.setInterval(() => {
+            intervalTries += 1
+            if(_this.hasCaptionTrackCues(this.firstCaptionTrack())) {
```
jcoyne, line 40: I don't see where `_this` is defined.  Should it be `this`?

app/javascript/controllers/media_player_controller.js
```diff
@@ -20,27 +20,61 @@ export default class extends Controller {
...
+              const event = new CustomEvent('media-data-loaded', { detail: _this.player })
```
jcoyne, line 41: here too.

jcoyne (changes requested):

app/javascript/controllers/media_player_controller.js
hudajkhan: "this" meant something different inside the setInterval scope. I updated the code to have "_this" defined.  Open to suggestions for handling scope in a better way.

app/javascript/controllers/media_player_controller.js
hudajkhan: Same as above - code has been updated, so I'll leave the first comment for now.

app/javascript/controllers/media_player_controller.js
hudajkhan: Removed the iteration and extra checks so this is no longer needed.

jcoyne (approved):

PR #2053, opened 2023-12-20
PDF Authorization using IIIF Probe Service

peetucket:
Based on #2032 

Removed some code I didn't think was needed and made it work for content that is world viewable or for stanford only content where you are already authed, but it doesn't work when you need to auth to continue.

The issue:
- this code always fetches a token before calling the probe service, but you can't get a token if you are not authorized... so it seems like we should just call the probe service first without a token, which lets us know if we even need one and how to login if possible to get one
- BUT calling the probe service without a token throws a CORs error...and not sure why this would make a difference or if has to do with the pre-flight options in some way

Still need to work out:

- [x] ~~changes to stacks needed to authorize access to files using the tokens granted by the probe token service (in addition to the current method)~~ Already exists
- [x] ~~refactoring of media authorization login flow into it's own component for re-use with other content types~~ HOLD for now
- [x] update this PR to make use of the above two items -- not needed


peetucket:
Closed in favor of #2032

PR #2078, opened 2024-01-10
Update design for item not accessible message

peetucket:
Two similar design changes in this PR.

1. When a PDF is not accessible due to it being no download/citation only, the UI currently looks like this:

Screen Shot 2024-01-09 at 4 19 53 PM

This updates it to more closely match the requested design in the [Google Doc](https://docs.google.com/document/d/1IIPYw4CVIaNCjPjtY427JxcPQngd5ZxgqhQLfhxVbi4/edit) (shown below).

Screen Shot 2024-01-10 at 10 07 57 AM

Here is what it looks like in this PR:

Screen Shot 2024-01-09 at 4 27 18 PM

2.  When a PDF is not viewable in the new PDF component (e.g. due to lack of browser support), the UI currently looks like this:

Screen Shot 2024-01-09 at 4 35 29 PM

This PR changes it to this (note that this change is only relevant for the new PDF viewer, not the legacy PDF viewer).

Screen Shot 2024-01-09 at 4 34 42 PM

### DIFF
--- a/app/assets/stylesheets/pdf.scss
+++ b/app/assets/stylesheets/pdf.scss
@@ -70,3 +70,27 @@
     }
   }
 }
+
+.authLinkWrapper {
+  --active-color: rgb(140, 21, 21);
+  align-items: center;
+  display: flex;
+  background-color: var(--active-color);
+  color: white;
+  padding: .5rem;
+
+  .MuiSvgIcon-root {
+    fill: currentColor;
+    width: 1em;
+    height: 1em;
+    display: inline-block;
+    font-size: 1.5rem;
+    flex-shrink: 0;
+  }
+
+  .loginMessage {
+    margin-bottom: 0;
+    margin-left: 1rem;
+    margin-top: 0;
+  }
+}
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/pdf.scss
```diff
@@ -70,3 +70,27 @@
...
+    font-size: 1.5rem;
+    flex-shrink: 0;
+  }
+
+  .loginMessage {
+    margin-bottom: 0;
+    margin-left: 1rem;
+    margin-top: 0;
+  }
+}
```
peetucket, line 96: these are some styles from the `companion_window` component we need for the legacy PDF viewer....  this whole CSS file will go away when the legacy component goes away

app/components/embed/legacy_pdf_component.html.erb
```diff
@@ -40,8 +40,9 @@
           <% # NOTE: Remove legacy_pdf_viewer.js file when this component is removed %>
           <%= javascript_pack_tag('legacy_pdf_viewer') %>
       <% else %>
-          <div style="padding: 10%;">
-            <strong><%= I18n.t('restrictions.not_accessible') %></strong>
+          <div class="authLinkWrapper">
```
jcoyne, line 43: Can we extract this whole chunk to a new component and share it between the LegacyPdf, Media and Pdf components?

jcoyne (changes requested):

app/components/embed/legacy_pdf_component.html.erb
peetucket: Yes, good idea.  Done.

jcoyne (approved):

PR #2082, opened 2024-01-11
Pdf viewer select and view other files

### LINKED ISSUE #2079: New PDF Viewer should show list of files for rendering
Now that we've changed the PDF controller to use the CompanionWindowComponent in #2075, the new PDF viwer no longer shows the list of PDFs available for download, which is important to show when there is more than one PDF available, since only the first PDF is shown in the viewer.

We need to update the left side panel that currently just says "Media Content" to show the list of PDFs available for rendering.  Click a file, and it updates the PDF shown in the main panel.

Screen Shot 2024-01-10 at 10 49 34 AM
.

peetucket:
Fixes #2079  - allow the user to select other PDFs in the PDF viewer and view them in the main panel.

Provides some facility to extending this to other content types in the future when we generalize the design.

Example with two files (shows content selector panel): https://sul-purl-stage.stanford.edu/bg077wm0255
Example with one file (no content selector panel): https://sul-purl-stage.stanford.edu/cv071xh8846

**Before:**

Screen Shot 2024-01-10 at 4 17 08 PM

**Now:**

Screen Shot 2024-01-18 at 9 27 52 AM

The icons are dependent on the file type, and are the same ones use in the File viewer (in this case, it's the PDF icon).

NOTE: If only a single PDF file is available, the content list panel is not shown (for PDFs):

Screen Shot 2024-01-12 at 3 21 39 PM


app/components/embed/companion_windows_component.html.erb
```diff
@@ -20,7 +20,7 @@
         <button data-controller="tab" data-action="click->tab#switch" class="active" aria-label="About this item" aria-controls="about" aria-selected="true" role="tab">
           <svg class="MuiSvgIcon-root" focusable="false" viewBox="0 0 24 24" aria-hidden="true"><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm1 15h-2v-6h2v6zm0-8h-2V7h2v2z"></path></svg>
         </button>
-        <button data-controller="tab" data-action="click->tab#switch" aria-label="Content" aria-controls="media-content" role="tab">
+        <button data-controller="tab" data-action="click->tab#switch" aria-label="Content" aria-controls="object-content" role="tab">
           <svg class="MuiSvgIcon-root" focusable="false" aria-hidden="true" viewBox="0 0 24 24"><path d="M4 10.5c-.83 0-1.5.67-1.5 1.5s.67 1.5 1.5 1.5 1.5-.67 1.5-1.5-.67-1.5-1.5-1.5zm0-6c-.83 0-1.5.67-1.5 1.5S3.17 7.5 4 7.5 5.5 6.83 5.5 6 4.83 4.5 4 4.5zm0 12c-.83 0-1.5.68-1.5 1.5s.68 1.5 1.5 1.5 1.5-.68 1.5-1.5-.67-1.5-1.5-1.5zM7 19h14v-2H7v2zm0-6h14v-2H7v2zm0-8v2h14V5H7z"></path></svg>
```
peetucket, line 24: trying to stay away from "media" terminology since this component is now used by the PDF as well

app/components/embed/companion_windows_component.html.erb
```diff
@@ -37,20 +37,13 @@
...
         </section>
 
-        <section id="media-content" role="tabpanel" hidden>
-          <div class="header-background">
-            <h3>Media content</h3>
-          </div>
-          <ul class="left-drawer-content" data-controller="content-list" data-action="thumbnails-found@window->content-list#draw" role="tablist"></ul>
+        <section id="object-content" role="tabpanel" hidden>
+          <%= render ContentListComponent.new(viewer:, heading: viewer.content_heading) %>
         </section>
```
peetucket, line 43: Use our new component for rendering content in this panel

app/components/embed/companion_windows_component.html.erb
```diff
@@ -37,20 +37,13 @@
...
 
         <%= drawer_content %>
 
         <section id="rights" role="tabpanel" hidden>
-          <div class="header-background">
-            <h3>Rights</h3>
-          </div>
-
           <%= render RightsComponent.new(viewer:) %>
         </section>
```
peetucket, line 48: pulled the heading into the existing `RightsComponent` to match what is done in the new `ContentList` component (which needs to have the heading vary by content type)

app/javascript/controllers/content_list_controller.js
```diff
@@ -2,10 +2,17 @@ import { Controller } from "@hotwired/stimulus"
...
 
 export default class extends Controller {
-  // This listens for a thumbnails-found event. Then it draws each on the page.
-  draw(evt) {
+  // This listens for a thumbnails-found event. Then it draws the media thumbnails in the contents panel
+  drawThumbnails(evt) {
     const thumbnails = evt.detail.
       map((thumbnailData, index) => new Thumbnail(thumbnailData).build(index))
     this.element.innerHTML = thumbnails.join('')
   }
```
peetucket, line 10: make this method name more explicitly about drawing thumbnails

app/javascript/controllers/content_list_controller.js
```diff
@@ -2,10 +2,17 @@ import { Controller } from "@hotwired/stimulus"
...
+
+  // show a different PDF when the user selects it in the list
+  showPdf(evt) {
+    evt.preventDefault()
+    const file_uri = evt.currentTarget.dataset.url
+    window.dispatchEvent(new CustomEvent('auth-success', { detail: file_uri }))
+  }
 }
```
peetucket, line 18: this new method allows us to swap to different PDFs when the user selects them

app/components/content_list_item_component.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  # Only PDF content is rendered in this panel for now
+  # Media content also does not need this component since it uses custom javascript to render thumbnails and link to content
+  def render?
+    return true if viewer_class == "Embed::Viewer::PdfViewer" and url.ends_with?('.pdf') # do not render any file that is not a PDF
+
+    false
+  end
+end
```
peetucket, line 34: New components which render the list of Contents in the side panel.  Designed to be flexible when extending to new content types

app/components/embed/companion_windows_component.html.erb
```diff
@@ -37,20 +37,13 @@
           <dl data-controller="iiif-metadata" class="metadataList left-drawer-content" data-action="iiif-manifest-received@window->iiif-metadata#drawMetadata"></dl>
         </section>
 
-        <section id="media-content" role="tabpanel" hidden>
-          <div class="header-background">
-            <h3>Media content</h3>
-          </div>
-          <ul class="left-drawer-content" data-controller="content-list" data-action="thumbnails-found@window->content-list#draw" role="tablist"></ul>
+        <section id="object-content" role="tabpanel" hidden>
+          <%= render ContentListComponent.new(viewer:, heading: viewer.content_heading) %>
```
jcoyne, line 41: How will this affect the media component?

jcoyne (changes requested):

app/javascript/controllers/content_list_controller.js
```diff
@@ -2,10 +2,17 @@ import { Controller } from "@hotwired/stimulus"
...
-  draw(evt) {
+  // This listens for a thumbnails-found event. Then it draws the media thumbnails in the contents panel
+  drawThumbnails(evt) {
     const thumbnails = evt.detail.
       map((thumbnailData, index) => new Thumbnail(thumbnailData).build(index))
     this.element.innerHTML = thumbnails.join('')
   }
+
+  // show a different PDF when the user selects it in the list
+  showPdf(evt) {
```
jcoyne, line 13: This behavior is already in `ThumbnailController#activate`. I don't think we should make this special solution for PDFs.

jcoyne (changes requested):

app/components/embed/companion_windows_component.html.erb
peetucket: Media content is unaffected partly because of this (i.e. the component does nothing if not in the PDF viewer and a PDF file in the object): https://github.com/sul-dlss/sul-embed/pull/2082/files#diff-42a38cc64b6677f81d752edaa0880848b977216a28209c764ffeaf15ff57699fR30-R35

But also, the javascript method that draws the media thumbnails completely replaces the HTML content with the dynamically generated list of thumbnails anyway: https://github.com/sul-dlss/sul-embed/pull/2082/files#diff-72decc26473133ce4c39fa2e64877e55a38bbb13fb955b948cb6efd805155521L7-L9

So the component not rendering anything for media just saves us some time producing something that would be get replaced later anyway.

Screen shot: 
Screen Shot 2024-01-11 at 9 12 49 AM

app/javascript/controllers/content_list_controller.js
jcoyne: In either case, I think the action that handles switching the viewer should be in the same controller for both.  It's confusing to have the "change item" be in one place for media and another place for PDF.

peetucket: Makes sense.  I moved it to the content-list controller and named it similarly to the new `showPdf` method for consistency.

app/javascript/controllers/content_list_controller.js
```diff
@@ -2,10 +2,36 @@ import { Controller } from "@hotwired/stimulus"
...
+    const tabs = this.element.closest('[role="tablist"]').querySelectorAll('[role="tab"]')
+    tabs.forEach((target) => {
+      target.classList.remove('active')
+      target.setAttribute("aria-selected", false)
+    })
+
+    const target = this.element
+    target.classList.add('active')
+    target.setAttribute("aria-selected", true)
+  }
```
peetucket, line 28: Moved from `app/javascript/controllers/thumbnail_controller.js` controller and renamed for consistency.  Updated references in the code to ensure it works.

app/components/embed/companion_windows_component.html.erb
jcoyne: Why don't we render this data out of the manifest.json? It seems like we're moving toward using the manifest for more of the information, so why not here too?

peetucket: I did consider doing it client side in javascript parsing the manifest like how it's done for the media thumbnails, but it was more complicated and required writing more code, which seemed unnecessary given the simplicity of what needs to happen for PDFs.  So basically because it's easier to do server side with existing methods (see https://github.com/sul-dlss/sul-embed/blob/bb92d7125aa7c74d5f870eb637f9a8756e2429e4/app/models/embed/purl.rb#L58-L60) rather than client side in javascript, given that we can re-use existing methods and technique (e.g. already used by the download panel in the same player).

app/viewers/embed/viewer/file.rb
```diff
@@ -17,14 +17,6 @@ def height
...
 
-      def file_type_icon(mimetype)
-        if Constants::FILE_ICON[mimetype].nil?
-          'sul-i-file-new-1'
-        else
-          Constants::FILE_ICON[mimetype]
-        end
-      end
-
       def file_size_text(file_size)
```
peetucket, line 20: Move this method above into the common viewer class so we can use it for the PDF viewer too

peetucket:
@jcoyne trying to close out this work from the previous workcycle ... any other concerns?

peetucket:
@jcoyne: Trying to wrap this work up, can you check my comments so we can move forward, or remove your request changes if you want someone else to review?

thanks

app/views/pages/sandbox.html.erb
```diff
@@ -207,6 +207,8 @@
       <li>
         pdf (object emgargoed, file available) - bb990sg8412
       </li>
+      <li>
+        pdf (multiple files) - ds777pr3860
```
thatbudakguy, line 211: does this need a closing tag?

thatbudakguy (approved):

app/views/pages/sandbox.html.erb
peetucket: yup, thanks -- added

PR #2084, opened 2024-01-11
Show "Logged in panel" after user completes authentication in PDF viewer

peetucket:
This is an attempt to match what happens with the new media player in the new PDF player.  If the user is presented with a Stanford only PDF, and they are asked to login, once login is complete, the "You are logged in" message will be shown, sort as as described/shown [here](https://docs.google.com/document/d/1IIPYw4CVIaNCjPjtY427JxcPQngd5ZxgqhQLfhxVbi4/edit) (shown below).

Screen Shot 2024-01-10 at 10 07 57 AM

The main difference is that there isn't a "Log out" button, there is an "X" which dismisses the message.  If you come back to the a Stanford only PDF after authenticating once, you won't see this panel anymore, you will just immediately see the PDF, so it is basically just an indication that you have authed successfully.

**When you need to login:**
Screen Shot 2024-01-11 at 2 11 20 PM

**After logging in:**
Screen Shot 2024-01-11 at 2 10 22 PM


app/components/embed/pdf_component.html.erb
```diff
@@ -1,15 +1,15 @@
...
+            data-action="iiif-manifest-received@window->file-auth#parseFiles auth-success@window->pdf#show">
+          </div>
     <% else %>
-       <div class='sul-embed-body sul-embed-pdf' style="width: 100%">
-         <%= render Embed::ContentNotAvailableComponent.new %>
-       </div>
+       <%= render Embed::ContentNotAvailableComponent.new %>
     <% end %>
+    </div>
   <% end %>
```
peetucket, line 14: some re-arranging of the divs here so that we can show the login/logout messages independently of the full PDF (otherwise when we replace the `sul-embed-pdf` div with the actual PDF, it would have also destroyed out "logged in" message)

jcoyne (changes requested):

config/locales/en.yml
```diff
@@ -23,7 +23,7 @@ en:
   hello: "Hello world"
   restrictions:
     not_accessible: 'This item cannot be accessed online. See Access conditions for more information.'
-    download_the_file: 'download the file'
+    logged_in: 'Logged in.'
   date:
```
peetucket, line 27: Add new translation, remove unused translation

jcoyne (approved):

PR #2085, opened 2024-01-11
adding default attribute for first alphabetical track to fix Safari transcript loading error

### LINKED ISSUE #2062: When video has multiple caption files, Safari does not load all the cues for the other tracks
Context:
The transcript controller JS file retrieves the text tracks from the VideoJS player, and expects the caption tracks to have the cues already available and populated.  In Firefox and Chrome, all the cues are loaded when the TextTrack objects are available.

In Safari, we have seen the following happening instead:
a) The cues are loaded only for the TextTrack which corresponds to the captions that are currently playing/active in the video.  In other words, if the first track is Russian and selected by default in the captions list visible within the video player itself and the second track is English, the cues will be loaded for the Russian track but not the English track.
b) The current logic in the code expects all the cues to be available, so it populates the transcript language dropdown based on which tracks it sees which also have cues available.  In this case, the code will see the Russian cues but not the English ones, so will only provide Russian in the transcript language dropdown.

We made some attempts to see if we could change the English track mode from "disabled" to "hidden" or "showing", but the cues were still not available after we changed the mode.

### LINKED ISSUE #2068: Sidebar transcript tab does not load in Safari
In Safari only, the Sidebar transcript tab does not load. 

PURL Examples:
- https://purl.stanford.edu/qf378nj5000
- https://purl.stanford.edu/gc743bz7434

Image

Console logs:
- `"Failed to load resource: the server responded with a status of 403 (Forbidden)" from https://sul-streamengine-prod.stanford.edu/stacks/...`
- `"VIDEOJS: WARN: Text Tracks are being loaded from another origin but the cross origin attribute isn't used. This may prevent text tracks from loading."`

In Firefox and Chrome, the Sidebar transcript tab loads as expected.
Image

hudajkhan:
Closes #2068 .  And should close #2062 

Background: 
(a) In Safari, videojs text tracks do not have cues loaded for any track that is not already selected in the video player. Our transcript code retrieves all the cues for the different caption tracks and loads them when the video player has dispatched events indicating some level of readiness. Unfortunately, in the case of Safari, while the code can see caption tracks even if they are not selected in the video player itself, the cues are not available and will return length zero (even if console messages can show us the cues themselves).

(b) There are OS settings on Macs that will determine if a caption is selected by default in the video player. Alan and Andrew tested out the settings and found when the Accessibility "Prefer closed captions and SDH" setting was turned off, the video player had no captions selected by default. In that case, the transcript would not load. When they turned the setting on, a caption file was selected by default and the transcript would load in the sidebar.

What this PR does:

- For the first caption track (alphabetically sorted for multiple caption files), this code adds the "default" attribute to the track.  This also selects that caption track in the video player itself by default.  (If this behavior is not preferred, we will have to find a different solution).  
- Because this track is now selected, the cues are now available to our code.  The transcript is thus able to load. 
- In addition, in the context of Safari, the code converts any caption tracks with mode "disabled" to mode "hidden", enabling the cues to be retrievable in Safari.  We also use "remoteTextTracks" because mode changes work with that code but do not with the base "TextTracks" method.  Also, when the video player is first loaded, a slight delay is required once disabled mode tracks are changed to hidden for the cues to actually be available to the code and to return a non-zero length.


hudajkhan:
Discussed with @andrewjbtw and @dinahhandel: agreement that we can pursue this approach.

hudajkhan:
Testing this out on safari for an item with multilingual captions https://sul-purl-stage.stanford.edu/wm783fb6812: When the video first loads, English is selected in the video caption list.  The English transcript is visible in the sidebar, and the highlighting and auto-scrolling work correctly.

When Russian is selected in the video, the highlighting no longer works in the transcript.  This is an additional error in the multilingual caption situation.

For a video like https://sul-purl-stage.stanford.edu/pf759xf5671, there is only one language and the transcript loads fine.
When captions are turned to off though in the video itself, the transcript highlighting and autoscrolling no longer work.

app/javascript/controllers/transcript_controller.js
```diff
@@ -15,25 +15,77 @@ export default class extends Controller {
...
     this.renderCues()
     this.loaded = true
   }
 
+  // Safari cues require special handling.
+  // We want the track cues to be available so we can properly generate the transcript sidebar language dropdown
+  // if there is more than one language track.
+  convertDisabledTracks() {
+    const captions = this.player.remoteTextTracks()?.tracks_.filter(track => track.kind === 'captions')
+    let hiddenTracks = false
```
mjgiarlo, line 36: I don't see where this variable is used. Maybe cruft?

hudajkhan: Agreed. Removed.

app/javascript/controllers/transcript_controller.js
```diff
@@ -15,25 +15,77 @@ export default class extends Controller {
...
+    captions.forEach(track => {
+      if (track.mode == 'disabled') {
```
mjgiarlo, line 37: I don't think it makes a functional difference, but at least for consistency within this file we might consider checking for strict equality here (`===` vs. `==`).

mjgiarlo (approved):
Left a couple comments, none of them blocking. Thanks for sticking with this work! Looks great. 😄

app/javascript/controllers/transcript_controller.js
hudajkhan: Agreed. Using ===

PR #2089, opened 2024-01-17
set minimum height for PDF main panel so it does not shrink

### LINKED ISSUE #2086: Smaller side panels in companion window shouldn't shrink the main content
When you click on the left sidebar, and view a panel which is shorter in height (because there is less content), it shrinks the overall height of the entire sul-embed panel, which has the effect of shrinking the main content as well.  It should retain the full height of the main panel instead.

For example, here is what a PDF looks like when the page loads:

Screen Shot 2024-01-12 at 1 52 24 PM

Here is what it looks like when you click the left drawer and select the small content list:

Screen Shot 2024-01-12 at 1 53 26 PM

peetucket:
Fixes #2086 - do not shrink the main PDF panel too much even if you open a sidebar that is smaller

In the screenshots below, ignore the fact that the content panel is blank and says "Media content"...this is fixed in #2082

**Without change**
When you view the contents panel
Screen Shot 2024-01-17 at 10 43 01 AM

**With change**
When you view the contents panel
Screen Shot 2024-01-17 at 10 42 30 AM

### DIFF
--- a/app/assets/stylesheets/companion_window.scss
+++ b/app/assets/stylesheets/companion_window.scss
@@ -456,5 +456,5 @@
 }
 
 .sul-embed-pdf {
-  min-height: 700px;
+  min-height: 90vh;
 }
--- a/app/assets/stylesheets/companion_window.scss
+++ b/app/assets/stylesheets/companion_window.scss
@@ -454,3 +454,7 @@
     }
   }
 }
+
+.sul-embed-pdf {
+  min-height: 700px;
+}


app/assets/stylesheets/companion_window.scss
```diff
@@ -454,3 +454,7 @@
     }
   }
 }
+
+.sul-embed-pdf {
+  min-height: 700px;
+}
```
thatbudakguy, line 460:
```suggestion
.sul-embed-pdf {
  min-height: 90vh;
}
```

I tested this and it seems to have more or less the same effect on desktop. On mobile, it prevents the scenario where the container is forced to be too large, and so you can scroll down "out of sight" and the scroll handle vanishes. On both platforms, it got computed to about 468px. Try it and see what you think.

peetucket: Thanks for checking this on mobile.  I made your change and it looks great for me on my laptop (localhost).

thatbudakguy (approved):

PR #2102, opened 2024-02-01
Use new headless Chrome for testing

taylor-steve:
Switch to the new headless Chrome for testing.

With `--headless=old` we did need `--disable-site-isolation-trials` or some tests would fail.

The new headless Chrome crashes on some embed tests with that option. Capybara's default driver definition includes this option. Here we redefine it without that option.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -15,16 +15,16 @@ require 'view_component/system_test_helpers'
 
 # Capybara 3.40.0 started using the new chrome headless by default:
 # https://developer.chrome.com/docs/chromium/new-headless
-# We need a driver with the "--headless=old" argument until
-# our tests work with "new".
-Capybara.register_driver :selenium_chrome_headless_old do |app|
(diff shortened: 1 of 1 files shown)


jcoyne:
@taylor-steve nice find.  Did you report this upstream to the capybara repo?

taylor-steve:
I just did. For reference: https://github.com/teamcapybara/capybara/issues/2742

For further reference, despite initial reports and the issue with the arclight test the other day, I'm finding the new chrome to be *faster* so far.

jcoyne:
@taylor-steve thanks for reporting it upstream, but I don't think you should have closed the issue.  It seems like a problem to me that we can't use the driver built into capybara.

taylor-steve:
That's fair. Their defaults work for us in other projects, so I felt like I'd need to better understand the specifics of the interaction between that option and the crashing tests before going beyond reporting "here's what worked for me".

jcoyne (approved):

PR #2103, opened 2024-02-01
show download all files links in geo viewer

### LINKED ISSUE #2099: Support downloading of the entire object as a .zip in the geo viewer
Related to https://github.com/sul-dlss/gis-robot-suite/issues/673.

This pathway is exposed in the file viewer currently, but not in the geo viewer. Once we do https://github.com/sul-dlss/gis-robot-suite/issues/673 and/or stop shelving pre-zipped archives of GIS data, it will be helpful for users to be able to "just download everything", especially if the total number of shelved items is large (e.g. for a shapefile).

Presumably we could add a "download everything" option to this download menu in the viewer:

Screenshot 2024-01-30 at 16 30 46

peetucket:
Fixes #2099 

For Geo objects, when there is more than 1 downloadable file present in an object, show the "Download all files" link, the same as is done for File objects.  This link will download a ZIP file with all of the files in the object (which may itself include ZIPs). 

This will work right now for all Geo objects, but if we don't want to release just yet, we could wait and/or put a feature flag in.

This is what it looks like (exactly the same as the File viewer):

Screen Shot 2024-02-01 at 12 09 10 PM

### DIFF
--- a/app/components/embed/footer_component.html.erb
+++ b/app/components/embed/footer_component.html.erb
@@ -12,6 +12,7 @@
 
     <% if viewer.show_download? %>
       <button
+        id='sul-embed-download-panel-toggle'
         class='sul-embed-footer-tool btn btn-toolbar sul-i-download-3'
         aria-expanded='false' aria-label='<%= download_file_label %>'
         data-sul-embed-toggle='sul-embed-download-panel'>
@@ -25,6 +26,20 @@
       <a class='sul-embed-footer-tool btn btn-toolbar sul-i-navigation-next-4'
         href="<%= external_url %>" target='_parent'>
         <span class='sul-embed-sr-only'><%= external_url_text %></span>
+      </a>
+    <% end %>
+
+    <% if viewer.display_download_all? %>
+      <a id='sul-embed-footer-download-all' href='<%= viewer.download_url %>' target="_blank" rel="noopener noreferrer nofollow" data-controller="download" data-action="click->download#download">
+        <button class='sul-embed-footer-tool btn btn-toolbar sul-i-download-3' aria-label="download all files">
+          Download all <%= pluralize(viewer.purl_object.downloadable_files.length, "file") %>
+          (<%= viewer.pretty_filesize(viewer.purl_object.size) %>)
+          <% if viewer.any_stanford_only_files? %>
+            <span class="sul-embed-location-restricted-text sul-embed-stanford-only-text" aria-label="restricted to stanford">
+              <span class="sul-embed-text-hide"> Stanford only </span>
+            </span>
+          <% end %>
+        </button>
       </a>
     <% end %>
 
--- a/app/viewers/embed/viewer/common_viewer.rb
+++ b/app/viewers/embed/viewer/common_viewer.rb
@@ -26,6 +26,20 @@ module Embed
         nil
       end
 
+      def any_stanford_only_files?
+        @purl_object.all_resource_files.any?(&:stanford_only?)
+      end
+
+      # indicates if viewer should display the Download All link in footer (override in specific viewer classes)
+      def display_download_all?
+        false
+      end
+
+      # link that will download all of the files in this object
+      def download_url
+        "#{Settings.stacks_url}/object/#{@purl_object.druid}"
+      end
+
       def stylesheet
         "#{purl_object.type}.css"
       end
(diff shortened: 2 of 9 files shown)


app/components/embed/footer_component.html.erb
```diff
@@ -28,6 +29,20 @@
...
+          (<%= viewer.pretty_filesize(viewer.purl_object.size) %>)
+          <% if viewer.any_stanford_only_files? %>
+            <span class="sul-embed-location-restricted-text sul-embed-stanford-only-text" aria-label="restricted to stanford">
+              <span class="sul-embed-text-hide"> Stanford only </span>
+            </span>
+          <% end %>
+        </button>
+      </a>
+    <% end %>
+
```
peetucket, line 45: This makes the download all link available to any viewer that uses the common `FooterComponent`

app/viewers/embed/viewer/common_viewer.rb
```diff
@@ -26,6 +26,20 @@ def external_url
...
+      def display_download_all?
+        false
+      end
+
+      # link that will download all of the files in this object
+      def download_url
+        "#{Settings.stacks_url}/object/#{@purl_object.druid}"
+      end
+
       def stylesheet
```
peetucket, line 43: This moves some code from the File viewer (below) to the common viewer so it can be used by any viewer.  By default, the download all link is not shown, unless overridden by a specific content type viewer.

spec/factories/purls.rb
```diff
@@ -52,7 +52,7 @@
       type { 'geo' }
       bounding_box { [['-1.478794', '29.572742'], ['4.234077', '35.000308']] }
       public { true }
-      contents { [build(:resource, :file, files: [build(:resource_file, filename: 'data.zip'), build(:resource_file, filename: 'data_EPSG_4326.zip')]), build(:resource, :image)] }
+      contents { [build(:resource, :file, files: [build(:resource_file, :world_downloadable, filename: 'data.zip'), build(:resource_file, :world_downloadable, filename: 'data_EPSG_4326.zip')]), build(:resource, :image)] }
     end
```
peetucket, line 57: This ensures our test data mimics reality, which is that geo objects have world downloadable data files.  We want them to have this or else the "download all" link won't be shown for our new test below.

spec/lib/embed/viewer/common_viewer_spec.rb
```diff
@@ -69,4 +69,36 @@
...
+      it { is_expected.to be false }
+    end
+  end
+
+  describe '#download_url' do
+    subject { file_viewer.download_url }
+
+    it { is_expected.to eq 'https://stacks.stanford.edu/object/abc123' }
+  end
 end
```
peetucket, line 104: move specs from File viewer to common viewer since the methods moved above

test/components/previews/embed/geo_component_preview.rb
```diff
@@ -6,8 +6,12 @@ module Embed
...
 
-    def public
-      render_viewer_for(url: 'https://purl.stanford.edu/bc843cm1713')
+    def public_raster
+      render_viewer_for(url: 'https://purl.stanford.edu/tg926kp6619')
+    end
+
+    def public_vector
+      render_viewer_for(url: 'https://purl.stanford.edu/cz128vq0535')
     end
```
peetucket, line 15: same examples to match what is on the sandbox page

app/components/embed/footer_component.html.erb
```diff
@@ -12,6 +12,7 @@
 
     <% if viewer.show_download? %>
       <button
+        id='sul-embed-download-panel-toggle'
         class='sul-embed-footer-tool btn btn-toolbar sul-i-download-3'
```
peetucket, line 16: disambiguate this element from the download all link (which shares classes), helpful for testing

app/viewers/embed/viewer/geo.rb
```diff
@@ -40,6 +40,13 @@ def index_map?
         index_map.present?
       end
 
+      # Returns true or false whether the viewer should display the Download All
+      # link. True for all geo objects, unless there is only one file
+      # (because the file download link will suffice for that)
+      def display_download_all?
+        @purl_object.downloadable_files.length > 1
+      end
```
thatbudakguy, line 48: honestly, this seems like it could even be the default behavior in `common_viewer.rb` and then nothing to override here?

thatbudakguy (approved):
this solution is intuitive to me and avoids the need for new design. nice work!

PR #2155, opened 2024-05-28
Link to collection without requiring a second lookup

### LINKED ISSUE #2154: Items should be visible in embed even if their parent collection is "dark'
### Expected behavior

An item should be visible in embed if the rights are set to enable a purl. This should not depend on the rights status of the parent collection object.

### Actual behavior

When an item is in a collection where the collection object is set to "dark", the item Purl renders but the viewer just shows a 404 for the collection druid:

Screenshot 2024-05-28 at 12 35 46 PM

See this test item in prod: https://purl.stanford.edu/ct931nf5570

### Additional information

I think this issue may have started when we switched to read from public JSON instead of public XML.

jcoyne:
Fixes #2154

The link is only used for analytics, so it shouldn't matter if the collection is dark or not

### DIFF
--- a/app/views/embed/iframe.html.erb
+++ b/app/views/embed/iframe.html.erb
@@ -20,7 +20,7 @@
     <%= javascript_importmap_tags @embed_response.viewer.importmap if @embed_response.viewer.importmap %>
     <%= stylesheet_link_tag @embed_response.viewer.stylesheet %>
     <%= stylesheet_link_tag 'sul_icons.css' %>
-    <%= tag :link, rel: "up", href: Embed::Purl.find(@embed_response.request.purl_object.collections.first).purl_url if @embed_response && @embed_response.request.purl_object.collections.any? %>
+    <%= tag :link, rel: "up", href: @embed_response.request.purl_object.first_collection_url if @embed_response && @embed_response.request.purl_object.first_collection_url %>
   </head>
   <body>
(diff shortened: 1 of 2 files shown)


app/views/embed/iframe.html.erb
```diff
@@ -20,7 +20,7 @@
     <%= javascript_importmap_tags @embed_response.viewer.importmap if @embed_response.viewer.importmap %>
     <%= stylesheet_link_tag @embed_response.viewer.stylesheet %>
     <%= stylesheet_link_tag 'sul_icons.css' %>
-    <%= tag :link, rel: "up", href: Embed::Purl.find(@embed_response.request.purl_object.collections.first).purl_url if @embed_response && @embed_response.request.purl_object.collections.any? %>
+    <%= tag :link, rel: "up", href: @embed_response.request.purl_object.first_collection_url if @embed_response && @embed_response.request.purl_object.first_collection_url %>
```
mjgiarlo, line 23: If a `Purl` instance has no collections, won't `Purl#first_collection_url` `return "#{Settings.purl_url}/#{nil}"` (truthy)?

jcoyne: This is now handled in the model. Thank you.

mjgiarlo: Excellent. :shipit:

mjgiarlo (approved):

PR #2159, opened 2024-05-31
limit rendering to first file in manifest

### LINKED ISSUE #2158: (new pdf viewer) Odd behavior when loading multiple PDFs in the browser PDF viewer
### Expected behavior

The browser-based PDF viewer (enabled on stage and QA) is capable of displaying more than one PDF. Users can switch between PDFs by navigating in the left companion window:

Screenshot 2024-05-29 at 4 21 08 PM

The expected behavior is that the browser loads the first PDF listed in the metadata.

### Actual behavior

I've found it's unpredictable which PDF will load when there's multiple PDFs. Sometimes it's the first one, sometimes it's one later on the list. I've reloaded the same purl in the same browser and seen the browser switch between PDFs upon reload.

Two examples in stage:

- Purl with two pdfs: https://sul-purl-stage.stanford.edu/bg077wm0255
- Purl with six pdfs: https://sul-purl-stage.stanford.edu/kp686sg8638

1. Load those purls and check if the first PDF on the list was the PDF to render
2. Load the purls in different browsers and see if the browsers load the same file first
3. Do a hard refresh and see if the same file is loaded each time. Try 4-5 refreshes, not just one.

### Browser / Environment

I've checked with firefox and chrome and both exhibit the odd behavior of not always loading the first PDF on the list.

dnoneill:
closes #2158 

Stops rendering on load of every page in the manifest.

### DIFF
--- a/app/javascript/controllers/file_auth_controller.js
+++ b/app/javascript/controllers/file_auth_controller.js
@@ -4,6 +4,7 @@ export default class extends Controller {
   static targets = ["container", "loginPanel", "messagePanel", "loginButton", "loginMessage"]
 
   resources = {} // Hash of messageIds to resources
+  firstFile = '' // for multiple files we need to be able to render the first on load
 
   addPostCallbackListener() {
     window.addEventListener("message", (event) => {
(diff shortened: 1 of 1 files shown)


peetucket (approved):
Approved pending deployment to stage or QA and sign off by Andrew

dnoneill:
Approved by andrew.

PR #2164, opened 2024-06-05
update js, css to indicate when pdf is selected

### LINKED ISSUE #2161: (new pdf viewer) Clearly indicate which PDF is selected when there are multiple PDFs
The left companion window should indicate which document has been selected when there are multiple PDFs. Currently, you can't tell which PDF you are viewing.

Screenshot 2024-06-03 at 10 12 00 AM

dnoneill:
closes #2161 

Based on @alundgard comments updated to be 
Screenshot 2024-06-06 at 10 46 12 AM

### DIFF
--- a/app/assets/stylesheets/companion_window.scss
+++ b/app/assets/stylesheets/companion_window.scss
@@ -396,7 +396,7 @@
 
     --thumb-height: 75px;
 
-    .media-thumb {
+    .media-thumb, .file-thumb {
       list-style-type: none;
       border: 1px solid gray;
       margin-bottom: 0.5rem;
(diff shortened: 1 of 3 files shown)


peetucket:
Nice.  Not really my call, but maybe bolded too would make it even more obvious?

alundgard:
Since there isn't a non-color visual indicator of active state, I believe the active state color will need a 3:1 contrast ratio with both the background (white) and the inactive state color (black). https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast#user-interface-components

peetucket (approved):
Approved pending PO approval.

alundgard:
> Approved pending PO approval.

Approved here: https://github.com/sul-dlss/sul-embed/issues/2161#issuecomment-2152950467

PR #2167, opened 2024-06-10
Open the left drawer by default for media player

### LINKED ISSUE #2165: Auto-expand the contents panel in the media viewer
### Requested behavior

To make it easier for users to find multiple resources within a single media item, default to showing the media viewer with the contents panel open when the viewer is first loaded.

Screenshot 2024-06-07 at 10 37 26 AM

We've received quite a bit of feedback from public services staff, curators, and the broader public that it is too difficult to find out when a media item contains multiple parts or components. 

### Current behavior

The viewer currently opens without showing the contents list and it takes multiple clicks + knowledge of where to click to find it.

marlo-longley:
Closes #2165

https://github.com/sul-dlss/sul-embed/assets/1328900/c83ae6f2-c648-431f-9829-9b233a035eb9

### DIFF
--- a/app/javascript/controllers/companion_window_controller.js
+++ b/app/javascript/controllers/companion_window_controller.js
@@ -7,20 +7,29 @@ export default class extends Controller {
                      "downloadModal", "accessibility", "modalComponentsPopover"]
   connect() {
     this.element.hidden = false
+    // Open the drawer on load
+    this.openLeftDrawer()
     EmbedThis.init();
   }
 
   toggleLeft() {
-    const classList = this.leftDrawerTarget.classList
-    if (classList.contains('open')) {
-      classList.remove('open')
-      setTimeout(() => {
-        this.leftDrawerTarget.style.visibility = 'hidden' // remove from accessability tree
-      }, 500)
+    if (this.leftDrawerTarget.classList.contains('open')) {
+      this.closeLeftDrawer()
     } else {
-      this.leftDrawerTarget.style.visibility = ''
-      classList.add('open')
+      this.openLeftDrawer()
     }
+  }
+
+  openLeftDrawer() {
+    this.leftDrawerTarget.classList.add('open')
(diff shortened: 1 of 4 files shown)


jcoyne:
@alundgard is this supposed to open the "metadata" pane or the file list pane?

alundgard:
> @alundgard is this supposed to open the "metadata" pane or the file list pane?

The file list ("Media content") pane.

app/javascript/controllers/companion_window_controller.js
```diff
@@ -7,22 +7,33 @@ export default class extends Controller {
...
-      }, 500)
+      this.closeLeftDrawer(classList)
     } else {
-      this.leftDrawerTarget.style.visibility = ''
-      classList.add('open')
+      this.openLeftDrawer(classList)
     }
   }
 
+  openLeftDrawer(classList) {
```
jcoyne, line 25: It feels a little awkward passing classList as an argument here since we're accessing `this.leftDrawerTarget` elsewhere in the method. Can we make this method parameter-less?

app/javascript/controllers/tab_controller.js
```diff
@@ -1,6 +1,17 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
+  connect() {
+    const objectContentTab = this.#tabList.querySelector('[aria-controls="object-content"]');
```
jcoyne, line 5: This is going to run once for each tab (3 times).  A better way may be to do:

```js
if (this.element.querySelector('[aria-controls="object-content"]').attributes['aria-controls'] === 'object-content')
  this.#makeActive(this.element)
```

jcoyne (changes requested):

marlo-longley:
@jcoyne thank you. I addressed your comments.

jcoyne (approved):

PR #2177, opened 2024-06-20
disable the fullscreen button for the native PDF viewer in Chromium based browsers

### LINKED ISSUE #2162: (new pdf viewer) No escape (literally) from fullscreen PDF in Chrome and Edge
In Chrome, if you full screen a PDF in the browser viewer, you can get stuck without being able to escape from full screen.

To reproduce:

- Go to a PDF in stage where the browser viewer is deployed: https://sul-purl-stage.stanford.edu/bg077wm0255
- Make embed full screen
- Click on the PDF display: just click within the PDF viewer window, highlight text, etc.
- Press escape and see that you can't get out of the full screen

Additional information:

- All my testing is on a Mac (it's what I have)
- I have not been able to reproduce this in Firefox or Safari, just Chrome and Edge
- If you don't click within the PDF window, you can still use escape to leave full screen. The issue only comes up if you interact with the PDF

jmartin-sul:
closes #2162

firefox:
Screenshot 2024-06-20 at 11 58 26 AM

chrome:
Screenshot 2024-06-20 at 12 04 13 PM

edge (uses chromium as its rendering engine, and has the same problem with the esc key not working to exit from full screen):
Screenshot 2024-06-20 at 12 04 47 PM

### DIFF
--- a/app/components/embed/pdf_component.html.erb
+++ b/app/components/embed/pdf_component.html.erb
@@ -1,10 +1,14 @@
 <%= render Embed::CompanionWindowsComponent.new(viewer:, stimulus_controller: 'file-auth') do |component| %>
   <% component.with_header_button do %>
+    <%# chromium has a bug preventing the esc key from exiting full screen in some situations, see
+        https://issues.chromium.org/issues/40131988 and https://github.com/sul-dlss/sul-embed/issues/2162 %>
+    <% unless component.requested_by_chromium? %>
       <button
         aria-label="Full screen"
         class="btn sul-i-expand-1 sul-fullscreen-button"
         id="full-screen-button">
       </button>
+    <% end %>
   <% end %>
   <% component.with_body do %>
     <div class="sul-embed-body" style="width: 100%">
--- a/spec/components/embed/pdf_component_spec.rb
+++ b/spec/components/embed/pdf_component_spec.rb
@@ -3,13 +3,34 @@
 require 'rails_helper'
 
 RSpec.describe Embed::PdfComponent, type: :component do
-  pending "add some examples to (or delete) #{__FILE__}"
+  include PurlFixtures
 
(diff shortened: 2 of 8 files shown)


spec/components/embed/pdf_component_spec.rb
```diff
@@ -3,13 +3,35 @@
...
+  let(:embed_request) { Embed::Request.new(url:) }
+  let(:viewer) { Embed::ViewerFactory.new(embed_request).viewer }
+
+  before do
+    stub_request(:get, 'https://purl.stanford.edu/sq929fn8035.json')
+      .to_return(status: 200, body: pdf_public_json, headers: {})
+  end
+
+  it 'renders the full screen button by default' do
+    # debugger
```
jcoyne, line 18: Cruft?

app/components/embed/pdf_component.html.erb
```diff
@@ -1,10 +1,13 @@
 <%= render Embed::CompanionWindowsComponent.new(viewer:, stimulus_controller: 'file-auth') do |component| %>
   <% component.with_header_button do %>
+    <%# chromium has a bug preventing the esc key from exiting full screen in some situations %>
```
jcoyne, line 3: would you mind linking to https://issues.chromium.org/issues/40131988 here.

jcoyne (changes requested):

spec/components/embed/pdf_component_spec.rb
jmartin-sul: yup, thanks!

app/components/embed/pdf_component.html.erb
jmartin-sul: good call, will do

jcoyne (approved):

PR #2185, opened 2024-07-02
Use console.debug rather than console.log for debugging

jcoyne:


app/javascript/controllers/file_auth_controller.js
```diff
@@ -1,73 +1,92 @@
...
+          return contentResource;
+        });
+      });
+    });
+    this.firstFile = resources[0]["id"];
   }
 
   // Try to render the resource, checks for any required authorization and shows login window if needed
   maybeDrawContentResource(contentResource) {
-    console.log("Now figure out if we can render", contentResource)
```
corylown, line 47: Did you intend to remove this or should it change to `console.debug` like the others?

jcoyne: Yes, I fixed this.

corylown (approved):

PR #2191, opened 2024-07-09
add in leaflet opacity slider, update code to accomodate geojson

### LINKED ISSUE #254: Add opacity slider for geo public objects

dnoneill:
This PR:

1. Moves var Module = (function()  to export default so we can import leaflet_opacity (when I tried with var Module I was unable to import files
2. Updates the opacity on the WMS layer to 75%
3. adds and opacity slider.

**When looking at the file changes, turning on hide whitespace will make the changes more clear. I had to fix the tabbing with change 1**

Closes #254 
Screenshot 2024-07-09 at 1 24 08 PM
Screenshot 2024-07-09 at 1 23 59 PM

### DIFF
--- a/app/assets/stylesheets/geo.scss
+++ b/app/assets/stylesheets/geo.scss
@@ -67,3 +67,92 @@
     // scss-lint:enable SelectorDepth
   }
 }
+
+.opacity-control {
+  background-color: #a9acb1;
+  border-radius: 15px;
+  color: black;
+  font: bold 18px 'Lucida Console', Monaco, monospace;
+  display: block;
+  height: 200px;
+  left: 11px;
+  position: relative;
+  top: 15px;
+  width: 5px;
+
+  .opacity-handle {
+    background-color: #fff;
+    border-radius: 4px;
+    border: 1px solid #eee;
+    cursor: ns-resize;
+    font-size: 10px;
+    height: 26px;
+    left: -11px;
+    line-height: 26px;
+    position: absolute;
+    text-align: center;
+    top: 0;
+    width: 26px;
+    z-index: 20;
+    box-shadow: 0 1px 5px rgba(0,0,0,0.65);
+
+    &:hover {
+      background-color: $color-fog-light;
+    }
+  }
+
+  .opacity-arrow-up {
+    color: #aaa;
+    position: absolute;
+    top: -11px;
+    text-align: center;
+    width: 100%;
+
+    &:before {
+      content: '=';
+    }
+  }
+
+  .opacity-arrow-down {
+    bottom: -10px;
+    color: #aaa;
+    position: absolute;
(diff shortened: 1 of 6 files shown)


app/javascript/src/modules/leaflet_opacity.js
```diff
@@ -0,0 +1,80 @@
+// Adopts the Mapbox opacity control into a Leaflet plugin
+import L from "leaflet";
+
+!function(global) {
```
jcoyne, line 4: Why do we have the negation operator here?

jcoyne (commented):
Is it possible to control the slider with keyboard?

dnoneill:
Doesn't look like it. Nothing in the leaflet canvas seems to be accessible via keyboard.

dnoneill:
@jcoyne added code to allow a user to move the slider with the tab button (to get to the element) and up and down keys to control it.

app/assets/stylesheets/geo.scss
```diff
@@ -67,3 +67,92 @@
...
+    line-height: 26px;
+    position: absolute;
+    text-align: center;
+    top: 0;
+    width: 26px;
+    z-index: 20;
+    box-shadow: 0 1px 5px rgba(0,0,0,0.65);
+
+    &:hover {
+      background-color: #f4f4f4;
```
jcoyne, line 100: Can you use `var(--stanford-fog-light)` here instead? You might need to add this to https://github.com/sul-dlss/sul-embed/blob/main/app/assets/stylesheets/palette.scss

dnoneill: Done. I also updated it everywhere we used the color (1 other place).

app/assets/stylesheets/sul_variables.scss
```diff
@@ -46,4 +46,5 @@ $sul-link-hover-color: #00548f;
 $sul-metadata-bg: $color-beige-05;
 $sul-metadata-heading-color: $gray-54-percent;
 $sul-metadata-label-color: dimgrey;
-$sul-restricted-text-color: $color-cardinal-red;
\ No newline at end of file
+$sul-restricted-text-color: $color-cardinal-red;
+$sul-fog-light: #f4f4f4;
```
jcoyne, line 50: I think this should be `stanford-fog-light` as the palette is defined by the university, not the library.

jcoyne: Can it go into palette.scss?

dnoneill: Yep. I renamed it color-fog-light in pallette because that conforms with what is in there. I think if we want to rename everything stanford we can in a separate PR but I would like to keep things in line for this one.

jcoyne (approved):

PR #2208, opened 2024-08-07
Version file links

### LINKED ISSUE #2206: Download link should be the versioned file in stacks
### Expected behavior
When you go here: https://sul-purl-stage.stanford.edu/hs471kp0346/v1?version_feature=1

The "Download" link for the file should be the versioned download
Screenshot 2024-08-05 at 2 58 29 PM

### Actual behavior

It's not versioned.

lwrubel:
Resolves #2206 to version file download links (including Download all) when the purl is versioned. 

Tested on stage.

### DIFF
--- a/app/components/embed/file_row_component.rb
+++ b/app/components/embed/file_row_component.rb
@@ -12,7 +12,7 @@ module Embed
 
     attr_reader :viewer, :file, :pos_in_set, :set_size, :level
 
-    delegate :file_url, :label, to: :file
+    delegate :label, to: :file
 
     def file_size_text
       viewer.file_size_text(file.size)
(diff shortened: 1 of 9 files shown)


app/components/embed/file_row_component.rb
```diff
@@ -34,6 +34,14 @@ def download?
       !viewer.embargoed_to_world?(file)
     end
 
+    def version
+      viewer.purl_object.version_id
+    end
+
+    def url
+      file.file_url(download: false, version:)
```
lwrubel, line 42: I think something got left out of my PR, so looking at this.

aaron-collier:
See: https://github.com/sul-dlss/stacks/blob/main/config/routes.rb#L5

jcoyne (approved):

PR #2214, opened 2024-08-08
fix point not having icon

dnoneill:
When you click on a section here: https://purl.stanford.edu/rn106gn9916 you will get a broken image because leaflet uses map marker icons by default for geojson points. This PR updates the point to use a circle.

Screenshot 2024-08-08 at 2 07 28 PM


marlo-longley (approved):

peetucket (approved):

dbranchini:
This looks okay to me

PR #2221, opened 2024-08-09
move basemap into settings

dnoneill:


app/viewers/embed/viewer/geo.rb
```diff
@@ -21,7 +21,7 @@ def map_element_options
           options['data-layers'] = "druid:#{@purl_object.druid}"
         end
         options['data-index-map'] = index_map.file_url if index_map?
-        options['data-geo-viewer-colors'] = Settings.geo_viewer_colors.to_json
+        options['data-geo-viewer'] = Settings.geo_viewer.to_json
```
jcoyne, line 24: Is there are reason to put these settings in yaml?  Are we planning on making them different in different environments?

dnoneill: We are planning on updating the basemap to something different for our earthworks work. I updated because it seems cleaner to not have to change it in the code. I don't think we are planning on having this be different unless other POs have a problem with the new basemap.

dnoneill: @jcoyne Should I just update the map in the code?

jcoyne: I prefer that as you don't have to go hunting around to find it.

PR #2225, opened 2024-08-12
fix pan for when there is no geometry

dnoneill:
see https://purl.stanford.edu/tg926kp6619

### DIFF
--- a/app/javascript/src/modules/geo_viewer.js
+++ b/app/javascript/src/modules/geo_viewer.js
@@ -200,7 +200,15 @@ export default {
       layer = L.geoJSON(data, {color: color, opacity: opacity})
       geoJSON = true;
     }
-    this.map.setView(layer.getBounds().getCenter());
+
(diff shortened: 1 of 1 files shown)


app/javascript/src/modules/geo_viewer.js
```diff
@@ -200,7 +200,13 @@ export default {
       layer = L.geoJSON(data, {color: color, opacity: opacity})
       geoJSON = true;
     }
-    this.map.setView(layer.getBounds().getCenter());
+
+    try {
```
jcoyne, line 204: Would you mind adding a comment here telling the reader what we're trapping for and how an item would get into this situation?  Can you give an exemplar object so that people can test this path?

jcoyne (changes requested):

app/javascript/src/modules/geo_viewer.js
dnoneill: Yup. done.

jcoyne (approved):

PR #2260, opened 2024-11-20
fix captions on load

### LINKED ISSUE #2258: Captions are off-center if they appear on the screen prior to playback
### Expected behavior

Captions should use the same text alignment whenever they are displayed on screen.

### Actual behavior

If a video starts with the captions already displayed, the text is pushed to the right. It then aligns properly after the next set of text gets loaded.

https://github.com/user-attachments/assets/c937a1ba-2c93-4ada-9987-52bfc94199ce

Go to [this item](https://sul-purl-stage.stanford.edu/vh170nm3897) and then play it back.

### Browser / Environment
Chrome, Firefox, Edge on Mac. Safari is, as usual, weird in its own way.

dnoneill:
closes #2258

timeupdate in loadedmetadata was causing this issue.

### DIFF
--- a/app/javascript/controllers/media_player_controller.js
+++ b/app/javascript/controllers/media_player_controller.js
@@ -8,18 +8,13 @@ export default class extends Controller {
     this.player.on('loadedmetadata', () => {
       const event = new CustomEvent('media-loaded', { detail: this.player })
       window.dispatchEvent(event)
+    })
 
-      // Stop the `loadedmetadata` event and don't bother listening for
-      // `timeupdate` events until after `loadedmetadata` fires completes.
-      // Why? If we don't, `timeupdate` will be triggered before the video
-      // (and its captions) are fully present, and we use the custom
-      // `time-update` event to scroll the transcript panel.
-      this.player.off('loadedmetadata')
-      this.player.on('timeupdate', () => {
-        const timestamp = this.player.currentTime()
-        const event = new CustomEvent('time-update', { detail: timestamp })
-        window.dispatchEvent(event)
-      })
+    // occurs when time of the video changes.
+    this.player.on('timeupdate', () => {
+      const timestamp = this.player.currentTime()
+      const event = new CustomEvent('time-update', { detail: timestamp })
+      window.dispatchEvent(event)
     })
 
     // The loadeddata event occurs when the first frame of the video is available, and 


app/javascript/controllers/media_player_controller.js
```diff
@@ -20,6 +20,7 @@ export default class extends Controller {
         const event = new CustomEvent('time-update', { detail: timestamp })
         window.dispatchEvent(event)
       })
+      this.player.textTrackDisplay.updateDisplay();
```
jcoyne, line 23: Can you add a comment here about why we need this?

jcoyne (dismissed):

app/javascript/controllers/media_player_controller.js
```diff
@@ -20,6 +20,10 @@ export default class extends Controller {
         const event = new CustomEvent('time-update', { detail: timestamp })
         window.dispatchEvent(event)
       })
+
+      // patch for https://github.com/sul-dlss/sul-embed/issues/2258
+      // TODO: check we can remove after videojs is updated to 8.19.2.
```
jcoyne, line 25: Can we upgrade to 8.20.0 now and then drop this? https://github.com/videojs/video.js/commit/62f38446a5f01e85739bdb7207c7e992b57c1c86

dnoneill: I should update the todo. The upgrade did not fix the issue. I am not quite sure what is going on but it seems to be specific to our sul-embed instance.

dnoneill:
@jcoyne This bug is being caused by https://github.com/sul-dlss/sul-embed/blob/99ac3d3e471f30518e4e28bdd13afc31e7ffa4d1/app/javascript/controllers/media_player_controller.js#L17-L23

Do you know why you put a listener for timeupdate in another listener?

app/javascript/controllers/media_player_controller.js
```diff
@@ -20,6 +20,12 @@ export default class extends Controller {
         const event = new CustomEvent('time-update', { detail: timestamp })
         window.dispatchEvent(event)
       })
+
+      // patch for https://github.com/sul-dlss/sul-embed/issues/2258
+      // For some reason loadedmetadata is being triggered before texttrackchange.
+      // https://github.com/videojs/video.js/blob/main/src/js/tracks/text-track-display.js#L127-L128
+      // This is not true in video.js. This patch will fix the problem.
```
jcoyne, line 27:
```suggestion
      // This behavior does not occur in plain video.js. Calling updateDisplay() explicitly fixes the problem in this app.
```

dnoneill: I am happy to update this comment. Do you have an answer to the question above. It looks like you are the one who implemented that listener and the comment isn't giving me any clarity because I am not quite sure how timeupdate could be available before loadedmetadata is done firing.

jcoyne: Wasn't it @mjgiarlo https://github.com/sul-dlss/sul-embed/commit/8c4dba7a0cbee3dcd25946c473cdd98af31999ba ?

dnoneill: right. i read the code wrong. thanks

dnoneill: @jcoyne I talked to Mike. It seems that the listener was for autoscroll and moving it out of timeupdate doesn't affect autoscroll and fixes the bug outlined in the issue. This PR is ready for review.

peetucket (approved):

PR #2267, opened 2024-12-04
add auto-generated label to caption

### LINKED ISSUE #2266: Add auto-generated parenthetical to language selectors
For media items with auto-generated captions, add the `(auto-generated)` parenthetical to the language selector in both the media player view ("CC" selector) and the transcript pane ("Select language" dropdown).

Media Viewer - Caption Language Selector

**NB**: Adding this parenthetical will increase the width of the "Select language" dropdown in the transcript pane. Currently, this dropdown is in-line with the "Auto scroll" checkbox. To make room for the wider dropdown, move it below the checkbox.

dnoneill:
closes #2266 
Screenshot 2024-12-05 at 5 01 24 PM

Screenshot 2024-12-05 at 12 58 50 PM

### DIFF
--- a/app/assets/stylesheets/companion_window.scss
+++ b/app/assets/stylesheets/companion_window.scss
@@ -234,6 +234,11 @@
     }
   }
 
+  .transcript-header {
+    background: $color-fog-light;
+    padding-bottom: 1.5rem;
+  }
+
   .header-background {
     background: $color-fog-light;
     display: flex;
@@ -253,6 +258,7 @@
 
     &.transcript-controls {
       font-size: 0.8rem;
+      flex-direction: column;
 
       input {
         margin-left: 0.6rem;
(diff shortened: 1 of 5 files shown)


alundgard:
Could we move the dropdown language selector to its own line? (described in the original issue)
Also would it be possible to not wrap the label in the CC selector? Maybe that's a video.js constraint.

dnoneill:
Oops. Missed that. Updated. And in answer to `Also would it be possible to not wrap the label in the CC selector? Maybe that's a video.js constraint.`. Not in a good way. I could make it not wrap but it is going to look weird with captions without auto-generated. We would basically need to add JS to fix this and it will be weird.

dnoneill:
I was able to test this with a json locally that didn't contain sdrGeneratedText. It worked so I am more comfortable merging this.

peetucket (approved):

PR #2268, opened 2024-12-04
fix transcript sidebar for multiple media objects

### LINKED ISSUE #2257: In the media viewer, the transcript in the left companion window should correspond to the currently selected video
### Expected behavior

When viewing a captioned media item, the transcript displayed in the left companion window should correspond to the currently selected media item. If there is only one set of captions for a media recording, then there should be only one transcription option to display.

### Actual behavior

When an item has two or more captioned media files (multiple video and/or audio files), the transcript view shows a language dropdown that gives access to all the transcriptions.

In the screenshot below, the two "English" selections correspond to different files.

Screenshot 2024-11-18 at 12 56 30 PM

Additionally, it appears that the highlighted words in the auto-scroll correspond to the first listed transcript even when a different file is being played back.

The screenshot below shows the transcription from the first audio recording in the item while the captions in the viewer show the words actually being spoken in the recording being played back - the second file in that item.

Screenshot 2024-11-18 at 12 57 34 PM

Beyond that, it looks like in some circumstances, choosing the different "English" options will change the transcript being displayed, and sometimes nothing changes when you choose English. It seems like you can choose either transcription right after loading the viewer, but at some point after interacting with the files, you can no longer change the transcript until you reload again.

This item in stage has two audio files and can be used for testing: https://sul-purl-stage.stanford.edu/gj640kc3831

Both screenshots above are from this item.

### Browser / Environment

Firefox, Chrome, Edge, and Safari on Mac OS all have trouble displaying transcripts when an item has multiple recordings. They don't all do the same thing, with Safari maybe being the most wrong.

I think targeting Chrome/Edge and Firefox will likely cover most desktop users, followed by Safari. It's possible that fixing the issue for one browser may also improve (or at least change) things for others, so I'm holding off on describing each browser's behavior in detail.

dnoneill:
closes #2257

### DIFF
--- a/app/javascript/controllers/media_wrapper_controller.js
+++ b/app/javascript/controllers/media_wrapper_controller.js
@@ -11,13 +11,18 @@ export default class extends Controller {
   toggleVisibility(event) {
     const index =  event.detail.index
     this.element.hidden = this.indexValue !== index
-    this.pauseAllMedia()
+    this.pauseAllMedia(index)
   }
 
-  pauseAllMedia() {
+  // switch transcript if media object index is the same as the visible index.
+  pauseAllMedia(index) {
     const mediaObject = this.element.querySelector('.video-js')
     if (mediaObject) {
-      videojs(mediaObject.id).pause()
+      const playerObject = videojs(mediaObject.id);
+      playerObject.pause()
+      if (mediaObject.dataset.index == index){
+        window.dispatchEvent(new CustomEvent('switch-transcript', { detail: playerObject }))
+      }
     }
   }
 }
--- a/app/components/embed/media_component.html.erb
+++ b/app/components/embed/media_component.html.erb
@@ -30,7 +30,7 @@
           </div>
         </div>
       </div>
-      <div class="left-drawer-content transcript" data-action="media-loaded@window->transcript#persistPlayer media-data-loaded@window->transcript#load time-update@window->transcript#highlightCue">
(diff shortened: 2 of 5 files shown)


app/javascript/controllers/media_wrapper_controller.js
```diff
@@ -11,13 +11,17 @@ export default class extends Controller {
   toggleVisibility(event) {
     const index =  event.detail.index
     this.element.hidden = this.indexValue !== index
-    this.pauseAllMedia();
+    this.pauseAllMedia(index);
   }
 
-  pauseAllMedia() {
+  pauseAllMedia(index) {
```
jcoyne, line 17: Can you document the purpose of the index parameter?

dnoneill: Yep. Done.

peetucket (approved):

peetucket:
@jcoyne any other concerns about this (after the conflict is fixed)?

jcoyne:
@peetucket I haven't had the bandwidth to look into this.

peetucket:
> @peetucket I haven't had the bandwidth to look into this.

Thanks and no problems, just wanted to give you an opportunity to weigh in (since you previously reviewed) before we moved forward.

dnoneill:
@peetucket hold off on merging

dnoneill:
Checked multi-lingual. Fixes the problem there and is working correctly. merging

PR #2269, opened 2024-12-05
Prefer const over var

jcoyne:


app/javascript/controllers/media_tag_controller.js
```diff
@@ -50,15 +50,15 @@ export default class extends Controller {
...
+
     window.dispatchEvent(new CustomEvent('thumbnails-found', { detail: thumbnails }))
   }
 
   // Open the login window in a new window and then poll to see if the auth credentials are now active.
   logIn(event) {
     const windowReference = window.open(event.params.loginService);
     this.loginStart = Date.now();
-    var checkWindow = setInterval(() => {
+    const checkWindow = setInterval(() => {
```
dnoneill, line 61: Can we update this to match? https://github.com/sul-dlss/sul-embed/blob/99ac3d3e471f30518e4e28bdd13afc31e7ffa4d1/app/javascript/controllers/file_auth_controller.js#L231

dnoneill: Also I am more inclined to let than const. I am worried about a weird edge case that probably won't ever happen.

jcoyne: I'm not sure I understand the case where checkWindow would ever be reused for a second value.  I think `const` is better when appropriate as `let` tells the reader that you expect it to be used more than one time and it can be confusing if they can't find that other time.

thatbudakguy: I also can't think of any possible way that this would get redeclared. Imo `const` is appropriate almost everywhere; `let` is a signal that you intend to mutate later and if that isn't the case it shouldn't be used

thatbudakguy (approved):

PR #2275, opened 2024-12-05
don't display auto-generated captions by default

### LINKED ISSUE #2262: Do not display uncorrected captions by default
We currently enable caption display by default when the viewer loads. If we can detect when captions are auto-generated and not corrected, we should not display those by default. The user can still enable them by choice. 

We should be able to rely on Cocina to detect auto-generated and uncorrected captions. The values will be:

- sdrGenerated: true
- correctedForAccessibility: false

**Additional information**

This ticket will match up with a future ticket to label when captions are auto-generated and uncorrected so that users know what they are viewing.

dnoneill:
closes #2262

### DIFF
--- a/app/components/embed/media_tag_component.rb
+++ b/app/components/embed/media_tag_component.rb
@@ -98,7 +98,7 @@ module Embed
         @resource.caption_files.map.with_index do |caption_file, i|
           tag.track(src: caption_file.file_url, kind: 'captions',
                     srclang: caption_file.language_code, label: caption_file.media_caption_label,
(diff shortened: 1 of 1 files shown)


peetucket:
LGTM.  There are conflicts now, may be worth re-basing to avoid even more divergence.

dnoneill:
Not really ready for review. https://github.com/sul-dlss/sul-embed/pull/2267 needs to be merged first

dnoneill:
@peetucket ready for review

peetucket (approved):

PR #2278, opened 2024-12-09
fix sidebar load for cached pages

### LINKED ISSUE #2175: Media viewer left companion window sometimes opens without showing any text or images
### Expected behavior

The media viewer should open to show a list of the media contents.

https://github.com/sul-dlss/sul-embed/assets/8975583/7965546e-4cd9-486d-bc94-c3c1ea61130c

### Actual behavior
Sometimes it will open to show a blank panel instead of the contents list.

https://github.com/sul-dlss/sul-embed/assets/8975583/96617955-392d-4edc-8647-38e2ca4bb03b

I haven't been able to reproduce this with a reliable set of steps. Instead, it happens every now and then when I open videos. I've also heard it reported by users. To get the screen recording above, I went through the canonical list of media items and opened each one until I saw the behavior.

I'm pretty sure I saw this issue before we started auto-expanding the contents list, but it would only happen if you expanded the sidebar almost immediately after loading the viewer.

### Browser / Environment

Firefox and Chrome on a Mac.

dnoneill:
closes #2175

### DIFF
--- a/app/javascript/controllers/media_tag_controller.js
+++ b/app/javascript/controllers/media_tag_controller.js
@@ -50,7 +50,11 @@ export default class extends Controller {
                  fileLabel: dataset.fileLabel || '' }
       })
 
-    window.dispatchEvent(new CustomEvent('thumbnails-found', { detail: thumbnails }))
+    // Timeout is set because when the page is cached, the event fires before the content_list_controller is mounted.
+    // This causes the sidebar not to load: https://github.com/sul-dlss/sul-embed/issues/2175
+    setTimeout(() => {
+      window.dispatchEvent(new CustomEvent('thumbnails-found', { detail: thumbnails }))
+    }, "100");
   }
 
   // Open the login window in a new window and then poll to see if the auth credentials are now active.


app/javascript/controllers/media_tag_controller.js
```diff
@@ -50,7 +50,11 @@ export default class extends Controller {
                  fileLabel: dataset.fileLabel || '' }
       })
 
-    window.dispatchEvent(new CustomEvent('thumbnails-found', { detail: thumbnails }))
+    // Timeout is set because when the page is cached, the event fires before the content_list_controller is mounted.
```
taylor-steve, line 53: Is `findThumbnails` idempotent? Could you dispatch a custom event in `content_list_controller#connect` and listen for it in `media_tag_controller` and run `findThumbnails` again? If you've been down this road already, I certainly won't stand in the way of progress.

dnoneill: I have. It ends up being stuck in an endless loop. Also sometimes it doesn't work because the media_tag_controller loads after the content_list_controller and so it gets really funky.

taylor-steve: I believe it. Thanks!

peetucket (approved):

PR #2279, opened 2024-12-11
sul-embed work for exhibits select image section

cbeer:


jcoyne:
It looks like this needs a rebase.

corylown (approved):

PR #2285, opened 2025-01-10
Use new logo without rosette

jcoyne:


dnoneill (dismissed):
I am getting an error, `The asset 'sul_stacked_color.svg' was not found in the load path.` when running locally. Also maybe this is something we want to add the component library so we can point both this and stacks to this URL? It seems like the current version is brittle and could use updating.

jcoyne:
@dnoneill I don't think this should be in the component library, this is the only app that uses this stacked svg.

dnoneill:
@jcoyne It is being used in every manifest for the logo field: https://purl.stanford.edu/bg262qk2288/iiif/manifest

jcoyne:
Stacks does auth for every asset. This image doesn't require auth. The svg is a smaller download than a jpg.

corylown (approved):

PR #2286, opened 2025-01-10
refactor download for items with captions/transcripts

dnoneill:


app/components/embed/companion_windows_component.html.erb
```diff
@@ -90,7 +90,8 @@
           Files
         </h3>
         <ul>
-          <%= render Embed::Download::FileListItemComponent.with_collection(downloadable_files) %>
+          <% show_headers = downloadable_files.any?{|file|file.caption? || file.transcript? } %>
```
jcoyne, line 93: Please move this code out of the template and into the class.

app/components/embed/download/file_list_item_component.html.erb
```diff
@@ -1,3 +1,8 @@
+<% if show_headers %>
```
jcoyne, line 1: Can we combine these two conditionals into one?  Can they be put into a method on the class (with documentation about why we are showing this and in what situations)?

jcoyne (changes requested):
Can you add a test for when headers are shown on the CompanionWindowComponent?

dnoneill:
test added, other comments implemented

jcoyne (approved):

PR #2294, opened 2025-01-23
fix for mixed language media with transcripts

### LINKED ISSUE #2293: second language video switching
When choosing a second language on a video, it will stay in the transcript sidebar until you choose a new language. https://sul-purl-stage.stanford.edu/ws924ny0572

https://github.com/user-attachments/assets/a65470ae-b166-4662-85b5-419251fc51ab

dnoneill:
closes #2293

### DIFF
--- a/app/javascript/controllers/transcript_controller.js
+++ b/app/javascript/controllers/transcript_controller.js
@@ -89,8 +89,11 @@ export default class extends Controller {
     // changing disabled mode to "hidden" to prevent losing cue information.
     if(videojs.browser.IS_ANY_SAFARI) {
       this.convertDisabledTracks()
+      // the function this.trackCues returns the wrong number of captions when in Safari
+      // this has something to do with loadedmetadata triggering before all the captions have loaded.
+      // This is a temporary fix. We really should figure out a way to update this function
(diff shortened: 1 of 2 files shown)


dnoneill:
This doesn't work on safari. I am working on it.

app/javascript/controllers/transcript_controller.js
```diff
@@ -89,6 +89,8 @@ export default class extends Controller {
     // changing disabled mode to "hidden" to prevent losing cue information.
     if(videojs.browser.IS_ANY_SAFARI) {
       this.convertDisabledTracks()
+      // track.cues returns 0 in safari, even when there are cues
```
jcoyne, line 92: I'm not sure why this is relevant.  Is there more to context you can pass on here?

dnoneill: Sure. I can update the comment, the long and short is if there are two vtt files with captions, return captions.filter(track => this.trackCues(track).length) returns a list of one file. This has to do with something weird in safari that returns track.cues.length == 0, on hidden tracks.

corylown (approved):

PR #2322, opened 2025-01-31
Include Source Sans 3

jcoyne:
Previously we were using Source Sans Pro, which was not included in the styles

### DIFF
--- a/app/assets/stylesheets/common.scss
+++ b/app/assets/stylesheets/common.scss
@@ -3,6 +3,7 @@
 @import "modules/buttons";
 @import "modules/popup_panels";
 @import "modules/utility";
+@import "https://fonts.googleapis.com/css2?family=Source+Sans+3:ital,wght@0,200..900;1,200..900&display=swap";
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/common.scss
```diff
@@ -5,6 +5,8 @@
 @import "modules/popup_panels";
 @import "modules/utility";
 
+@import "https://fonts.googleapis.com/css2?family=Source+Sans+3:ital,wght@0,200..900;1,200..900&display=swap";
```
thatbudakguy, line 8: Do we use all these weights? Might be able to save a few kB if it's only 300,400,700 which are more common

jcoyne: I was using the same URL as in the component-library, so that the user will have a cached version.

thatbudakguy: Interesting. I guess if we get people to cache them all across our sites, it'll matter less

thatbudakguy (approved):

PR #2332, opened 2025-01-31
Use i18n for text differences between viewer types

jcoyne:
because i18n is the built in solution for text labels, which we are already using extensively.  We don't need to maintain a separate way of doing this in the class.  Additionally, it's will become even more important as we implement new viewers (see https://github.com/sul-dlss/sul-embed/pull/2331).

### DIFF
--- a/app/components/companion_windows_component.html.erb
+++ b/app/components/companion_windows_component.html.erb
@@ -43,7 +43,7 @@
 
         <% if render_content_list_panel? %>
           <section id="object-content" role="tabpanel">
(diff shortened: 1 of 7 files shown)


dnoneill:
@jcoyne Can you put a quick note about why this change is being implemented?

jcoyne:
@dnoneill because i18n is the built in solution for text labels, which we are already using extensively.  We don't need to maintain a separate way of doing this in the class.  Additionally, it's will become even more important as we implement new viewers (see https://github.com/sul-dlss/sul-embed/pull/2331).

PR #2333, opened 2025-01-31
Update to Mirador 4 alpha

cbeer:


app/javascript/src/plugins/shareMenuPlugin.jsx
```diff
@@ -24,18 +21,7 @@ const WindowTopBarShareMenu = (props) => (
   />
 )
 
-/**
- *
- * @param theme
- * @returns {{ctrlBtn: {margin: (number|string)}}}
- */
-const styles = theme => ({
```
cbeer, line 32: Find out if we need this style...

jcoyne:
🎉

app/javascript/src/plugins/shareMenuPlugin.jsx
marlo-longley: This style can provide a class to Mirador here: https://github.com/ProjectMirador/mirador/blob/mui6/src/components/WindowTopMenuButton.js#L34

Visually, this button's background when selected looks to match the other buttons without this class.
So it seems on the surface that it isn't needed. 

Screenshot 2025-02-12 at 5 40 15 PM
Screenshot 2025-02-12 at 5 40 25 PM

Issue #2255, opened 2024-11-11
Enable captions for items where the captions have the rights: "view: stanford", "download: stanford


andrewjbtw:
Currently, we display captions only if the rights to the caption file are 

- "view: world" and 
- "download: world"

Can we show the captions if the caption file is:

- "view: stanford" and
- "download: stanford"

My understanding is that "download: none" will not work, but I'm unclear on "download: stanford".

This is related to #2138 and #2145.

jcoyne:
What should the UX be for this situation?

dnoneill:
The same as when the item is not stanford-only. Basically if the item is stanford only captions will not load in the viewer and the transcript sidebar also doesn't load.

andrewjbtw:
> What should the UX be for this situation?

We have an existing login flow for Stanford-Stanford items. What happens now is that after logging in, the user sees the caption selector on the lower right but the captions never get displayed. This ticket is to have them display, following the same UX for when an item has world-world rights.

jcoyne:
Isn't the solution here to put download: world on the caption file?

dnoneill:
https://purl.stanford.edu/nm886yc3832 is an example of this.

jcoyne:
Currently [this](https://purl.stanford.edu/nm886yc3832 causes a "Not same-origin" exception in Firefox when loading https://stacks.stanford.edu/file/druid:nm886yc3832/nm886yc3832_GMT20240306-221001_Recording_cap.vtt

dnoneill:
@jcoyne https://sul-purl-stage.stanford.edu/bg043zg7219

andrewjbtw:
I've fixed up https://sul-purl-stage.stanford.edu/wv603wp3744 so that it correctly has just one caption file and the rights are Stanford/Stanford on all the files. So it should match the prod example now.

dnoneill:
Thank you

PR #2345, opened 2025-02-03
Set crossorigin=use-credentials on video tag

jcoyne:
This should hopefully allow restricted VTT to be downloaded

Ref #2255


dnoneill (approved):

dnoneill:
tested on stage. works.

Issue #2347, opened 2025-02-04
File viewer redesign


alundgard:
_Implement file viewer redesign._

Figma design: https://www.figma.com/design/wFeyaNc10k8rSaPuZFK3F9/SUL-Embed-%2F-Viewers-2025?node-id=17-1801&t=8LOvLzHEaZdya1JY-4.

### Left companion window - Closed
![Image](https://github.com/user-attachments/assets/7ab7d076-ea93-4c11-8e7b-7b7cc54414b3)

### Left companion window - Open
![Image](https://github.com/user-attachments/assets/18053582-bab9-49d2-b4ff-d8146eb1b509)

### Menus
<img width=100 src="https://github.com/user-attachments/assets/143d3fc8-6a14-4b21-962e-b8902afeb6f9">
<img width=300 src="https://github.com/user-attachments/assets/a28891cf-e9a0-490f-a9aa-1ce1f3589521">

jcoyne:
@alundgard what is this for? <img width="31" alt="Image" src="https://github.com/user-attachments/assets/c3c7db96-5fdd-43f8-9116-69ec6a7d0915" />

dnoneill:
closing and creating new tickets for items outlined in pr

PR #2348, opened 2025-02-04
Add new file component

jcoyne:
TODO (as separate PRs):
- [ ] Add search and file count
- [ ] Add download component that only shows "Download all" (blocked by #2349)
- [ ] Add full screen.

Ref #2347

### DIFF
--- a/app/assets/stylesheets/common.css
+++ b/app/assets/stylesheets/common.css
@@ -1,0 +1,23 @@
+/* From Bootstrap 5.3 */
+.visually-hidden-focusable:not(:focus):not(:focus-within):not(caption),
+.visually-hidden:not(caption) {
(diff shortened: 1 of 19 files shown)


jcoyne:
@dnoneill did you `rm` all the old styles in app/assets/builds/ ?

dnoneill:
@jcoyne. Yup that was it. Thank you.

dnoneill (approved):
I am approving but I want to note a couple things
1. hide_title=true doesn't seem to work
2. There is a weird thing that iframe changes size based on the sidebar. This might not be an issue when it is embedded, but I am noting it.

https://github.com/user-attachments/assets/cb134096-937f-457d-ab2b-e17133ed8d52

PR #2363, opened 2025-02-05
Move Login component to the Document namespace

jcoyne:
This makes it clear it's not a shared component


dnoneill:
Why Document and not PDF or something along those lines? I know Document is the type in Purl but we use PDF more in the code

jcoyne:
I think we should move toward the object type names to be consistent.  We don't currently have a "Mp4" viewer, it's a "media viewer".  We don't have a Jp2 viewer, we have a "image viewer".

dnoneill:
Yeah but we don't have a document_viewer, we have a pdf_viewer. Maybe this should be a move to everything to document viewer. I guess my worry is it isn't consistent if there is only one thing that is named document.

jcoyne:
@dnoneill I think that's a great idea as a follow up.  But I don't want this PR to be large. I just want to show that this piece is not a shared component.

dnoneill:
@jcoyne Okay but shouldn't it be consistent with what we currently have and not introduce an entirely new namespace we don't use anywhere else? I don't have a problem with this being small, if it is I think it should be PDF and then we can migrate to document.

jcoyne:
@dnoneill we do have https://github.com/sul-dlss/sul-embed/blob/main/app/javascript/document.js already.

dnoneill (approved):

dnoneill:
@jcoyne Can we close this? It is going to be shared with 3d and probably media when it is refactored.

PR #2383, opened 2025-02-07
don't remove iframe if message is from untrusted source

### LINKED ISSUE #2370: When viewing the /iframe url for PDF there is a continous error scrolling
Image

dnoneill:
closes #2370

### DIFF
--- a/app/javascript/controllers/file_auth_controller.js
+++ b/app/javascript/controllers/file_auth_controller.js
@@ -9,12 +9,12 @@ export default class extends Controller {
   addPostCallbackListener() {
     const permittedOrigins = ["https://stacks.stanford.edu", "https://sul-stacks-stage.stanford.edu", "https://sul-stacks-uat.stanford.edu"]
     window.addEventListener("message", (event) => {
-      this.iframe.remove()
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/file_auth_controller.js
```diff
@@ -8,7 +8,7 @@ export default class extends Controller {
 
   addPostCallbackListener() {
     window.addEventListener("message", (event) => {
-      this.iframe.remove()
+      if (this.iframe) this.iframe.remove()
```
jcoyne, line 11: Where would the callback come from if not in the iframe?

dnoneill: This is really for locally development. If you go to the iframe URL it gives this error.

jcoyne: can you show me? The iframe I'm referring to is the one created here: https://github.com/sul-dlss/sul-embed/blob/main/app/javascript/controllers/file_auth_controller.js#L206

marlo-longley (approved):

jcoyne (approved):

PR #2387, opened 2025-02-07
Add fullscreen for File viewer

### LINKED ISSUE #2380: Add full screen toggle to new file viewer
See #2347 for design.  Note this should be suppressed on Safari iOS as it doesn't support fullscreen API.

jcoyne:
Fixes #2380


dnoneill (approved):
I am going to approve this but I am leaving a note about how this might not be how we want fullscreen to work, because the only way out of fullscreen is esc. It would probably be good to have the header, sidebar, etc. go fullscreen.

jcoyne:
I leave it up to @alundgard to decide, but that seems unlike any of the other viewers, which only show the current content panel in full screen.

PR #2388, opened 2025-02-07
Rename PdfViewer to DocumentViewer

jcoyne:
All other viewers use a more abstract name, not focused on implementation details


dnoneill:
Are we keeping PdfComponent as PdfComponent and not DocumentComponent?

jcoyne:
@dnoneill I saw that as the boundary, where we're actually picking a specific implementation, e.g. linking to a PDF file.

corylown (approved):

PR #2390, opened 2025-02-07
Add download all link for new file viewer

### LINKED ISSUE #2381: Add download component that only shows "Download all"
Currently blocked by https://github.com/sul-dlss/sul-embed/pull/2349

jcoyne:
Fixes #2381


dnoneill:
I don't think it's a blocker but the file size is for the unzipped files and not the zipped files so I wonder if we should put something to tell the user that the file size is unzipped. @alundgard 
Screenshot 2025-02-10 at 1 13 42 PM
Screenshot 2025-02-10 at 1 12 24 PM

dnoneill (approved):

Issue #2401, opened 2025-02-10
Geo viewer redesign


alundgard:
_Implement geo viewer redesign._

Figma design: https://www.figma.com/design/wFeyaNc10k8rSaPuZFK3F9/SUL-Embed-%2F-Viewers-2025?node-id=1-1261.

### Left companion window - Closed
![Image](https://github.com/user-attachments/assets/88fab1bc-d570-464b-9e64-8c04331d21cd)

### Left companion window - Open
![Image](https://github.com/user-attachments/assets/027c27e4-0279-431c-8ad6-1e47c3fd5ceb)

### Dropdown menu - Open
![Image](https://github.com/user-attachments/assets/a87ebffc-395d-4a3a-ae4d-c5126390dd49)

### Download modal
![Image](https://github.com/user-attachments/assets/eca0fd0d-080d-4315-a858-dda4f046c640)

PR #2411, opened 2025-02-10
Create geo component using companion window layout

jcoyne:


dnoneill:
Feature sidebar is grey and there is no highlighting on clicked element



Screenshot 2025-02-11 at 11 15 39 AM

dnoneill:
Actually there is highlighting. Something about the wms map is being added wrong because it is highlighting the wrong area based on clicking

dnoneill:
With nothing clicked Features should include click the map to inspect features
Screenshot 2025-02-11 at 11 46 45 AM

dnoneill:
Also same with the modal?

dnoneill:
@jcoyne Could you put a comment about why we are going to a vendorized approach to leaflet instead of using npm?

jcoyne:
@dnoneill regarding vendorizing Leaflet. We've already vendorized the javascript, this just completes the process by vendorizing the css.  I had do to this as the new viewer doesn't use SASS and thus can't include CSS in `node_modules/`.

dnoneill:
[#2401](https://github.com/sul-dlss/sul-embed/issues/2401)

dnoneill (approved):

PR #2430, opened 2025-02-11
Enable hide-title for component window viewers

jcoyne:
This worked on the legacy viewers and Mirador, so this PR brings it to the new viewers.


dnoneill (changes requested):
How it should work
Screenshot 2025-02-13 at 6 02 08 PM

How it's working in new viewers
Screenshot 2025-02-13 at 6 03 17 PM

The fullscreen/etc icons are not being right aligned

jcoyne:
@dnoneill I've added a fix.

dnoneill (approved):

PR #2441, opened 2025-02-12
The geo download page should show filenames

### LINKED ISSUE #2429: Download modal geo viewer
There is a grouped_files? function in app/components/download/file_list_component.rb. I am guessing most of this would be updating the grouped_files? functionality. 🤞 That should be most of what needs to be done to group these files correctly

Design

Image
Image

jcoyne:
Fixes #2429

### DIFF
--- a/app/components/download/file_list_component.html.erb
+++ b/app/components/download/file_list_component.html.erb
@@ -2,19 +2,23 @@
   <h3 class='sul-embed-options-label modal-subheader'>
     Files
   </h3>
-  <% if grouped_files? %>
-    <% grouped_downloadable_files.each do |content| %>
-      <% if content.files.any? %>
-        <h4><%= content.label %></h4>
-        <ul>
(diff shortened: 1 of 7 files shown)


app/components/download/file_list_component.html.erb
```diff
@@ -2,19 +2,23 @@
...
-    <% grouped_downloadable_files.each do |content| %>
-      <% if content.files.any? %>
-        <h4><%= content.label %></h4>
-        <ul>
-          <%= render Download::FileListItemComponent.with_collection(content.files) %>
-        </ul>
+  <%= render Download::AllFilesComponent.new(viewer:) %>
+
+  <% if single_file_download? %>
+    <% if grouped_files? %>
```
dnoneill, line 8: Can we combine these two? 
```suggestion
    <% if single_file_download? && grouped_files?  %>
```

Issue #2361, opened 2025-02-05
Do not auto-expand the contents panel on mobile


alundgard:
Follow-up to #2165 that goes with #2342. If the user visits the embedded content from a mobile device, do not auto-expand the contents panel upon page load. If the user is on desktop, continue to auto-expand the panel.

dnoneill:
This was originally for the media-viewer but we want it on all viewers. https://github.com/sul-dlss/sul-embed/pull/2457#issuecomment-2654769468

dnoneill:
closed with https://github.com/sul-dlss/sul-embed/pull/2465

PR #2457, opened 2025-02-12
fix table overflowing out of viewer div

### LINKED ISSUE #2415: new file viewer is not responsive on medium and lower screens
https://github.com/user-attachments/assets/8e271114-9419-4868-8926-f607d5e6a83d

dnoneill:
This closes #2415  but it is incredibly cramped on small screens due to the sidebar. @alundgard Do we want to collapse the sidebar on smaller screens like we do with media? https://github.com/sul-dlss/sul-embed/issues/2361 or was this ticket for all viewers?

https://github.com/user-attachments/assets/ea1004ad-38ad-4985-a968-e6161212760a

### DIFF
--- a/app/assets/stylesheets/companion_window.css
+++ b/app/assets/stylesheets/companion_window.css
@@ -317,6 +317,7 @@
     #main-display {
       display: flex;
       flex: 1 1 auto;
+      min-width: 0;
     }
   }
 


alundgard:
Yes, I think for all viewers:we should not auto-expand the sidebar on mobile.

jcoyne (approved):

PR #2469, opened 2025-02-13
Use interactive element to improve a11y for geo

jcoyne:


jcoyne:
@corylown thank you for catching that.  I think this has now been resolved.

corylown (approved):

PR #2470, opened 2025-02-13
Set transform-origin in center of rotating caret

### LINKED ISSUE #2478: File viewer caret rotation/alignment
Probably don't want to do this if it requires a bunch of weirdness, but at least look into and if possible implement. Currently the carets aren't aligned properly.

Current
Image

Design
Image

jcoyne:
This prevents the rotation from messing with the row padding.

closes #2478

### DIFF
--- a/app/assets/stylesheets/file.css
+++ b/app/assets/stylesheets/file.css
@@ -121,6 +121,7 @@
   tr[aria-expanded="true"] > td:first-child::before,
   td[aria-expanded="true"]:first-child::before {
     transform: rotate(90deg);
+    transform-origin: center center;
   }
 
   .sul-embed-filename-cell,


jcoyne:
@dnoneill I believe so.

dnoneill (changes requested):
I know this is probably going to be a pain because you are going to have replicate it for the legacy viewer, this still doesn't look right. Right now we are rotating the right arrow down. I think rotating the down arrow up so that it a right arrow might be the fix for this. Right now it looks like this.

Screenshot 2025-02-13 at 3 13 47 PM

The figma has the arrows center on the folder icon.

jcoyne:
A contributing factor here is that this is the icon font, not the SVGs.

PR #2482, opened 2025-02-13
Create tooltips similar to Mirador

### LINKED ISSUE #2449: Add tooltips window a la Mirador
In mirador when you hover over the header and sidebar you see tooltips for each button.

Technically since we are doing tooltips we should add [aria-describedby](To make this accessible make sure to follow the guide here that describes how to use aria-describedby https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) but considering our tooltips are most likely going to match our aria-label, we probably shouldn't add accessibility because the screenreader will repeat the label and the tooltip which will be the same.

Image
Image
Image
Image

jcoyne:
Screenshot 2025-02-13 at 2 28 03 PM

Fixes #2449

### DIFF
--- a/app/assets/stylesheets/companion_window.css
+++ b/app/assets/stylesheets/companion_window.css
@@ -4,6 +4,21 @@
   --color-fog-light: #f4f4f4;
 }
 
+.tooltip {
+  position: absolute;
+  opacity: 0;
+  transition: opacity 100ms linear;
+  color: #ddd;
+  font-family: "Roboto", Helvetica, sans-serif;
+  font-size: 0.6rem;
+  background-color: #555;
+  padding: 0.3rem;
+  border-radius: 5px;
+  &.show {
+    opacity: 1;
+  }
+}
+
 .authLinkWrapper {
   align-items: center;
   display: flex;
--- a/app/assets/stylesheets/companion_window.css
+++ b/app/assets/stylesheets/companion_window.css
@@ -5,9 +5,7 @@
 }
 
 .tooltip {
-  position: absolute;
-  opacity: 0;
-  transition: opacity 100ms linear;
+  display: none;
   color: #ddd;
   font-family: "Roboto", Helvetica, sans-serif;
   font-size: 0.6rem;
@@ -15,7 +13,7 @@
   padding: 0.3rem;
   border-radius: 5px;
   &.show {
-    opacity: 1;
+    display: block;
   }
 }
 
--- a/app/assets/stylesheets/companion_window.css
+++ b/app/assets/stylesheets/companion_window.css
@@ -6,12 +6,12 @@
 
 .tooltip {
   display: none;
-  color: #ddd;
+  color: white;
   font-family: "Roboto", Helvetica, sans-serif;
-  font-size: 0.6rem;
-  background-color: #555;
-  padding: 0.3rem;
-  border-radius: 5px;
+  font-size: 0.6875rem;
+  background-color: rgba(97, 97, 97, 0.92);
+  padding: 4px 8px;
+  border-radius: 4px;
   &.show {
     display: block;
   }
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/companion_window.css
```diff
@@ -4,6 +4,19 @@
   --color-fog-light: #f4f4f4;
 }
 
+.tooltip {
+  display: none;
+  color: #ddd;
+  font-family: "Roboto", Helvetica, sans-serif;
+  font-size: 0.6rem;
+  background-color: #555;
```
dnoneill, line 12:
```suggestion
  background-color: #555;
  background-color: #616161;
```
Mirador has a background color of rgba(97, 97, 97, .92) which is #616161 in hex

dnoneill, line 12:
```suggestion
  font-weight: 500;
```

app/assets/stylesheets/companion_window.css
```diff
@@ -4,6 +4,19 @@
   --color-fog-light: #f4f4f4;
 }
 
+.tooltip {
+  display: none;
+  color: #ddd;
```
dnoneill, line 9:
```suggestion
  color: #ddd;
  color: white;
```
Mirador uses white

app/assets/stylesheets/companion_window.css
```diff
@@ -4,6 +4,19 @@
...
+  font-family: "Roboto", Helvetica, sans-serif;
+  font-size: 0.6rem;
+  background-color: #555;
+  padding: 0.3rem;
+  border-radius: 5px;
```
dnoneill, line 14:
```suggestion
  border-radius: 5px;
  border-radius: 4px;
```
Mirador uses 4px

app/assets/stylesheets/companion_window.css
```diff
@@ -4,6 +4,19 @@
   --color-fog-light: #f4f4f4;
 }
 
+.tooltip {
+  display: none;
+  color: #ddd;
+  font-family: "Roboto", Helvetica, sans-serif;
+  font-size: 0.6rem;
```
dnoneill, line 11:
```suggestion
  font-size: 0.6rem;
  font-size: 0.6875rem;
```

app/assets/stylesheets/companion_window.css
```diff
@@ -4,6 +4,19 @@
...
+  background-color: #555;
+  padding: 0.3rem;
```
dnoneill, line 13:
```suggestion
  padding: 0.3rem;
  padding: .25rem .5rem;
```

app/assets/stylesheets/companion_window.css
```diff
@@ -4,6 +4,19 @@
...
+  border-radius: 5px;
+  &.show {
+    display: block;
```
dnoneill, line 16: `z-index: 1000;` or whatever you think makes more sense

Without this the tooltips get cut off in the header

Screenshot 2025-02-13 at 5 40 13 PM
Screenshot 2025-02-13 at 5 39 56 PM
Screenshot 2025-02-13 at 5 39 53 PM

dnoneill (dismissed):

app/assets/stylesheets/companion_window.css
jcoyne: I'm not seeing this cut off behavior. How are you getting it?

dnoneill: Just by hovering with my mouse. Let me try in Chrome. I forgot my old version of Firefox does weird stuff.

dnoneill: Yup. It is an old firefox issue. I am not seeing this behavior in chrome or safari

dnoneill (approved):
I think this is probably a separate issue but we should apply this tooltip to https://github.com/sul-dlss/sul-embed/blob/354c44a6c8a026c956be94baca0ae92528e7ff2a/app/components/media/wrapper_component.rb#L21-L26

jcoyne:
@dnoneill I've added tooltips for the prev/next controls too.

dnoneill (approved):

PR #2487, opened 2025-02-13
Stop using the icon font for the expand/contract arrow for files

jcoyne:


dnoneill:
@jcoyne This isn't fixing the issue.

Screenshot 2025-02-13 at 5 30 24 PM

jcoyne:
@dnoneill The issue this PR attempts to fix is that we're using an icon font rather than MUI.

dnoneill (approved):

PR #2503, opened 2025-02-14
Be explicit about all the places the svg is getting styled

jcoyne:
I believe this gives the developer a better idea of all the components they should test.


dnoneill (changes requested):
with the add of the new web archive viewer this needs to be added.

dnoneill (approved):

PR #2504, opened 2025-02-14
Move transcript styles to the media stylesheet

jcoyne:
The companion window styles are shared amongst every viewer. Only the medai viewer needs these styles


marlo-longley:
@jcoyne needs a rebase, but works locally for me.

corylown (approved):

PR #2514, opened 2025-02-14
fix levels for file display

### LINKED ISSUE #2512: fix hierarchical files margin
This is happening both in the new and old viewer
Previously
Image

Now
Image

dnoneill:
This solves two problems:
1. There is tabbing on the parent (top level) folder. The old viewer (and figmas) have the "File name" heading and first folder tabbed at the same level. This was fixed by moving aria-level on back 1 (instead of 1, 0, instead of 2, 1, etc). This is what **level_step**
2. Files in the folders had the same tabbing as their parent folder. Fixed by regression to **"padding-left: #{((level - 1) * 2) + 1.5}ch;"**
Screenshot 2025-02-14 at 4 30 21 PM

closes #2512

### DIFF
--- a/app/components/file_component.html.erb
+++ b/app/components/file_component.html.erb
@@ -48,7 +48,8 @@
           </tr>
         </thead>
         <tbody>
-          <%= render Embed::File::DirRowComponent.new(viewer: viewer, dir: viewer.purl_object.hierarchical_contents)%>
+          <%# we set the level to -1 because without it, the level starts at 1 and that creates odd tabbing issues. %>
+          <%= render Embed::File::DirRowComponent.new(viewer: viewer, dir: viewer.purl_object.hierarchical_contents, level: -1)%>
         </tbody>
       </table>
     </div>
--- a/app/components/embed/file/dir_row_component.rb
+++ b/app/components/embed/file/dir_row_component.rb
@@ -20,7 +20,7 @@ module Embed
       end
 
       def first_td_style
-        "padding-left: #{((level - 1) * 4) + 2}ch;"
+        "padding-left: #{((level - 1) * 2) + 1.5}ch;"
       end
     end
   end


app/components/file_component.html.erb
```diff
@@ -48,7 +48,7 @@
           </tr>
         </thead>
         <tbody>
-          <%= render Embed::File::DirRowComponent.new(viewer: viewer, dir: viewer.purl_object.hierarchical_contents)%>
+          <%= render Embed::File::DirRowComponent.new(viewer: viewer, dir: viewer.purl_object.hierarchical_contents, level_step: 0)%>
```
jcoyne, line 51: Could we just revert the rest of this and send `level: -1` here?

app/components/embed/file/dir_row_component.html.erb
```diff
@@ -8,7 +8,7 @@
   </tr>
 <% end %>
 <% dirs.each_with_index do |dir, index| %>
-  <%= render Embed::File::DirRowComponent.new(viewer: viewer, dir: dir, pos_in_set: index + 1, set_size: child_set_size, level: level + 1)%>
+  <%= render Embed::File::DirRowComponent.new(viewer: viewer, dir: dir, pos_in_set: index + 1, set_size: child_set_size, level: level + level_step)%>
 <% end %>
 <% files.each_with_index do |file, index| %>
   <%= render Embed::File::FileRowComponent.new(viewer: viewer, file: file, pos_in_set: dirs.size + index + 1, set_size: child_set_size, level: level + 1)%>
```
jcoyne, line 14: Should this be using level_step too?

app/components/file_component.html.erb
dnoneill: that's smarter. lol

jcoyne (approved):

PR #2516, opened 2025-02-14
add location restriction banner for file viewer

dnoneill:
for monday:
- move messages for restricted content into locales
- fix icon padding


app/viewers/embed/viewer/file.rb
```diff
@@ -36,6 +36,17 @@ def display_file_search?
                                  @purl_object.contents.map(&:files).flatten.length >= min_files_to_search
       end
 
+      def message
+        return { type: 'embargo', message: embargo_message } if @purl_object.embargoed?
+
+        if @purl_object.location_restriction
```
jcoyne, line 42: Won't this show the banner even when the person is in this access location?  Shouldn't we run the auth/probe service on stacks to see if they are in a restricted situation or not?

config/settings.yml
```diff
@@ -50,3 +50,13 @@ min_files_to_search_default: 10
 language_subtag_registry:
   path: vendor/data/language-subtag-registry
   url: "https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry"
+
+locations:
```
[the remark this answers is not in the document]
jcoyne: I believe the code you should be looking at in sul-embed is here: https://github.com/sul-dlss/sul-embed/blob/3b70cc585d1f064a6593e0dc06367e6abdb01d23/app/javascript/controllers/file_auth_controller.js#L124-L148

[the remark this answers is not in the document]
dnoneill: Yeah the file_auth_controller isn't used by the file viewer

app/viewers/embed/viewer/file.rb
dnoneill: Nope. The ticket says ```If possible the banner should persist even when the user is in the authorized location, as there should be a persistent indicator that the items are restricted.```

config/settings.yml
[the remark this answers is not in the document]
jcoyne: I'm wondering if we should try to be consolidating the auth rather than having each viewer have separate auth mechanisms.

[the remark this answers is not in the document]
dnoneill: I think that would be preferable but I also think that might need to be a storytime. From what I can tell each viewer might use a different auth mechanism. I think the mechanism for PDF vs. IIIF might be different. I honestly would need to have more context for this before I am comfortable with writing a ticket for this.

app/models/embed/purl_json_loader.rb
```diff
@@ -49,6 +51,17 @@ def embargoed
       json.dig('access', 'embargo').present?
     end
 
+    def location_restriction
```
hudajkhan, line 54: This works too but am wondering about possible connections with the resource_file/file implementations: https://github.com/sul-dlss/sul-embed/blob/main/app/models/embed/purl/resource_file.rb#L24 and https://github.com/sul-dlss/sul-embed/blob/main/app/models/embed/purl/file_json_deserializer.rb#L23, although for file, it only seems to be checking for download whereas you are checking for both view and download.

app/viewers/embed/viewer/file.rb
```diff
@@ -36,6 +36,17 @@ def display_file_search?
...
+          return { type: 'location-restricted',
+                   message: I18n.t('restrictions.restricted_access', location: @purl_object.restricted_location) }
```
hudajkhan, line 44: Might it help to have a fallback empty string in the case where there is a code that (for some reason) does not map to what we have in settings.  That will probably not the case, but may be good to have something just in case.

dnoneill: Yeah I was trying to figure out what the best way of doing this is. Because I would think we would want the error or something to display wrong if the cocina is wrong. @andrewjbtw Do you have a preference?

andrewjbtw: This is in case a new location is introduced in cocina and used before this code is updated? Cocina validation should prevent an item from being published with a code not on the list.

If we can assume that an item with an unexpected code (for banner purposes) is still valid Cocina, I would fall back on a generic message rather than an error since the new rights setting should be active and supported in Stacks by the time someone uses it. 

> Access is restricted to on-site visitors to the Stanford Libraries. See Access conditions for more information.

hudajkhan:
I tried it out locally and looks good!
Screenshot 2025-02-20 at 8 38 10 PM

app/models/embed/purl_json_loader.rb
dnoneill: ResourceFile is the files on the Purl. Basically in order to do what you are suggesting I would need to go into the structural and look at every file to get this data, which is already in the access field in Cocina.

hudajkhan (approved):

PR #2519, opened 2025-02-18
Remove unused XML deserializer

jcoyne:
We've now switched to Cocina, we don't need this


dnoneill:
@jcoyne I think the FileXmlDeserializer would also get removed with this PR

jcoyne:
Sounds good.

dnoneill (approved):

Issue #2521, opened 2025-02-18
Standardize PDF Auth view


dnoneill:
In all other viewers, the auth banner spans both the sidebar and the viewer:
![Image](https://github.com/user-attachments/assets/56306035-1490-41df-b89e-60df4b162d80)

In PDF it just spans the viewer

![Image](https://github.com/user-attachments/assets/09f3603c-3c61-4dd0-a08d-8542e53781c1)

Issue #2320, opened 2025-01-31
Document/PDF: Add placeholder image for access restricted items


alundgard:
When a document/PDF is access restricted (rights download = none, stanford, or location) and the user is not authenticated, display a placeholder locked image in the viewer. We can use the image currently used for media:
https://github.com/sul-dlss/sul-embed/blob/main/app/assets/images/locked-media-poster.svg

### Current
![Image](https://github.com/user-attachments/assets/698bb325-2070-455c-9370-52d992546d31)

### Design
![Image](https://github.com/user-attachments/assets/b790de8f-3d7d-4757-9ff5-40f117cd6f97)

### Examples
stage druid | stage argo | stage purl | content type | rights view | rights download
-- | -- | -- | -- | -- | --
xz332pr7270 | [argo-stage](https://argo-stage.stanford.edu/view/xz332pr7270) | [purl-stage](https://sul-purl-stage.stanford.edu/xz332pr7270) | document | world | none
rc562ym8157 | [argo-stage](https://argo-stage.stanford.edu/view/rc562ym8157) | [purl-stage](https://sul-purl-stage.stanford.edu/rc562ym8157) | document | stanford | stanford
vg127st8023 | [argo-stage](https://argo-stage.stanford.edu/view/vg127st8023) | [purl-stage](https://sul-purl-stage.stanford.edu/vg127st8023) | document | stanford | none
dd501hp6742 | [argo-stage](https://argo-stage.stanford.edu/view/dd501hp6742) | [purl-stage](https://sul-purl-stage.stanford.edu/dd501hp6742) | document | location | location
bm739vc0410 | [argo-stage](https://argo-stage.stanford.edu/view/bm739vc0410) | [purl-stage](https://sul-purl-stage.stanford.edu/bm739vc0410) | document | location | none
wj528md5003 | [argo-stage](https://argo-stage.stanford.edu/view/wj528md5003) | [purl-stage](https://sul-purl-stage.stanford.edu/wj528md5003) | document | citation | none

jcoyne:
@alundgard we should consider the case where "download: stanford" and the user is signed in, or when the user is in the location specified.

Issue #2523, opened 2025-02-18
Fix weird scroll behaivor


dnoneill:
Seems to be a document page issue.
https://sul-purl-uat.stanford.edu/bb005wc0080?new_viewer=true
https://purl.stanford.edu/bb461xx1037.

But for some reason the viewer scrolls when it shouldn't.

https://github.com/user-attachments/assets/1c41de3f-2756-4d5f-b70f-e1229272c3bd

dnoneill:
@jcoyne The PR did not fix the issue. It is still a problem: https://sul-purl-uat.stanford.edu/bb461xx1037

dnoneill:
I think this might be related. In the new 3d viewer the height isn't quite right either. Let me know if you want me to open a new ticket.
https://sul-purl-uat.stanford.edu/bc769sr4504?new_viewer=true

![Image](https://github.com/user-attachments/assets/3ed6ebc2-c3e5-4f6f-9948-3c36366e9a15)

dnoneill:
It is only happening on some 3d items.

dnoneill:
Also happens on maps after you click on an item on the map. https://purl.stanford.edu/bb033gt0615?new_viewer=true

dnoneill:
This is still an issue with PDF

corylown:
@dnoneill I think what you're seeing is related to https://github.com/sul-dlss/sul-embed/issues/2534. I don't think we want to be setting `overflow: scroll` on `.sul-embed-container`. I'm looking at a fix.

jcoyne:
@dnoneill is there anything left to do here?

dnoneill:
@jcoyne I think so. I think removing the scroll just hid the problem of the container not being the right side. There is some weirdness at the bottom of the sidebar. Also if you look at the scroll bar on the pdf it is being cut off.

![Image](https://github.com/user-attachments/assets/4e3f97a2-7680-4c1e-a41b-05953946bb31)

![Image](https://github.com/user-attachments/assets/ccf0ed9e-ac30-453b-9a31-ff6e06752574)

dnoneill:
@jcoyne Something must have changed in 17 hours because I am not seeing this issue anymore

PR #2525, opened 2025-02-18
Display the login banners as headers

jcoyne:
This provides parity with the media viewer
Screenshot 2025-02-18 at 1 08 29 PM

closes [#2320](https://github.com/sul-dlss/sul-embed/issues/2320)
closes [#2521](https://github.com/sul-dlss/sul-embed/issues/2521)

### DIFF
--- a/app/javascript/controllers/locked_poster_controller.js
+++ b/app/javascript/controllers/locked_poster_controller.js
@@ -1,0 +1,12 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  // Display the lock
+  show() {
+    this.element.hidden = false
+  }
+
+  hide() {
+    this.element.hidden = true
+  }
+}
--- a/app/components/pdf_component.html.erb
+++ b/app/components/pdf_component.html.erb
@@ -7,19 +7,23 @@
     <%= render CompanionWindows::FullscreenButtonComponent.new  %>
   <% end unless requested_by_chromium? %>
 
+  <% component.with_authorization_messages do %>
+    <%= render LoginComponent.new %>
+    <%= render ContentNotAvailableComponent.new unless viewer.available? %>
+  <% end %>
+
   <% component.with_body do %>
+    <div data-controller="locked-poster" data-action="auth-success@window->locked-poster#hide auth-denied@window->locked-poster#show" tabindex="-1" role="presentation" style="flex: 1 0 100%; text-align: center;" <%= 'hidden' if viewer.available? %>>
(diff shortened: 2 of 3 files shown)


dnoneill (changes requested):
It is not catching changes. When I deploy this to uat and try to log in I get this view:

Screenshot 2025-02-18 at 2 31 43 PM

app/javascript/controllers/locked_poster_controller.js
```diff
@@ -0,0 +1,12 @@
+import { Controller } from "@hotwired/stimulus"
```
dnoneill, line 1: Also there is already a `locked-media-poster` controller. Can we consolidate these?

jcoyne: No, they are different because media may have a poster to show, but for a pdf, we always show the lock icon.

dnoneill: Wait shouldn't we always show the lock icon, even if there is a poster?

jcoyne:
@dnoneill it's working fine for me on stage: Screenshot 2025-02-18 at 1 48 35 PM

dnoneill:
Also there is a little bit or margin or padding at the end of the header:
Screenshot 2025-02-18 at 3 27 45 PM

When you open the drawer, the image does not responsively resize. It just gets pushed to the side
Screenshot 2025-02-18 at 3 27 54 PM

app/javascript/controllers/locked_poster_controller.js
jcoyne: I don't know what the behavior is expected for media. My inclination is not to change anything not related to the ticket.

jcoyne: Presently media displays the poster, not a lock.

dnoneill: I am going to put this in post standup for tomorrow but I don't think it's a blocker. We can always consolidate in another ticket.

dnoneill:
The login icon spilling over is still an issue only in the iframe view but for some reason isn't causing a problem outside the already existing problem of https://github.com/sul-dlss/sul-embed/issues/2523. Just noting in case.

jcoyne:
@dnoneill currently we're saying the lock will be `height: 100%`.  I'm not sure how we make that change to width: 100% at a different aspect ratio.

dnoneill:
I am just making a note if this comes up as an issue but right now when it is in purl it doesn't cause an issue.

dnoneill (approved):

PR #2537, opened 2025-02-18
add su-underline to download modal, sidebar, contents, download link

### LINKED ISSUE #2454: Standardize link style
We would like the link styling to be consistent with the `su-underline` style appearing on Purl: digital blue text color, gray dotted underline.

Image

Example: https://purl.stanford.edu/yx282xq2090.

Note: This is already the link style in the legacy viewer, so we could revert to using that. This will be a divergence from the current Mirador style, since we are moving away from using Cardinal red for links.

There are a few places where links appear in sul-embed:
- Download modal (all viewers)
- Sidebar Purl link (all viewers)
- Contents panel (Media, Document/PDF)
- File list download links (File)

dnoneill:
closes #2454 
Screenshot 2025-02-18 at 5 17 04 PM
Screenshot 2025-02-18 at 5 15 54 PM
Screenshot 2025-02-18 at 5 15 38 PM

### DIFF
--- a/app/assets/stylesheets/common.css
+++ b/app/assets/stylesheets/common.css
@@ -1,6 +1,7 @@
 /* For styles shared between the legacy and modern viewers */
 :root {
   --link-color: #006cb8;
+  --stanford-40-black: #ababa9;
 }
 /* From Bootstrap 5.3 */
 .visually-hidden-focusable:not(:focus):not(:focus-within):not(caption),
@@ -39,4 +40,8 @@ a:focus {
 
 .sul-embed-location-restricted-text {
   color: var(--stanford-digital-red);
+}
+
+.su-underline {
+  text-decoration: underline dotted var(--stanford-40-black) 1px;
 }
--- a/app/assets/stylesheets/file.css
+++ b/app/assets/stylesheets/file.css
@@ -57,6 +57,10 @@
 
   th,
   td {
+    a {
+      text-decoration: none;
+    }
+
     &:nth-of-type(2) {
       padding-inline: 1rem;
       text-align: center;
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/common.css
```diff
@@ -47,4 +47,10 @@ a:focus {
 .logoContainer {
   border-top: 1px solid var(--border-color);
   border-bottom: 1px solid var(--border-color);
-}
\ No newline at end of file
+}
+
+
+.su-underline {
+  --underline-color: #adb5bd;
```
jcoyne, line 54: Where does this color come from? I don't see it in https://identity.stanford.edu/design-elements/color/accent-colors/

dnoneill: I got it from the component library. https://github.com/sul-dlss/component-library/blob/349ef122c2724f7375e0d5cd686a650c4550d657/styles/links.css#L24

jcoyne: We're changing that to stanford 40% black: https://github.com/sul-dlss/component-library/pull/119

dnoneill: Sounds good. Updated.

app/assets/stylesheets/common.css
```diff
@@ -40,3 +40,8 @@ a:focus {
 .sul-embed-location-restricted-text {
   color: var(--stanford-digital-red);
 }
+
+.su-underline {
+  --underline-color: #ababa9;
```
jcoyne, line 45: would you mind calling this variable `--stanford-40-black` to align with the component library?

dnoneill:
@hudajkhan @jcoyne done. I also moved --stanford-40-black up into root instead of having it be in the .su-underline class

hudajkhan (approved):

PR #2442, opened 2025-02-12
Improve the sandbox

jcoyne:

PR #2546, opened 2025-02-19
add new_viewer to sandbox

dnoneill:


jcoyne:
Can we do https://github.com/sul-dlss/sul-embed/pull/2442 first? Then we can just click the links and it'll take you to the new viewer?

jcoyne (approved):

Issue #2533, opened 2025-02-18
File icons appear smaller and off-center


alundgard:
https://purl.stanford.edu/mk520hp6879?new_viewer=true
<img width="1305" alt="Image" src="https://github.com/user-attachments/assets/d0a28032-0c88-430e-af03-5baeb8e4e899" />

<img width="1301" alt="Image" src="https://github.com/user-attachments/assets/e861f98a-0408-4fad-8b6f-b934f571886f" />

jcoyne:
@alundgard the smaller I can fix. I think @dnoneill purposely added that extra indent, but I will remove it.

jcoyne:
@alundgard that design shows two different indent levels. See files under "2F555" vs the files under "Folder name". I'm not sure I understand the intent.

PR #2554, opened 2025-02-19
fix non hierarchical file padding

### LINKED ISSUE #2536: Not enough padding on file name
You will have to be careful because it seems like we need two different things based on the file structure. Structured files are appearing correctly: https://sul-purl-uat.stanford.edu/fg478vy8624?new_viewer=true but files without hierarchy aren't. https://sul-purl-uat.stanford.edu/bb007hx5508?new_viewer=true. 

I think this is related to https://github.com/sul-dlss/sul-embed/pull/2514 and/or https://github.com/sul-dlss/sul-embed/commit/5d63b27e0d8598f1d98c2c7ca2a1160dc9dea533#diff-d26d1174d86492e8815c8126aa8e0438dd2e1f4bfe4bda43d8eb5ce8a133e4fdR31. 

This is working fine in the old viewer

dnoneill:
closes #2536 
Screenshot 2025-02-19 at 1 11 48 PM
Screenshot 2025-02-19 at 1 11 41 PM

### DIFF
--- a/app/components/embed/file/file_row_component.rb
+++ b/app/components/embed/file/file_row_component.rb
@@ -8,7 +8,9 @@ module Embed
         @file = file
         @pos_in_set = pos_in_set
         @set_size = set_size
-        @level = level
+        # level needs to be set to -1 for hierarchical files
+        # but causes problems padding problems for non-structured files
+        @level = level.zero? ? 1 : level
       end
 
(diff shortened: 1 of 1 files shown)


dnoneill:
I think this should get fixed by https://github.com/sul-dlss/sul-embed/issues/2533

dnoneill:
Fixed by https://github.com/sul-dlss/sul-embed/issues/2533

PR #2575, opened 2025-02-20
Make visited links on sandbox page purple

marlo-longley:
@dnoneill re conversation in https://github.com/sul-dlss/sul-embed/pull/2442

### DIFF
--- a/app/views/pages/sandbox.html.erb
+++ b/app/views/pages/sandbox.html.erb
@@ -90,12 +90,22 @@
       &> div {
         margin-inline-end: 1rem;
       }
+      a {
+        color: blue;
+        text-decoration: none;
+      }
+
+      a.visited {
+        color: purple;
+      }
     }
   </style>
   <script>
-     function openPurl(purl) {
+     function openPurl(purl, event) {
        const url = document.getElementById('url-scheme')
        url.value = `https://purl.stanford.edu/${purl}`
+       const linkElement = event.target;  
+       linkElement.classList.add('visited');
        embed.render()
      }
   </script>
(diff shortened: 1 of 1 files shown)


dnoneill (approved):
@marlo-longley needs a rebase but then should be good to go

app/views/pages/sandbox.html.erb
```diff
@@ -441,6 +451,15 @@
     });
 
   });
+
+  document.addEventListener("DOMContentLoaded", function () {
+    const links = document.querySelectorAll('.examples a');
```
jcoyne, line 456:
```suggestion
    const links = document.querySelectorAll('[data-purl]');
```

jcoyne:
This doesn't persist the visited state after the page is refreshed.  @dnoneill was that your intention?

jcoyne:
Also noting that if you fill in  `https://purl.stanford.edu/sf815vr1246` and submit, it doesn't set `.visited` on the link with `data-purl="sf815vr1246"`.

marlo-longley:
Thanks for reviewing @jcoyne and @dnoneill  -- this was just my quick thought on a fix to what Niqui mentioned. I don't know what standard we want to hold this too but I'm happy to make changes. It is always going to be a bit hacky because of the `#` hrefs I think.

jcoyne:
@marlo-longley my assumption was that @dnoneill wanted some kind of client-side router so that the href would be different for each link and use the browser native `:visited` style.

marlo-longley:
@jcoyne gotcha. I definitely don't feel strongly about this so if it's not what we intended, happy to close.

dnoneill:
I mean I figured if you changed the href #druid it would fix it, but I honestly don't care, I would just like the sandbox to stop making everything visited when I haven't visited it.

PR #2577, opened 2025-02-20
Set leaflet NoWrap and bounds to prevent map & layers from repeating

### LINKED ISSUE #2550: Odd behaivor new geo viewer
In the new geo viewer, https://sul-purl-uat.stanford.edu/bb058zh0946?new_viewer=true the overlay is repeated multiple times on page load
Image

In the old viewer, https://sul-purl-uat.stanford.edu/bb058zh0946?new_viewer=false, it does not have this behavior on page load.

Image

corylown:
Fixes #2550 

Before:
Screenshot 2025-02-20 at 11 00 08 AM

After:
Screenshot 2025-02-20 at 10 59 38 AM

### DIFF
--- a/app/javascript/src/modules/geo_viewer.js
+++ b/app/javascript/src/modules/geo_viewer.js
@@ -15,6 +15,8 @@ export default {
     const attribution = `&copy <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors, &copy <a href="http://carto.com/attributions">Carto</a>`
     L.tileLayer('https://{s}.basemaps.cartocdn.com/light_all/{z}/{x}/{y}@2x.png', {
       maxZoom: 19,
+      noWrap: true,
+      bounds: [[-90, -180],[90, 180]],
       attribution: attribution,
     }).addTo(this.map)
     this.highlightLayer = L.layerGroup().addTo(this.map)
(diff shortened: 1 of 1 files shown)


dnoneill:
@corylown The zoom is still doing odd things compared to the legacy viewer

corylown:
@dnoneill I suspect that has to do with the different size/shape of the map container in the new viewer. There's a way to increase the zoom beyond what gets calculated by leaflet via `fitBounds`, but I'm not sure we should do that until we migrate completely to the new geo viewer so we don't negatively impact the legacy geo viewer. I can spend a little more time on it now but I think it might be better to handle it in a separate commit.

jcoyne:
Could this cause problems for any datasets that span  180 longitude?  I know the Aleutian island chain crosses 180. Do we have any data like that to test on?

jcoyne:
gk831nh0251 is a good test item for crossing 180.

dnoneill:
@corylown I think there is something wrong with the calculations. I think the calculations for the area to zoom to need to be done after we know if the sidebar is open or not. This is the view of gk831nh0251 with the sidebar open vs closed. As you can see the data on the right is showing up only after you close the sidebar. Also I think this crossing 180 is an okay. I am also including a world repeat in this and it looks pretty similar.
Screenshot 2025-02-20 at 5 38 09 PM
Screenshot 2025-02-20 at 5 38 03 PM

World repeat in current viewer
Screenshot 2025-02-20 at 5 40 32 PM

corylown:
@dnoneill I can't duplicate what you're seeing. When I load https://purl.stanford.edu/gk831nh0251 in the new geo viewer the map is loading at the correct size so the data on the far left and right is all visible:

Screenshot 2025-02-21 at 10 52 08 AM

dnoneill:
@corylown seems to be dependent on browser size.

PR #2611, opened 2025-02-20
Remove tabindex from presentational image

### LINKED ISSUE #2603: pdf lock image, mark as decorative
Presentational element is exposed to assistive technologies

An element has been marked as decorative, but it remains exposed to assistive technologies.

This may be because:

    the element has an ARIA attribute (these contain "instructions" for assistive technologies), or
    the element is focusable (meaning it can still be reached by keyboard)

Page elements can be marked as decorative if they do not add any new information to the page. This can improve the user experience for people who use screen readers - by cutting out irrelevant "noise".

In this case, the element seems to be for more than just decoration. Either the element can be reached by keyboard (it's interactive), or it has an ARIA attribute which communicates something to assistive technologies.

It's important to note that marking an element as decorative (or giving it a presentational role) does not always mean it will be ignored by assistive technology.

    This is a best practice only and not currently required for WCAG conformance
    Siteimprove Alfa: [Opens in new tab:Technical documentation for rule SIA-R86](https://alfa.siteimprove.com/rules/sia-r86)

Abilities affected:
Vision

Image

jcoyne:
Fixes #2603

### DIFF
--- a/app/components/media_component.html.erb
+++ b/app/components/media_component.html.erb
@@ -88,11 +88,11 @@
   <% end %>
 
   <% component.with_body do %>
-      <div data-controller="locked-media-poster" data-action="auth-success@window->locked-media-poster#hide auth-denied@window->locked-media-poster#show" tabindex="-1" role="presentation" style="flex: 1 0 100%;" hidden>
+      <div data-controller="locked-media-poster" data-action="auth-success@window->locked-media-poster#hide auth-denied@window->locked-media-poster#show" role="presentation" style="flex: 1 0 100%;" hidden>
         <%= LockedStatusComponent.new %>
       </div>
 
-      <div data-controller="locked-audio-poster" data-action="auth-success@window->locked-audio-poster#hide auth-denied@window->locked-audio-poster#show" tabindex="-1" role="presentation" style="flex: 1 0 100%;" hidden>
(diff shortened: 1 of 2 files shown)


dnoneill:
@jcoyne The error is still appearing with the extension on this branch

dnoneill (changes requested):
I think you need to put the aria-hidden on the div. 
Screenshot 2025-02-24 at 11 41 11 AM

jcoyne:
That div has `role="presentation"` on it already.  Doesn't that take care of this?

dnoneill:
I think the problem is it has presentation on it. According to the site improve ```Presentational element is exposed to assistive technologies. An element has been marked as decorative, but it remains exposed to assistive technologies.```

jcoyne:
> The presentation role and its synonym none remove an element's implicit ARIA semantics from being exposed to the accessibility tree.

According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role. 

It seems like there may be a bug in siteimprove here.

thatbudakguy:
This is total speculation, but I wonder if SiteImprove is getting caught up on the interaction between `role=presentation` and `tabindex=-1`. The explicit assignment of a tabindex (even though it's negative) might mark the element as focusable (although only programmatically), which is conflicting with it being removed from the a11y tree entirely.

dnoneill:
@thatbudakguy You were right.
@jcoyne can you do this on the pdf viewer too?

jcoyne:
yep, got all three cases now.

dnoneill (approved):

dnoneill:
Thank you

PR #2613, opened 2025-02-20
don't render videojs full screen button when in fullscreen for safari

### LINKED ISSUE #2567: Safari: Video.js full screen button does nothing when in Viewer full screen
Safari 18.2. Works as expected in Chrome and Firefox.
- Example: https://sul-purl-uat.stanford.edu/qf378nj5000
- Open the Viewer full screen by clicking Full screen button (Viewer top-right)
- Attempt to open Video.js full screen by clicking Full screen button (Video.js controls botton-right)

dnoneill:
checked in firefox, chrome and safari

closes #2567

### DIFF
--- a/app/components/media/tag_component.rb
+++ b/app/components/media/tag_component.rb
@@ -74,6 +74,7 @@ module Media
           media_tag_target: 'authorizeableResource',
           controller: 'media-player',
           action: 'media-seek@window->media-player#seek ' \
+                  'fullscreenchange@window->media-player#fullscreenChange ' \
                   'auth-success@window->media-player#initializeVideoJSPlayer'
         },
         poster: poster_url_for,
--- a/app/javascript/controllers/media_player_controller.js
+++ b/app/javascript/controllers/media_player_controller.js
@@ -37,6 +37,18 @@ export default class extends Controller {
     })
   }
 
+  // Don't show videojs fullscreen button when sul-embed is in fullscreen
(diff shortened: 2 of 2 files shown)


jcoyne:
I thought the purpose here was not to render the fullscreen button only when on iOS and the viewer was already in fullscreen mode.

dnoneill:
that is a good question.

dnoneill:
@jcoyne You were right about the correct way this is implemented. I have updated the code to work that way.

app/components/media/tag_component.rb
```diff
@@ -87,6 +86,18 @@ def media_tag # rubocop:disable Metrics/MethodLength
       end
     end
 
+    def action
+      action = 'media-seek@window->media-player#seek ' \
+               'auth-success@window->media-player#initializeVideoJSPlayer'
+
+      action += ' fullscreenchange@window->media-player#fullscreenChange' if safari?
```
jcoyne, line 93: Could we just do this all the time, and put all the conditionals on the front end? Then it's more or less all in the same place rather than spread across front and back ends.

dnoneill: I thought about that, but I decided against it because everytime a user goes fullscreen, then this function is going to be run, no matter the browser and that just adds processing that might not be necessary

jcoyne (approved):

Issue #734, opened 2016-12-19
full-page embed (related to #718)


abrin:
Hi,
  Would it be possible to for the embed to fill both the full width (which it currently does thanks to #718 and #716) as well as the full page height?  I've attached a video (animated gif)  of how it's being used in the Stanford map libraries to better demonstrate the use case.  On a large 52 inch? screen the normal mode renders a super-tiny image.

![ezgif com-29c713f981](https://cloud.githubusercontent.com/assets/192686/21299001/6fe6f08e-c554-11e6-90de-3880c3aca269.gif)

thanks!

jkeck:
@abrin as best I can tell, it doesn't appear that this page is using the `fullheight` option as mentioned in https://github.com/sul-dlss/sul-embed/issues/718#issuecomment-250636275.  Have you updated the page to take advantage of the new option as described in the comment linked above?

abrin:
you're right it got dropped out, and I missed it when I was debugging, apologies.

Issue #716, opened 2016-09-22
`100%` or `100%25` as a width parameter no longer forces embed to 100% width


mejackreed:
Functional change in sul-embed, that when using `100%25` (100% html encoded) embed no longer forces the width to 100%.

Reported by @abrin in relation to some issues they are having with the Rumsey map room.

Example: http://www.davidrumsey.com/luna/stamen.jsp?pid=wv304jy9677

jkeck:
@mejackreed thanks for reporting this.

Interesting, I wasn't aware that anybody was passing through things like percentages.  This has changed because I realized, that in certain cases, we were blindly placing the user's parameter directly into the markup (which as you know can be hugely problematic).  So, I think this _happened_ to work because it was piggybacking on a vulnerability in our codebase (whoops).

You'll notice in the [oEmbed spec](http://oembed.com/) that `maxwidth` and `maxheight` are always an integer.  So, we now call `to_i` on the param to ensure that we're getting a sane value before putting it into the viewer markup.  This is causing `100% or `100%25` to get turned into `100`.

```
>> '100%'.to_i
=> 100
>> '100%25'.to_i
=> 100
```

But, fear not! 100% width is obviously a primary use case of SUL Embed.  What we've done now is allow the embed content to flow to 100% width whenever a `maxwidth` is not specified.  So in order for @abrin to get 100% width on SUL Embed, they simply need to not provide any `maxwidth` parameter and everything should work out fine ([see PURL](https://purl.stanford.edu/wv304jy9677))

jkeck:
Also, just curious if there is a particular reason why these reports are filtering through a single dev?

I worry that is not a scalable solution and could potentially delay our ability to respond in as timely of a manner as we could if the reports were going through the proper channels (although I trust @mejackreed will always disseminate the information properly and quickly).

abrin:
@jkeck it's more a matter of I wasn't aware of the GitHub repo and asked @mejackreed for help and to point me at the issue tracker.  He was kind enough to just create the ticket.  

Removing the parameters in the URL appeared to have worked.

thanks!

jkeck:
Wonderful @abrin.  Glad to see that it worked out.

Please feel free to submit feedback, bugs, and feature requests here in the future!

PR #2628, opened 2025-02-21
Remove legacy fullheight feature

jcoyne:
Nothing in sul-dlss uses this feature

### DIFF
--- a/app/components/iframe_component.rb
+++ b/app/components/iframe_component.rb
@@ -14,10 +14,6 @@ class IframeComponent < ViewComponent::Base
     width || '100%'
   end
 
-  def height_style
-    embed_request.fullheight? ? '100%' : height
-  end
-
   def src
     query_params = embed_request.as_url_params.merge(version ? { _v: version } : {}).to_query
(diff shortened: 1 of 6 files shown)


corylown:
Are you sure the Map Center is no longer relying on this feature?

There's some history:

- https://github.com/sul-dlss/sul-embed/issues/734
- https://github.com/sul-dlss/sul-embed/issues/718
- https://github.com/sul-dlss/sul-embed/issues/716

The example the map center shared still works: https://www.davidrumsey.com/luna/stamen.jsp?pid=wv304jy9677

jcoyne:
I'm pretty sure they are not using `fullheight`, because their embed response in the source says `"height":520`.  They are using css to override it.  Additionally, I don't see `fullheight` in the server logs.  With #2653 , they will be able to request `maxheight=100vh` or `maxheight=100%` instead.

dnoneill:
Waiting on hearing back from Stace

jcoyne:
Stace said that it was okay to remove this.

corylown (approved):

PR #2629, opened 2025-02-21
Recalculate geo/map fitBounds after the left drawer opens

corylown:
Relates to #2550


dnoneill:
Is there a reason we aren't using an event in https://github.com/sul-dlss/sul-embed/blob/main/app/javascript/controllers/companion_window_controller.js#L13 that the geo controller would know to listen for and wouldn't init until it triggers? The behavior might be a little more expected? I did have to combine this with your height PR

https://github.com/user-attachments/assets/a5efb4a2-9f2e-4a29-8005-b58624cfb28e

PR #2630, opened 2025-02-21
Set up location restriction message/allow pdf viewer to authorize location restriction

### LINKED ISSUE #2360: Location-restricted PDFs do not display in the PDF viewer even at an authorized location
**Describe the bug**

When at an authorized location, location-restricted PDFs do not open in the viewer. It is possible to download the PDFs from direct links to the Stacks URL but the viewer presentation looks the same as when you are not in an authorized location.

I tested this while on campus visiting the Special Collections reading room. Geoff Willard (@oceanofsound ) also verified this from his workstation that is authorized to view location-restricted materials.

**Expected behavior**

When at an authorized location, the viewer should present PDFs as if the user has logged in. This is what happens with location-restricted media and images.

**Additional information**

I'm not sure how we can test this remotely. Whoever picks up the ticket may need to coordinate with Geoff if they need to see what the browser is doing when visiting PDF.

**To reproduce**

From an authorized location:

1. Visit [this purl](https://purl.stanford.edu/zs884sy8048).
2. See that the PDF is not available
3. Try this [Stacks link](https://stacks.stanford.edu/file/druid:zs884sy8048/GitaiKW_cover24_2.pdf): 
4. See that the PDF can be downloaded, which means the location-restriction is being applied to the file.

From outside an authorized location:

Follow the steps above and see that neither the PDF viewer nor the Stacks link will allow access to the PDF.

hudajkhan:
Closes #2360 

What this pull request does:
- Adds location restriction methods to pdf viewer: does the purl object have a location restriction, and what is the location
- Sets up a component to display the location restriction message based on a given location code 
- Displays the location restriction component in the PDF component display if the object has a location restriction
- If the iiif auth (in file_auth_controller) returns an access denied message AND the file in question is the first file (i.e. the file which will have the render viewer method called and the PDF displayed in the screen), then an auth denied event is dispatched which will trigger the locked icon in the screen.
- Allows an item that is restricted by location to be visible to a user at that location (i.e. when the authorization process returns success)
- Extends the viewer.available function to also include location restricted cases, so the above scenario is now possible whereas before any location restricted case would immediately display the content not available method
- Fixes a bug in the fallback message method which was needed a return statement

What this pull request does not do:
Check the authorization of files other than the first one to be rendered. This issue can address that separately: https://github.com/sul-dlss/sul-embed/issues/2651


dnoneill:
@hudajkhan this needs a rebase

hudajkhan:
@dnoneill just did the rebase. I do see the related PR, and open to additional suggestions for consolidating work.

app/components/pdf_component.html.erb
```diff
@@ -11,6 +11,7 @@
     <div role="banner" aria-label="Access message">
       <%= render LoginComponent.new %>
       <%= render CompanionWindows::ContentNotAvailableComponent.new unless viewer.available? %>
+      <%= render LocationRestrictionComponent.new(viewer.restricted_location) if viewer.location_restricted? %>
```
jcoyne, line 14: Can't we display this all of the time, without the check, and wait for the front end to say "yes location restricted".  Otherwise we are checking once on the front end and once on the back end.  That doesn't seem very efficient.

jcoyne (changes requested):

app/components/location_restriction_component.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# This is only used by the PdfComponent
+class LocationRestrictionComponent < ViewComponent::Base
+  def initialize(location = 'site visitors to the Stanford Libraries')
```
jcoyne, line 5: We never initialize this without an argument, do we? Why do we need a default value?

jcoyne (changes requested):

app/javascript/controllers/file_auth_controller.js
```diff
@@ -133,6 +134,11 @@ export default class extends Controller {
                                        // with a link to the media server file location (and media token)
                                        // and this can happen with a non-media object that happens to have
                                        // a media file in it, e.g. ds777pr3860
+        
+        if(this.isLocationRestricted(json, contentResourceId)) {
```
jcoyne, line 138: I thought we were going to do location restriction completely on the backend, so that we were able to show the location restriction banner even when they were in the location.  with that decision, I don't know why we want/need any location-restricted stuff in the frontend.

PR #2631, opened 2025-02-21
fix label_of_filename for pdfs

dnoneill:
We only want text_representation_label for media not pdfs

### DIFF
--- a/app/models/embed/purl/resource_file.rb
+++ b/app/models/embed/purl/resource_file.rb
@@ -51,7 +51,7 @@ module Embed
       end
 
       def label_or_filename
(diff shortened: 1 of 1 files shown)


jcoyne (changes requested):
Can we add unit test for these cases?

dnoneill:
@jcoyne If you want to take this over for me since it is something we would preferrably like to get in before merge on Monday I would appreciate it. It is almost 7 here. I am leaving for the day.

dnoneill:
We need to have a discussion on this logic. rolling back for now.

PR #2641, opened 2025-02-24
add nav role to prev next buttons

### LINKED ISSUE #2598: prev next aria landmark
Based on what I can tell this needs a role. 
Image

dnoneill:
closes #2598


app/components/media/prev_next_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="prev-next-container">
+<div class="prev-next-container" role="navigation">
```
jcoyne, line 1: Can we just make this a `<nav>` node instead?  Either way, I think this needs an aria-label as there is more than one navigation role on the page.

jcoyne (changes requested):

jcoyne (approved):

PR #2649, opened 2025-02-24
add banner for non download 3d

### LINKED ISSUE #2635: 3d viewer: when download: none, show a banner and the lock icon
The 3d viewer will not render an item when `download: none` regardless of what setting is used for the `view` rights. Similar to PDF, we should show a lock icon and a banner indicating the item is not available online when these rights are applied.

This what the PDF viewer looks like in that case. 3d should look the same.

Image

There are 4 items as test cases on stage:

- [world, none](https://argo-stage.stanford.edu/view/druid:xj417nj6220)
- [stanford, none](https://argo-stage.stanford.edu/view/druid:ps577gx9734)
- [location, none](https://argo-stage.stanford.edu/view/druid:qh368gn2894)
- [citation-only](https://argo-stage.stanford.edu/view/druid:zz953dm4408)

As with the Stanford-login case (#2634), there is a specific file that I think determines whether the viewer will render an object.

The 3D viewer looks for:

- A "resource type" of 3d
- A file with mimetype model/gltf-binary within that resource

Other files may or may not be downloadable, but that file should be what determines whether to show a lock icon.

dnoneill:
closes #2635 
Screenshot 2025-02-24 at 3 58 24 PM

### DIFF
--- a/app/assets/stylesheets/companion_window.css
+++ b/app/assets/stylesheets/companion_window.css
@@ -350,9 +350,16 @@ body {
     }
   }
 
+
+  [role="banner"]:not([hidden])
+  + .companion-window-component-body {
+    --auth-bar-height: 40px;
+  }
+
   .companion-window-component-body {
+    --auth-bar-height: 0px;
     --body-height: calc(
-      100vh - var(--header-size) - 5px
+      100vh - var(--header-size) - var(--auth-bar-height) - 5px
     ); /* the 5px is for the outer border and the top red border */
     display: flex;
     flex-grow: 1;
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/companion_window.css
```diff
@@ -101,6 +101,11 @@ body {
     display: none !important;
   }
 
+  .locked-status img {
```
jcoyne, line 104: Shouldn't this go in the `model.css`?

dnoneill: I considered it but I think it makes sense to have it be consistent across viewers. (PDF also uses it)

jcoyne: How was the image for the PDF formatted prior to this change?

dnoneill: Ah. It was in it's own file. I will go ahead an move that into the 3d 
```@import url("locked_status.css");```

jcoyne (approved):

PR #2653, opened 2025-02-25
Set height with relative units and new default height

### LINKED ISSUE #2623: Set default minimum viewer height
Follow-up to #2593.
Set default minimum height for all viewers to 500px.

### LINKED ISSUE #2624: Use relative units for viewer height
Follow-up to #2593 and #2623.
Instead of absolute pixel units, use relative units corresponding to 500px.

From Slack: _We already have the functionality for setting pixel units. We need to add functionality for passing in relative units (in addition to maintaining existing pixel functionality) and for a new default height (specified in relative units)._

jcoyne:
Fixes #2624
Fixes #2623

### DIFF
--- a/app/viewers/embed/viewer/geo.rb
+++ b/app/viewers/embed/viewer/geo.rb
@@ -67,7 +67,7 @@ module Embed
       end
 
       def default_height
-        493
+        new_viewer? ? super : '493px'
       end
     end
   end
--- a/app/viewers/embed/viewer/model_viewer.rb
+++ b/app/viewers/embed/viewer/model_viewer.rb
@@ -31,7 +31,7 @@ module Embed
       private
 
       def default_height
-        513
+        new_viewer? ? super : '513px'
       end
     end
   end
--- a/app/viewers/embed/viewer/was_seed.rb
+++ b/app/viewers/embed/viewer/was_seed.rb
@@ -58,9 +58,10 @@ module Embed
       private
 
       def default_height
-        return 340 if embed_request.hide_title?
+        return super if new_viewer?
+        return '340px' if embed_request.hide_title?
 
-        420
+        '420px'
       end
     end
   end
--- a/spec/components/iframe_component_spec.rb
+++ b/spec/components/iframe_component_spec.rb
@@ -14,25 +14,25 @@ RSpec.describe IframeComponent, type: :component do
   let(:viewer) do
     instance_double(
       Embed::Viewer::CommonViewer,
-      height: '555',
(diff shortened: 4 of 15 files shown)


app/viewers/embed/viewer/geo.rb
```diff
@@ -67,7 +67,7 @@ def index_map_files
       end
 
       def default_height
-        493
+        new_viewer? ? super : 493
```
corylown, line 70: I think `default_height` now needs to return a height string with units since `px` is no longer getting added in IframeComponent#height_style. It doesn't look like there are any tests that would have caught this for the geo viewer as there are for the file viewer.
```suggestion
        new_viewer? ? super : '493px'
```

app/viewers/embed/viewer/model_viewer.rb
```diff
@@ -30,7 +30,7 @@ def self.show_download?
       private
 
       def default_height
-        513
+        new_viewer? ? super : 513
```
corylown, line 33:
```suggestion
        new_viewer? ? super : '513px'
```

app/viewers/embed/viewer/was_seed.rb
```diff
@@ -58,6 +58,7 @@ def archived_site_url
       private
 
       def default_height
+        return super if new_viewer?
         return 340 if embed_request.hide_title?
```
corylown, line 62:
```suggestion
        return '340px' if embed_request.hide_title?
```

spec/components/iframe_component_spec.rb
```diff
@@ -14,25 +14,25 @@
...
+      width: '666px',
       embed_request:,
       iframe_title: 'Hello world',
       purl_object:
     )
   end
 
-  describe 'the height' do
-    subject { iframe['height'] }
+  describe 'the stytle' do
```
corylown, line 25:
```suggestion
  describe 'the style' do
```

corylown (changes requested):

app/viewers/embed/viewer/geo.rb
dnoneill: Except I think this will cause a problem for this comparison https://github.com/sul-dlss/sul-embed/pull/2653/files#diff-8f6c5a40f063a49b3d804a2dd5dd2bd828d879504714ed9ab9edb53204338ea9R20. I think you need to do 
```        return "#{default_height}px" if @embed_request.maxheight.to_i > default_height.to_i
```

PR #2658, opened 2025-02-25
add missing return

dnoneill:


jcoyne (changes requested):
Can you add a test for this?

PR #2682, opened 2025-02-26
align text center file viewer

### LINKED ISSUE #2533: File icons appear smaller and off-center

dnoneill:
Screenshot 2025-02-27 at 3 53 02 PM

closes #2533

### DIFF
--- a/app/assets/stylesheets/file.css
+++ b/app/assets/stylesheets/file.css
@@ -109,6 +109,7 @@
       width: 1.5rem;
       display: inline-block;
       vertical-align: middle;
+      flex-shrink: 0;
     }
     font-weight: bold;
   }
@@ -123,9 +124,11 @@
     align-items: flex-start;
     svg {
       margin-left: 0.25rem;
+      flex-shrink: 0;
     }
     .foldername {
       margin-left: 0.25rem;
+      align-self: center;
     }
   }
 
@@ -134,7 +137,6 @@
     font-size: 1rem;
 
     svg {
-      vertical-align: middle;
       fill: var(--button-color);
       width: 1.5rem;
       height: 1.5rem;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/file.css
```diff
@@ -120,7 +120,7 @@
 
   .folder-title {
     display: flex;
-    align-items: flex-start;
+    align-items: center;
```
jcoyne, line 123: Doesn't this mess with the alignment when the folder name wraps to two lines?

jcoyne (changes requested):

app/assets/stylesheets/file.css
dnoneill: I could be wrong but I don't think we can have both. If that is the expected behavior then all folder/files need to align with the top of the icon and not the middle.

jcoyne: I think this problem comes from having the icons bigger than the text.  Can we add additional padding on the top of the text and keep `flex-start`?

alundgard: I wonder if using `margin-top: auto` and `margin-bottom: auto` on the folder and file text (instead of changing align-items) would help with the vertical centering?

Example
margin-top-bottom

Example
margin-top-bottom-2

dnoneill: It sort of fixes things but it still looks weird. 
Screenshot 2025-02-27 at 3 30 19 PM

dnoneill: I think I figured it out. 
Screenshot 2025-02-27 at 3 46 54 PM

dnoneill: @jcoyne I have approval for the new view from @alundgard. This just needs a dev review.

jcoyne (approved):

PR #2683, opened 2025-02-26
[refactor] Extract MediaFile as subclass of ResourceFile

jcoyne:
Because media files have attributes and behaviors that other file types don't have, it makes sense to extract its own class


app/models/embed/purl/media_file.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Embed
+  class Purl
+    # Any file on a media object, may be a mp4 or a caption file.
+    class MediaFile < ResourceFile
+      attr_accessor :sdr_generated, :language
+      alias sdr_generated? sdr_generated
```
dnoneill, line 8: Can we note that sdr_generated is not a media only field, but we are currently only using it for media. I would imagine eventually we are going to want to use this field for PDFs at some point.

jcoyne: The representations we are using here are not the same as the representations in the Cocina.  This model is just for "as we use it in the context of sul-embed".  That said, I'd feel comfortable putting a note in the JSON deserializer as that does deal with Cocina.

jcoyne: https://github.com/sul-dlss/sul-embed/pull/2683/files#diff-29d922aaac25053d3f87aa4cb3eafedf452ee03342ec6deb21ccd4b962eb6bbfR71

corylown (approved):

PR #2684, opened 2025-02-26
don't user fullscreen stimulus on iphone/ipad

### LINKED ISSUE #2680: [sul-embed/prod] Error: Missing target element "button" for "fullscreen" controller
## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49295/faults/102906549)

dnoneill:
closes #2680 

https://github.com/sul-dlss/sul-embed/pull/2685 also works we should only merge one of these. The question which is the better solution?


app/components/file_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= render ::CompanionWindowsComponent.new(viewer:, stimulus_controller: 'file-auth tree fullscreen') do |component| %>
```
jcoyne, line 1: Was this change intentional?

jcoyne (changes requested):

app/components/companion_windows_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="sul-embed-container companion-window-component" id='sul-embed-object' hidden data-controller="companion-window iiif-manifest-loader fullscreen <%= @stimulus_controller %>" data-iiif-manifest-loader-iiif-manifest-value="<%= iiif_v3_manifest_url %>" data-fullscreen-target="area" data-fullscreen-close-value="Exit full screen">
```
jcoyne, line 1: Can you move the first part into the class file?  Just append it into `@stimulus_controller`.  Perhaps the second part could also be extracted into a method like `fullscreen_options`?

jcoyne (changes requested):
I prefer this method.

app/components/companion_windows_component.html.erb
dnoneill: I went ahead and did this along those lines. i wasn't able to figure out how to do fullscreen_options without having weird escaping that probably would have required html_safe to get it to render correctly

jcoyne (approved):

PR #2685, opened 2025-02-26
hide button instead of not rendering

### LINKED ISSUE #2680: [sul-embed/prod] Error: Missing target element "button" for "fullscreen" controller

dnoneill:
closes #2680 
https://github.com/sul-dlss/sul-embed/pull/2684 also works we should only merge one of these. The question which is the better solution?


dnoneill:
closing in favor of https://github.com/sul-dlss/sul-embed/pull/2684

PR #2694, opened 2025-02-27
Use iiif v2 auth for media

### LINKED ISSUE #2691: Front end for media should use IIIF auth service

jcoyne:
Test with
* https://sul-purl-stage.stanford.edu/vv407rn1804
* https://sul-purl-stage.stanford.edu/bn208cc3480
* https://sul-purl-stage.stanford.edu/bq834ky5156

Fixes #2691


app/javascript/controllers/file_auth_controller.js
```diff
@@ -266,4 +270,12 @@ export default class extends Controller {
     return json.status == '401' && 'heading' in json && 'en' in json.heading && json.heading.en.length
       && json.heading.en[0].startsWith('Access is restricted to the')
   }
+
+  // Checks the result of the probe auth request to see if access is restricted to location
+  // This code depends on the text returned by probe service, so changes to the heading
+  // should be reflected here as well.
+  isNoAccess(json) {
+    return json.status == '401' && 'heading' in json && 'en' in json.heading && json.heading.en.length
```
dnoneill, line 278: If we get https://github.com/sul-dlss/stacks/pull/1405 and https://github.com/sul-dlss/sul-embed/pull/2805 in this will be much easier and already done. Also noAccess below will not be necessary

jcoyne: Yes, I noticed your work in stacks.

jcoyne:
Currently if you test on https://sul-purl-stage.stanford.edu/vv407rn1804, and switch to the second item, the "This item cannot be accessed online" message disappears.

app/components/media/tag_component.rb
```diff
@@ -58,10 +58,10 @@ def media_element # rubocop:disable Metrics/MethodLength
...
-            auth_url: authentication_url,
-            media_tag_target: 'authorizeableResource',
+            index: @resource_iteration.index,
             controller: 'media-player',
-            action: 'media-seek@window->media-player#seek ' \
+            action: 'iiif-manifest-received@window->file-auth#parseFiles ' \
+                    'media-seek@window->media-player#seek ' \
                     'fullscreenchange@window->media-player#fullscreenChange ' \
                     'auth-success@window->media-player#initializeVideoJSPlayer'
           }
```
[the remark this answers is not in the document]
jcoyne: All media are no_download, because you stream them, and don't download.  What URL were you looking at?

dnoneill (changes requested):
- Banner isn't showing up on the image, but it does in our current implementation. https://sul-purl-stage.stanford.edu/fj472zf9945
Screenshot 2025-03-21 at 11 17 26 AM
- Whenever I log in with any stanford item I get an authentication error alert but if I refresh and click log in again it goes away: https://sul-purl-stage.stanford.edu/bn208cc3480
- The second item in a stanford only media item never seems to work: https://sul-purl-stage.stanford.edu/bn208cc3480
- I am getting a cardinal key login popup whenever I go to a stanford only item. This is before the page is even loaded. I am wondering if it is happening because the a/v material is half loading. I am not quite sure but it might have to do with the two errors above.

app/components/media/tag_component.rb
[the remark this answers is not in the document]
dnoneill: Ah. I was looking at  https://sul-purl-stage.stanford.edu/vv407rn1804. It is only happening locally I think.

dnoneill:
@jcoyne I found another bug. The prev/next buttons error out on the media. This is because the buttons do not have fileUri on them and sendThubmnailClickedEvent requires that field.

dnoneill (changes requested):
1. Screenshot 2025-03-25 at 10 35 10 AM are causing a bug in the console log.
2. The captions are loading really slowly. I think it doesn't actually load until you click a thumbnail. 

https://github.com/user-attachments/assets/1d481290-1fc8-447a-abea-48dd11e89f6e

app/javascript/controllers/prev_next_controller.js
```diff
@@ -4,6 +4,8 @@ export default class extends Controller {
   static outlets = [ "content-list" ]
 
   prevNextMedia(evt) {
-    this.contentListOutlet.showMedia(evt);
+    const index = evt.currentTarget.dataset.prevNextIndexParam
+    const thumbnail = this.contentListOutlet.element.querySelector(`[data-content-list-index-param="${index}"]`)
+    thumbnail.click()
```
dnoneill, line 9: This is kind of hacky basically you are going into the sidebar and clicking the next item. Can you add a comment about why we are doing things this way instead of something less strange.

jcoyne: We need to have the fileUri in the event details.  The content list on already has the fileUri as a data parameter.  So we either do this or we need to put a duplicate list of file uris on the page. I'm not sure how we want to structure that as a comment.

dnoneill: Maybe something along the lines of, making this work with showMedia would require adding duplicate data parameters so instead of duplicating the information we trigger the event in one place. Something along those lines?

jcoyne: "making this work" doesn't make sense, because it does work just as it is.

jcoyne: "making this work the same way it used to work" doesn't seem like a helpful comment.

dnoneill: @jcoyne I think we need a comment that makes sure when someone comes upon this in 1 year and is like why are we doing things this way we have an answer. I am worried someone is going to see this, think it is inefficient or not good and change it to something else that breaks things.

dnoneill (approved):

PR #2702, opened 2025-02-27
add location restricted banner for 3d, fix downloadable files

### LINKED ISSUE #2697: 3D viewer: show a location restriction banner when an item can be viewed and downloaded at a location
When a 3D item is location-restricted and can be viewed and downloaded at that location, show the appropriate location-restriction banner. This should follow the same pattern as location-restricted PDFs in the document viewer (https://github.com/sul-dlss/sul-embed/pull/2630).

Currently, the 3D viewer is showing the banner for "no access" based on https://github.com/sul-dlss/sul-embed/issues/2635 but that banner should only apply when "download = none".

dnoneill:
closes #2697 

Working on tests but the logic could be double checked

Screenshot 2025-02-27 at 5 39 03 PM


app/models/embed/purl/resource_file.rb
```diff
@@ -58,6 +58,10 @@ def downloadable?
           NON_DOWNLOADABLE_ROLES.exclude?(role)
       end
 
+      def viewable?
```
jcoyne, line 61: Can we change the name of this?  The name `viewable?` implies  "viewable to the current user" to me.  In this case it's more like "Is not no download". I know that's a double negative, but we talk a lot about "no-download" as a specific access level.

dnoneill: sounds good. Updated.

spec/lib/embed/viewer/model_viewer_spec.rb
```diff
@@ -23,4 +27,43 @@
       expect(viewer.three_dimensional_files.first).to match(%r{/file/druid:abc123/obj1\.glb$})
     end
   end
+
+  describe '#three_dimensional_files with no download' do
```
jcoyne, line 31: I'd put the `with no download` in a `context` block.

app/viewers/embed/viewer/common_viewer.rb
```diff
@@ -112,6 +112,15 @@ def iframe_title
         I18n.t('title', default: 'Viewer', scope: i18n_path, title: purl_object.title)
       end
 
+      def message
+        if @purl_object.location_restriction
+          return { type: 'location-restricted',
+                   message: I18n.t('restrictions.restricted_access', location: @purl_object.restricted_location) }
+        end
+
+        false
```
jcoyne, line 121:
```suggestion
        { type: 'location-restricted',
                   message: I18n.t('restrictions.restricted_access', location: @purl_object.restricted_location) } if @purl_object.location_restriction
```

dnoneill: I am pretty sure this is how I originally wrote this and rubocop corrected to this. I can double check.

dnoneill: Yup. Rubocop complained. It is back to the way it was. ```Style/MultilineIfModifier: Favor a normal if-statement over a modifier clause in a multiline statement```

dnoneill:
@jcoyne with the rebase i was able to change not_no_download to look at the download attribute

```
def not_no_download?
        download != 'none'
      end
 ```

File: app/models/embed/purl/resource_file.rb [hunk 4]
```diff
@@ -73,6 +73,10 @@ def stanford_only_downloadable?
         download == 'stanford'
       end
 
+      def not_no_download?
```
jcoyne, line 76: Since `no_download?` already exists, can you call that one instead (e.g. `!no_download?`) and remove this method?

jcoyne (changes requested):

app/models/embed/purl/resource_file.rb
dnoneill: done

spec/lib/embed/viewer/file_spec.rb
```diff
@@ -48,6 +48,40 @@
     end
   end
 
+  describe 'message' do
```
jcoyne, line 51: Should these tests go into authorization_spec.rb?  It seems like these tests are about the logic in Authorization.

spec/lib/embed/viewer/model_viewer_spec.rb
```diff
@@ -8,19 +8,66 @@
...
+        ]
+      end
+
+      it 'returns no files' do
+        expect(viewer.three_dimensional_files.length).to eq 0
+      end
+    end
+  end
+
+  describe 'message' do
```
jcoyne, line 48: Should these tests move to authorization_spec.rb?. They seem to be about that class, not ModelViewer.

dnoneill: Yeah sorry I wrote those before we moved everything to the authorization model

dnoneill: Also based on our conversation on banners I think I need to go look at doing this in the file_auth component and not using this. Otherwise we are going to have to undo this.

PR #2703, opened 2025-02-27
[refactor] Extract tabpanel component

jcoyne:


app/components/tabpanel_component.rb
```diff
@@ -0,0 +1,19 @@
...
+    @hidden = hidden
+    @title = title
+    @data = data
+  end
+
+  attr_reader :id, :title, :data
+
+  def hidden?
+    @hidden
+  end
```
dnoneill, line 18: Is there a reason hidden isn't added the attr_reader?

jcoyne: @dnoneill because we want a method ending in a question mark.

dnoneill: Why? Also can we use an alias if we really want the method to end in a question mark.

jcoyne: It's ideomatic ruby to use methods ending with a question mark to indicate boolean return value.  I don't want to have a method defined (by attr_accessor) that we don't need.

dnoneill: ok

dnoneill (approved):

Issue #2668, opened 2025-02-26
Fallback location restriction message is not grammatically correct


andrewjbtw:
I supplied fallback language for the location-restriction banner without realizing that the text:

> Access is restricted to the

is not part of the editable message.

https://github.com/sul-dlss/sul-embed/blob/961dc7fefb8c0739e1e2c888fe5802cb306094d5/app/models/embed/purl_json_loader.rb#L62

This results in the grammatically incorrect:

> Access is restricted to the site visitors to the Stanford libraries

This language should be revised to be grammatical. Leaving in the backlog for now as this is not a blocker and location-specific messages appear correctly.

PR #2719, opened 2025-03-03
Remove unused message

jcoyne:
I don't see this message used anywhere in the designs:
https://www.figma.com/design/LhMpFAHNwF6tPo53p2xzyz/Embed-Viewer-Components?node-id=0-1&p=f&t=xA1yRguMvgTUXyHk-0

### DIFF
--- a/app/models/embed/purl_json_loader.rb
+++ b/app/models/embed/purl_json_loader.rb
@@ -59,10 +59,10 @@ module Embed
     end
 
     def restricted_location
-      fallback_message = 'site visitors to the Stanford Libraries'
-      return Settings.locations[json.dig('access', 'location')] || fallback_message if location_restriction
+      return unless location_restriction
 
-      fallback_message
+      locations = Settings.locations
+      locations[json.dig('access', 'location')] || locations[:fallback]
     end
 
     def download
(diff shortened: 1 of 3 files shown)


dnoneill:
@jcoyne This was requested on the PR. https://github.com/sul-dlss/sul-embed/pull/2516

Screenshot 2025-03-03 at 2 30 47 PM

jcoyne:
@dnoneill okay, but that is only needed if there is a location_restriction then.  Currently we're returning the default message regardless of the presence of a location restriction.

dnoneill:
@jcoyne That is fine. I had the fallback without checking the restricted location because I was a little worried about https://github.com/sul-dlss/sul-embed/blob/81ebdaa55916e9eb3591277ca44a1053c5bea6c2/app/components/media_component.html.erb#L51. I wasn't sure if it might cause problems here. It shouldn't display incorrectly and the purl should be right but I wasn't sure if it was going to cause a problem for the probe service displaying nil because we aren't checking on the file level.

andrewjbtw:
I filed an issue to correct the grammar but haven't come up with new language yet: https://github.com/sul-dlss/sul-embed/issues/2668

dnoneill:
@jcoyne This looks good. Can we add a test for this? I could swear there was one but I can't find it.

dnoneill (approved):

PR #2731, opened 2025-03-04
Only authorize files as they are clicked on

jcoyne:


dnoneill:
@jcoyne This is using the probe service every time you clicked even if you already did. I think we want to no probe again if the user has already done auth.

dnoneill:
For example, if I choose the  00 Woodwint Quintet 2 - Score.pdf and then the  01 Woodwint Quintet 2 - Score.pdf and then 00 Woodwint Quintet 2 - Score.pdf again. 00 Woodwint Quintet 2 - Score.pdf  will query the auth service twice.

jcoyne:
@dnoneill I believe we have to re-authorize because we're going to refetch the resource. If the token has timed out, we need to get a new one.

dnoneill:
@jcoyne I don't think the token would have expired that quickly. Also it should already be rendered on the page. I will double check but I am pretty sure the PDF viewer is already rendered we are just hiding it. Let me double check

jcoyne:
@dnoneill you can keep the player open for days, then click another thumbnail.  We aren't tracking expiry of the token. It's just simpler to re-check. The extra call adds very little overhead.

app/javascript/controllers/file_auth_controller.js
```diff
@@ -32,18 +32,28 @@ export default class extends Controller {
...
         const paintingAnnotations = annotationPage.items.filter((annotation) => annotation.motivation === "painting")
         return paintingAnnotations.map((annotation) => {
           const contentResource = annotation.body
-          this.maybeDrawContentResource(contentResource)
           return contentResource
         })
       })
     })
-    this.firstFile = resources[0]['id']
+    console.log("Resources: ", this.documents)
```
dnoneill, line 45: Is this console statement here for a specific reason

app/javascript/controllers/content_list_controller.js
```diff
@@ -30,7 +30,9 @@ export default class extends Controller {
   showPdf(evt) {
     evt.preventDefault()
     const fileUri = evt.currentTarget.dataset.url
-    window.dispatchEvent(new CustomEvent('auth-success', { detail: fileUri }))
+    // window.dispatchEvent(new CustomEvent('auth-success', { detail: fileUri }))
```
dnoneill, line 33: Do we want to keep this or should be remove it?

app/javascript/controllers/file_auth_controller.js
jcoyne: Just for understanding. I can remove it.

app/javascript/controllers/content_list_controller.js
jcoyne: I will remove.

app/javascript/controllers/file_auth_controller.js
```diff
@@ -32,18 +32,28 @@ export default class extends Controller {
...
+    this.maybeDrawContentResource(this.documents[0]) // cause login to first resource
+    this.firstFile = this.documents[0]['id']
```
dnoneill, line 47: I am not sure the firstFile is needed anymore. This was for the line ```if (this.resources[messageId].contentResourceId !== this.firstFile) return``` in loginNeeded. This was needed because this used to loop through every file on render and authorize them. since we are no longer doing this I don't think we need this or the line if (this.resources[messageId].contentResourceId !== this.firstFile) return.

jcoyne: You are right. I removed it.

hudajkhan:
I can see it kicking off a probe service controller request for the first  PDF in the list when the page loads and then each time I click on the a separate PDF (test is this: https://purl.stanford.edu/bf427nq6311) .  Works as expected.

hudajkhan (approved):

PR #2748, opened 2025-03-05
show pdf location restriction banner per file

### LINKED ISSUE #2360: Location-restricted PDFs do not display in the PDF viewer even at an authorized location

hudajkhan:
Closes #2360  

What this pull request does:

- Sets up a location restricted banner that is hidden when the page loads
- Changes the logic for when the viewer is available. It is now true unless the status for the first file is "not downloadable"
- Sets up  the file auth controller to check the location restriction message from stacks, and then show the location restriction banner and update the message to be the one returned from stacks

Note: This should work in staging, but the stacks changes will need to be pushed to production first for these embed changes to work in production

### DIFF
--- a/app/components/pdf_component.html.erb
+++ b/app/components/pdf_component.html.erb
@@ -9,6 +9,7 @@
 
   <% component.with_authorization_messages do %>
     <%= render LoginComponent.new %>
+    <%= render LocationRestrictionComponent.new %>
     <%= render CompanionWindows::ContentNotAvailableComponent.new unless viewer.available? %>
   <% end %>
 
--- a/app/javascript/controllers/file_auth_controller.js
+++ b/app/javascript/controllers/file_auth_controller.js
@@ -286,22 +286,11 @@ export default class extends Controller {
   // This code depends on the text returned by probe service, so changes to the heading
   // should be reflected here as well.
   isLocationRestricted(json) {
-    if(json.status == '401' && 'heading' in json && 'en' in json.heading && json.heading.en.length
-      && json.heading.en[0].startsWith('Content is restricted to location'))
-      return true
-    return false
-  }
-
-  // Extract just the location from the message
-  extractRestrictedLocation(accessMessage) {
-    const probeServicePrefix = 'Content is restricted to location '
-    if(accessMessage != null && accessMessage.startsWith(probeServicePrefix))
-      return accessMessage.substring(probeServicePrefix.length, accessMessage.length)
-
-    return 'site visitors to the Stanford libraries'
+    return json.status == '401' && 'heading' in json && 'en' in json.heading && json.heading.en.length
+      && json.heading.en[0].startsWith('Access is restricted to the')
   }
 
   retrieveRestrictedLocationMessage(json) {
-    return 'Access is restricted to the ' + this.extractRestrictedLocation(json.heading.en[0]) + '. See Access conditions for more information.'
+    return json.heading.en[0]
   }
 }
--- a/app/javascript/controllers/file_auth_controller.js
+++ b/app/javascript/controllers/file_auth_controller.js
@@ -1,7 +1,7 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
-  static targets = ["container", "loginPanel", "messagePanel", "loginButton", "loginMessage"]
+  static targets = ["container", "loginPanel", "messagePanel", "loginButton", "loginMessage", "locationRestriction", "locationRestrictionMessage"]
 
   resources = {} // Hash of messageIds to resources
 
@@ -45,6 +45,7 @@ export default class extends Controller {
   }
 
   // Triggered when clicking on a thumbnail
+  // Leads to authorization check of the file and displays the correct access banner or renders viewer
   authFileAndDisplay(event) {
     const document = this.documents.find((document) => document.id == event.detail.fileUri)
     if (document)
@@ -56,6 +57,8 @@ export default class extends Controller {
   // Try to render the resource, checks for any required authorization and shows login window if needed
   maybeDrawContentResource(contentResource) {
     console.debug("Now figure out if we can render", contentResource)
+    // Ensure location restriction banner is hidden by default in case it was visible for the previous document 
+    this.locationRestrictionTarget.hidden = true
     if (!contentResource.service) {
       // no auth service is present, just render the resource
       this.renderViewer({ fileUri: contentResource.id })
@@ -135,6 +138,10 @@ export default class extends Controller {
     this.queryProbeService(messageId)
       .then((result) => this.renderViewer(result))
       .catch((json) => {
+        if(this.isLocationRestricted(json)) {
+          this.handleLocationRestricted(json, accessService)
+          return
+        }
         console.debug("Probe failed or access denied/restricted", json)
         // Check if non-expired token already exists in local storage,
         // and if it exists, query probe service with it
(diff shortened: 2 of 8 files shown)


app/javascript/controllers/file_auth_controller.js
```diff
@@ -264,4 +271,42 @@ export default class extends Controller {
...
+  }
+
+  // Checks the result of the probe auth request to see if access is restricted to location
+  // This code depends on the text returned by probe service, so changes to the heading
+  // should be reflected here as well.
+  isLocationRestricted(json) {
+    if(json.status == '401' && 'heading' in json && 'en' in json.heading && json.heading.en.length
+      && json.heading.en[0].startsWith('Content is restricted to location'))
+      return true
+    return false
```
jcoyne, line 297: can we do:
```suggestion
    return json.status == '401' && 'heading' in json && 'en' in json.heading && json.heading.en.length
      && json.heading.en[0].startsWith('Content is restricted to location')
```

jcoyne: I'm not sure we need to check `json.heading.en` or `json.heading.en.length` as we know our Stacks response will have these.

app/javascript/controllers/file_auth_controller.js
```diff
@@ -264,4 +271,42 @@ export default class extends Controller {
...
+  extractRestrictedLocation(accessMessage) {
+    const probeServicePrefix = 'Content is restricted to location '
+    if(accessMessage != null && accessMessage.startsWith(probeServicePrefix))
+      return accessMessage.substring(probeServicePrefix.length, accessMessage.length)
+
+    return 'site visitors to the Stanford libraries'
+  }
+
+  retrieveRestrictedLocationMessage(json) {
+    return 'Access is restricted to the ' + this.extractRestrictedLocation(json.heading.en[0]) + '. See Access conditions for more information.'
```
jcoyne, line 310: Can this whole first sentence move to the IIIF Auth response heading in Stacks?

app/viewers/embed/viewer/document_viewer.rb
```diff
@@ -27,15 +27,11 @@ def fullscreen?
...
-
-      # this indicates if the PDF is downloadable (though it could be stanford only)
-      # Stanford only and location restrictions are handled via a separate authorization flow,
-      # since it is possible for people to do something about the restriction
+      # This indicates if the first PDF is downloadable (though it could be stanford only)
+      # as well as location restricted.  Authorization for other documents will be
+      # checked as the user clicks on items in the content side bar.
       def available?
-        document_resource_files.first&.downloadable?
+        document_resource_files.first&.downloadable? || document_resource_files.first&.location_restricted?
```
jcoyne, line 34: I'm not sure why we're still checking this on the backend, since it looks like the frontend is also doing this check.

dnoneill: I think we keep doing as a patch because I don't think haven't added the JS to display this. this it should be `!document_resource_files.first&.no_download?` but yes I think we should be fixing this in another PR to the file_auth

app/javascript/controllers/file_auth_controller.js
hudajkhan: I am working on that in a stacks PR, so I agree that would be better

app/viewers/embed/viewer/document_viewer.rb
hudajkhan: I will leave a comment here to explain why. The code puts up a generic "content not available" block if it thinks the item is not available, hence this change to ensure an object that is not downloadable but is location restricted would not immediately display the content not available block.  @dnoneill I can try your suggestion if it covers all the cases, but that's the main motivation (to not show the content not available block)

app/javascript/controllers/file_auth_controller.js
dnoneill: For now, so we can get this in because it is a blocker, can we just return what we are getting from stacks since it is very close to what we want and we can have the language update done after this is in?

app/javascript/controllers/file_auth_controller.js
hudajkhan: yes to the first comment.  For the second comment, this is more to ensure if something weird happens on the response end, things don't just stop working.

app/viewers/embed/viewer/document_viewer.rb
hudajkhan: I will use that logic then, thank you!

app/javascript/controllers/file_auth_controller.js
hudajkhan: The other PR is merged so we should be good to go with the updated strings

app/components/pdf_component.html.erb
```diff
@@ -9,6 +9,7 @@
 
   <% component.with_authorization_messages do %>
     <%= render LoginComponent.new %>
+    <%= render LocationRestrictionComponent.new %>
```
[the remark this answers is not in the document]
hudajkhan: We just talked and agreed we can wait on that change until later, but that sounds like a good idea for subsequent work.

app/javascript/controllers/file_auth_controller.js
```diff
@@ -259,4 +266,31 @@ export default class extends Controller {
...
     this.messagePanelTarget.hidden = true
   }
+
+  // To see if item is restricted by location, check the probe service json response
+  handleLocationRestricted(json, accessService) {
+    // The probe auth service is called for each file separately.
+    // If the location restriction target is available, then trigger auth denied message.
+    // The event will lead to the locked icon being displayed for this item
+    // This allows the lock window to show
+    const event = new CustomEvent('auth-denied', { accessService: accessService })
```
jcoyne, line 276: I think we should stuff the location restricted message in this event.  Then we can have a controller devoted to displaying these messages, and this controller would not have to deal with this

jcoyne: This can be a follow-on PR though.

app/viewers/embed/viewer/document_viewer.rb
```diff
@@ -27,15 +27,11 @@ def fullscreen?
...
-
-      # this indicates if the PDF is downloadable (though it could be stanford only)
-      # Stanford only and location restrictions are handled via a separate authorization flow,
-      # since it is possible for people to do something about the restriction
+      # This indicates if the first PDF is downloadable (though it could be stanford only)
+      # as well as location restricted.  Authorization for other documents will be
+      # checked as the user clicks on items in the content side bar.
       def available?
-        document_resource_files.first&.downloadable?
+        !document_resource_files.first&.no_download?
```
jcoyne, line 34: I'm still not clear on why we need this at all.  Can't we just run this all through the IIIF auth?

hudajkhan: The PDF Component html will not show the viewer at all if the "viewer.available?" is false.  Instead, it will just show a "content not available" section.  This change ensures the content not available section is not displayed unless the first file is a "no download" option. If we want to change that logic, I'd suggest tackling that in a different issue because there is probably more to streamline here (or to work out).

app/javascript/controllers/file_auth_controller.js
hudajkhan: I think that approach makes sense and it would be fine to do in a follow up PR. @dnoneill and I were also discussing the other cases (embargo etc.) so it makes sense to have that logic in its own event/controller framework.

jcoyne (approved):

PR #2749, opened 2025-03-05
add check to see if released into earthworks

### LINKED ISSUE #2288: Link to earthworks appears in footer for geo items that are not in earthworks (i.e., link is broken)
### Expected behavior

For geo items, sul-embed includes a link in the footer that takes people to the item in EarthWorks. This link should only appear when it works, that is, when the item is in EarthWorks and the URL resolves.

Screenshot 2025-01-14 at 12 27 28 PM

### Actual behavior

The link to EarthWorks appears in embed is always present for geo items, even items that are not in EarthWorks. This results in broken links for geo items not released to EarthWorks.

1. Go to a geo item that has a Purl but has not been released to EarthWorks. [[example](https://purl.stanford.edu/gh138gt7402)]
2. Click on the link to EarthWorks
3. See a 404

[This Argo search](https://argo.stanford.edu/catalog?f%5Bcontent_type_ssim%5D%5B%5D=geo&f%5Bprocessing_status_text_ssi%5D%5B%5D=Accessioned&f%5Breleased_to_earthworks%5D%5B%5D=never) should turn up all items with this issue: geo, accessioned, not released to EarthWorks.

### Browser / Environment

All of them, probably.

dnoneill:
closes #2288

### DIFF
--- a/app/components/companion_windows/external_link_component.rb
+++ b/app/components/companion_windows/external_link_component.rb
@@ -10,5 +10,11 @@ module CompanionWindows
     attr_reader :viewer, :label
 
     delegate :external_url, to: :viewer
+
+    def stimulus_attributes
+      return unless viewer.instance_of?(::Embed::Viewer::Geo)
+
+      { controller: 'meta-check', meta_check_url_value: viewer.purl_object.meta_json_url }
+    end
   end
 end
--- a/app/components/companion_windows/external_link_component.html.erb
+++ b/app/components/companion_windows/external_link_component.html.erb
@@ -1,6 +1,6 @@
-<a class="modal-component-button" href="<%= external_url %>" target='_parent'>
+<%= link_to external_url, class: 'modal-component-button', target: '_parent', data: stimulus_attributes  do %>
   <svg class="MuiSvgIcon-root modal-component-icon LaunchSharp" focusable="false" aria-hidden="true" viewBox="0 0 24 24" aria-label="fontSize medium">
     <path d="M19 19H5V5h7V3H3v18h18v-9h-2zM14 3v2h3.59l-9.83 9.83 1.41 1.41L19 6.41V10h2V3z"></path>
   </svg>
   <span class="modal-component-label"><%= label %></span>
-</a>
+<% end %>
--- a/app/javascript/controllers/meta_check_controller.js
+++ b/app/javascript/controllers/meta_check_controller.js
@@ -1,0 +1,17 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = {
+    url: String
+  }
+
+  connect() {
+    fetch(this.urlValue)
+      .then((response) => response.json())
+      .then((json) => this.hideLink(json))
+  }
+
+  hideLink(meta_json) {
+    if (!meta_json['earthworks']) this.element.remove();
+  } 
+}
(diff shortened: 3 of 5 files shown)


jcoyne (commented):
Have you considered doing this on the front end instead?  This one request could block the rendering of the whole viewer. If we did it on the front end, the viewer would render right away and only the link would be delayed.

app/components/companion_windows/external_link_component.rb
```diff
@@ -10,5 +10,11 @@ def initialize(viewer:, label:)
     attr_reader :viewer, :label
 
     delegate :external_url, to: :viewer
+
+    def stimulus_controllers
+      return '' unless viewer.instance_of?(::Embed::Viewer::Geo)
+
+      "data-controller=meta-check data-meta-url=#{viewer.purl_object.purl_url}.meta_json"
```
jcoyne, line 17:
```suggestion
      "data-controller=\"meta-check data-meta-url=#{viewer.purl_object.purl_url}.meta_json\""
```

jcoyne, line 17:
```suggestion
      "data-controller=\"meta-check\" data-meta-url=\"#{viewer.purl_object.purl_url}.meta_json\""
```

jcoyne: How about making a `meta_json_url` method too?

dnoneill: Done

app/components/companion_windows/external_link_component.rb
```diff
@@ -10,5 +10,11 @@ def initialize(viewer:, label:)
     attr_reader :viewer, :label
 
     delegate :external_url, to: :viewer
+
+    def stimulus_controllers
+      return '' unless viewer.instance_of?(::Embed::Viewer::Geo)
+
+      "data-controller=meta-check data-meta-check-url-value=#{viewer.purl_object.meta_json_url}"
```
jcoyne, line 17:
```suggestion
      "data-controller=\"meta-check\" data-meta-check-url-value=\"#{viewer.purl_object.meta_json_url}\""
```

dnoneill: This breaks things. When you do this it adds double quotes.
Screenshot 2025-03-06 at 2 31 20 PM

When you do what I have it renders properly:
Screenshot 2025-03-06 at 2 32 22 PM

jcoyne: Try adding `.html_safe` to the string.

dnoneill: Why can't we leave it as is? It is rendering correctly and html_safe isn't something we really like doing unless we have to which we don't

dnoneill: Yeah, I went with link_to. I think it is cleaner and clearer.

app/components/companion_windows/external_link_component.rb
```diff
@@ -10,5 +10,11 @@ def initialize(viewer:, label:)
     attr_reader :viewer, :label
 
     delegate :external_url, to: :viewer
+
+    def stimulus_attributes
+      return unless viewer.instance_of?(::Embed::Viewer::Geo)
+
+      { controller: 'meta-check', 'meta-check-url-value': viewer.purl_object.meta_json_url }
```
jcoyne, line 17:
```suggestion
      { controller: 'meta-check', meta_check_url_value: viewer.purl_object.meta_json_url }
```

jcoyne (approved):

PR #2750, opened 2025-03-05
Separate dom manipulation from the IIIF Auth (in different controllers)

jcoyne:


app/javascript/controllers/iiif_auth_restriction_controller.js
```diff
@@ -0,0 +1,45 @@
...
+// Display auth controls for the file_auth_controller.js
+export default class extends Controller {
+  static targets = ["locationRestriction", "locationRestrictionMessage", "messagePanel", "loginPanel",
+                    "loginButton", "loginMessage"]
+
+  // Bound to auth-denied CustomEvents
+  displayMessage(event) {
+    const detail = event.detail
+    this.locationRestrictionTarget.hidden = false
+    this.locationRestrictionMessageTarget.innerHTML = this.#retrieveAuthResponseMessage(detail.authResponse)
```
hudajkhan, line 12: Not sure if this is commented out because of the #

jcoyne: that's not a comment. That's the sigil for a private method in javascript https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

hudajkhan: Cool!

dnoneill:
@jcoyne I am double checking this but I think you might need to do the same updates in the model component.

dnoneill (dismissed):
Old behavior had a login header that showed you were logged in. This gets rid of that. Please add it back in. 
Screenshot 2025-03-06 at 12 22 44 PM

Also the location restriction is not working for 3d, which it should because that PR is merged.

dnoneill (changes requested):
1. Old behavior had a login header that showed you were logged in. This gets rid of that. Please add it back in. 
Screenshot 2025-03-06 at 12 22 44 PM

2. Also the location restriction is not working for 3d, which it should because that PR is merged.

hudajkhan:
We also need some way to clear out the login panel when we click on a location restricted item, otherwise we get something like this:
Screenshot 2025-03-06 at 11 08 31 AM

app/components/pdf_component.html.erb
```diff
@@ -8,13 +8,15 @@
   <% end %>
 
   <% component.with_authorization_messages do %>
-    <%= render LoginComponent.new %>
-    <%= render LocationRestrictionComponent.new %>
-    <%= render CompanionWindows::ContentNotAvailableComponent.new unless viewer.available? %>
+    <div data-controller="iiif-auth-restriction" data-action="auth-denied@window->iiif-auth-restriction#displayMessage needs-login@window->iiif-auth-restriction#displayLoginPrompt auth-success@window->iiif-auth-restriction#hideLoginPrompt thumbnail-clicked@window->iiif-auth-restriction#clearLocationRestriction">
```
hudajkhan, line 11: Perhaps we could have thumbnail-clicked clear all the message panels (Except the one that says "logged in" for a logged in user).  So not just clear location restriction but clearAuthMessages where we hide both login prompt and the location restriction panel

hudajkhan: Please see the comment above for context.

jcoyne:
@hudajkhan I believe these are solved.

hudajkhan:
Yes, the persisting login banner issue has been resolved, thank you

dnoneill:
@hudajkhan Looks good to me

dnoneill:
I think we should get https://github.com/sul-dlss/sul-embed/pull/2755 in first. Otherwise this is going to put a login header on something and not work.

dnoneill (approved):

PR #2762, opened 2025-03-07
Prevent two banners from displaying when citation only and embargoed

### LINKED ISSUE #2760: Double banners for file viewer
File has double banners for embargo and access restricted. This should only be the embargo banner

https://sul-purl-uat.stanford.edu/gx074xz5520?new_viewer=true

Image

jcoyne:
Fixes #2760

### DIFF
--- a/app/components/file_component.rb
+++ b/app/components/file_component.rb
@@ -7,7 +7,16 @@ class FileComponent < ViewComponent::Base
 
   attr_reader :viewer
 
+  delegate :purl_object, to: :viewer
+  delegate :citation_only?, :embargoed?, to: :purl_object
+
   def message
     viewer.authorization.message
+  end
+
+  # If this method is false, then display the "content not available" banner
+  # We exclude the embargoed state, because we prefer to show the embargo banner and we don't want two banners
+  def display_not_available_banner?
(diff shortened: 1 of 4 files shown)


app/components/file_component.rb
```diff
@@ -7,7 +7,16 @@ def initialize(viewer:)
...
+  delegate :citation_only?, :embargoed?, to: :purl_object
+
   def message
     viewer.authorization.message
   end
+
+  # If this method is false, then display the "content not available" banner
+  # We exclude the embargoed state, because we prefer to show the embargo banner and we don't want two banners
+  def display_not_available_banner?
+    citation_only? && !embargoed?
```
dnoneill, line 20: I think this needs to be in the auth messages component. If there is something that is view stanford, special collections, etc and download none this banner is going to show up.

jcoyne: It makes sense to wait until  https://github.com/sul-dlss/sul-embed/pull/2756 is merged then.

dnoneill: @jcoyne #2756 is merged. Please rebase.

hudajkhan (approved):

PR #2768, opened 2025-03-07
don't display sidebar unless there is a content list

### LINKED ISSUE #2766: Don't auto-expand sidebar unless there is a content list

dnoneill:
closes #2766

### DIFF
--- a/app/components/companion_windows_component.rb
+++ b/app/components/companion_windows_component.rb
@@ -57,7 +57,8 @@ class CompanionWindowsComponent < ViewComponent::Base
   def outer_container_data
     {
       controller: controllers,
-      iiif_manifest_loader_iiif_manifest_value: iiif_v3_manifest_url
+      iiif_manifest_loader_iiif_manifest_value: iiif_v3_manifest_url,
+      companion_window_auto_expand_value: render_content_list_panel?
     }.tap do |data|
       if fullscreen?
         data[:fullscreen_target] = 'area'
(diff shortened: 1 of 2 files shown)


app/components/companion_windows_component.rb
```diff
@@ -63,6 +63,7 @@ def outer_container_data
         data[:fullscreen_target] = 'area'
         data[:fullscreen_close_value] = 'Exit full screen'
       end
+      data[:companion_window_expanded_value] = render_content_list_panel?
```
corylown, line 66: This data attribute name makes it sound like it expresses the current state of the companion window. Maybe something like `companion_window_auto_expand` would be clearer?

dnoneill: That makes sense. Updated.

app/components/companion_windows_component.rb
```diff
@@ -63,6 +63,7 @@ def outer_container_data
         data[:fullscreen_target] = 'area'
         data[:fullscreen_close_value] = 'Exit full screen'
       end
+      data[:companion_window_auto_expand_value] = render_content_list_panel?
     end
```
corylown, line 67: Sorry...one more thing. Couldn't this new attribute go in the initial hash definition under `iiif_manifest_loader_iiif_manifest_value: iiif_v3_manifest_url` as `companion_window_auto_expand_value: render_content_list_panel?`?

corylown (changes requested):

app/components/companion_windows_component.rb
dnoneill: @corylown done.

corylown (approved):

PR #2771, opened 2025-03-07
fix banner height for login

dnoneill:
The banner for login was 46px


corylown:
https://github.com/sul-dlss/sul-embed/pull/2780 also fixes this by adjusting the padding of the dismiss button.

corylown:
This was fixed by https://github.com/sul-dlss/sul-embed/pull/2780

PR #2776, opened 2025-03-10
add login to geo viewer

dnoneill:


hudajkhan:
This may be a local setup problem, but when I try out https://purl.stanford.edu/bb033gt0615 locally, with a local version of stacks, for some reason the geo_viewer code identifies the wmsUrl as the restricted geoserver URL instead.  The probe service controller is called and returns the location as the restricted path.  This may involve a change on the stacks level instead?

hudajkhan (approved):

PR #2779, opened 2025-03-10
fix 3d test

dnoneill:
Pretty sure this is happening because we don't have the sidebar open by default anymore.

### DIFF
--- a/spec/features/3d_viewer_spec.rb
+++ b/spec/features/3d_viewer_spec.rb
@@ -15,6 +15,8 @@ RSpec.describe '3D Viewer', :js do
   end
 
   it 'has working panels' do
+    expect(page).to have_no_content('About this item')
+    page.find('[aria-controls="left-drawer"]').click
     expect(page).to have_content('About this item')
   end
 end


spec/features/3d_viewer_spec.rb
```diff
@@ -15,6 +15,6 @@
   end
 
   it 'has working panels' do
-    expect(page).to have_content('About this item')
+    expect(page).to have_selector('h3', text: 'About this item', visible: :all)
```
corylown, line 18:
```suggestion
    expect(page).to have_css('h3', text: 'About this item', visible: :all)
```

corylown (changes requested):

spec/features/3d_viewer_spec.rb
jcoyne, line 18: Should we click the button to open the panel?

dnoneill: I think this is a good idea. It allows for us to make sure it isn't expanding when it shouldn't be

corylown (approved):

PR #2780, opened 2025-03-10
Adjust auth bar font size and padding to match designs

corylown:
[Figma designs](https://www.figma.com/design/LhMpFAHNwF6tPo53p2xzyz/Embed-Viewer-Components?node-id=148-896&t=RKK3CauCQbdVZHLm-0)

Increases font size to 1.125 rem (18px)
With padding increases overall height from 40px to 48px.
Adjust dismiss button so it doesn't change the height of the auth bar.

Screenshot 2025-03-10 at 4 51 08 PM
Screenshot 2025-03-10 at 4 50 45 PM
Screenshot 2025-03-10 at 4 50 19 PM

I'll fix the `companion-window-component-body` height selector in a separate PR.

### DIFF
--- a/app/assets/stylesheets/companion_window.css
+++ b/app/assets/stylesheets/companion_window.css
@@ -22,7 +22,8 @@ body {
   display: flex;
   background-color: var(--active-color);
   color: white;
-  padding: 0.5rem;
+  padding: 0.75rem;
+  font-size: 1.125rem;
 
   svg:not(button svg) {
     /* Buttons have a left/right padding of 6px */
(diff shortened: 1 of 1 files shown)


hudajkhan:
@corylown looks good in general. I did see that when the "Stanford users: log in" version has height 49.6 px but I think we discussed this earlier with @alundgard and that was fine. All the other variations appear to be exactly 48 px high.

corylown:
@hudajkhan thanks. I should be able to fix that so they're all the same height.

corylown:
@hudajkhan I think I fixed it

hudajkhan:
Fixed thank you!

hudajkhan (approved):

PR #2789, opened 2025-03-12
make image tools show up on all viewers

### LINKED ISSUE #2785: Add Parker "Image Tools" bar to Mirador?
[Example from Parker](https://embed.stanford.edu/iiif?canvas_id=&iiif_initial_viewer_config=&image_tools=true&url=https%3A%2F%2Fdms-data.stanford.edu%2Fdata%2Fmanifests%2FParker%2Fnm203xw8381%2Fmanifest.json).
Image

dnoneill:
closes #2785

### DIFF
--- a/app/javascript/src/modules/m3_viewer.js
+++ b/app/javascript/src/modules/m3_viewer.js
@@ -121,7 +121,7 @@ export default {
       }
     }, [
       ...((cdl && cdlAuthPlugin) || []),
-      ...((imageTools && miradorImageToolsPlugin) || []),
+      ...miradorImageToolsPlugin,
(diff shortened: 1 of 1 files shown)


jcoyne (changes requested):
Can you remove line 23?  How about the stuff in the HTML that adds that data property?

dnoneill:
@jcoyne No because that value is being used to determine whether we use embedModePlugin, the view (mosiac vs single). We can change the name but that is a lot of renaming. We would also have to update exhibits because it is getting the name from a url param

jcoyne (approved):

PR #2792, opened 2025-03-12
make sandbox urls persistent

dnoneill:
this should make it so you refresh the sandbox page and it will refresh the previous purl you were looking at and embed it.

### DIFF
--- a/app/views/pages/sandbox.html.erb
+++ b/app/views/pages/sandbox.html.erb
@@ -167,6 +167,12 @@
       }
     }
 
+    function updateSandboxUrl(purl_url) {
+      const currentUrl = new URL(window.location.href);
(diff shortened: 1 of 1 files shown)


hudajkhan:
@dnoneill I tried this out but it isn't quite working for me. I am using sul-purl-stage as my purl URL in local settings configuration.  I can see the code saves the state/history with the URL I entered, and the embed viewer works fine. When I hit refresh, the url is whatever the original value was and not the one I entered

hudajkhan:
What does work is that the purl object corresponding to the object in the URL when the page loads is rendered in the embed viewer

hudajkhan:
Note: This works as intended with a regular refresh, not a hard refresh.

PR #2795, opened 2025-03-13
Zebras

### LINKED ISSUE #2767: File viewer: Add table zebra stripes
Background color: _Fog Light_ `#F4F4F4` Opacity 70%.

Image

Image

Image

hudajkhan:
Closes #2767 

What this pull request does:
- Applies zebra stripes to the table which displays the directory names and files name in the file contents
- Adds a div after the table that fills the rest of the page. 
- Uses CSS to fill that div with zebra stripes. 
- Calculates how many rows are taken up by files and directories (directories get their own row if they have a title).  Based on that, sets the class for zebra stripes.  Since the rows start with the fog color, if there are an even number of rows, the background style should begin with a fog color. If odd, the style should begin with a white color.
- Updates the toggle function in the tree stimulus controller, to also update the background style based on the number of rows now visible once toggling has occurred. 
- Adds tests for the file component around the styling of the rows displaying files/directory titles

Screenshots after applying these changes:

With a single item:
Screenshot 2025-03-13 at 3 59 10 PM

With an even number of items:
Screenshot 2025-03-13 at 4 00 14 PM

Toggling the items in the above page to change from even to odd:
Screenshot 2025-03-13 at 4 00 46 PM

### DIFF
--- a/app/assets/stylesheets/file.css
+++ b/app/assets/stylesheets/file.css
@@ -125,9 +125,13 @@
     transform: rotate(90deg);
   }
 
+  tbody > tr {
+    border: 0
+  }
+
   /* Zebra stripes for file row */
   tbody > tr:nth-child(even of :not([hidden])) {
-    background-color: var(--stanford-fog-light)
+    background-color: rgba(244, 244, 244, 0.7);
   }
   
   .folder-title {
@@ -173,12 +177,12 @@
 
 /* Display this background if the number of rows in the file contents table is odd */
 .stripes-background.odd {
-  background: repeating-linear-gradient(#f4f4f4,#f4f4f4 37px,#d5d5d4 37px,#d5d5d4 38px,#fff 38px,#fff 75px, #d5d5d4 75px,#d5d5d4 76px)
+  background: repeating-linear-gradient(rgba(244, 244, 244, 0.7),rgba(244, 244, 244, 0.7) 38px,#fff 38px,#fff 76px)
 }
 
 /* Display this background if the number of rows in the file contents table is even */
 .stripes-background.even {
-  background: repeating-linear-gradient(#fff,#fff 37px,#d5d5d4 37px,#d5d5d4 38px,#f4f4f4 38px,#f4f4f4 75px, #d5d5d4 75px,#d5d5d4 76px)
+  background: repeating-linear-gradient(#fff,#fff 38px,rgba(244, 244, 244, 0.7) 38px,rgba(244, 244, 244, 0.7) 76px)
 }
 
 .sul-embed-downloads svg {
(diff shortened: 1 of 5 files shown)


alundgard:
Thanks for the screenshots - looking good! Some small design details:
- The first row item has a white background. The second row has the gray background.
- The rows no longer have a border stroke. It looks like gray stroke is still present above.
- The gray background above looks a little dark? If it isn't already, would it be possible to do  _Fog Light_ `#F4F4F4` Opacity 70%. That is, `rgba(244, 244, 244,0.7)`.

Image

hudajkhan:
The first and third I will work on. I am not sure what you mean by the second so maybe we can discuss it so I understand. Thanks!

hudajkhan:
Updated screenshots:
one item:
Screenshot 2025-03-13 at 5 32 59 PM

multiple items:
Screenshot 2025-03-13 at 5 33 35 PM

collapsing item to change number of visible rows for files:
Screenshot 2025-03-13 at 5 34 09 PM

hudajkhan:
Updated with above suggestions:
Multiple items
Screenshot 2025-03-13 at 5 52 51 PM

Multiple items with toggled row
Screenshot 2025-03-13 at 5 53 29 PM

app/assets/stylesheets/file.css
```diff
@@ -161,6 +170,21 @@
   }
 }
 
+/* The background applied relies on the number of rows displayed in the file viewer */
+.stripes-background {
+  height:100%;
```
dnoneill, line 175: @hudajkhan I think this needs to be calculated in the js. Right now it is making the viewer scroll when there is only two items

app/components/file_component.rb
```diff
@@ -19,4 +19,32 @@ def message
   def display_not_available_banner?
     citation_only? && message.blank?
   end
+
+  # Display of zebra stripes for remaining space depends on number of rows displayed
+  def display_row_count(hierarchical_contents)
```
dnoneill, line 24: I am pretty sure this can all be done in the js with less work

dnoneill (changes requested):
Right now this is adding random rows even if the viewer doesn't need the extra rows. 

Screenshot 2025-03-14 at 12 27 05 PM

hudajkhan:
@dnoneill I believe they want the extra rows in your screenshot. We can confirm with @alundgard

hudajkhan:
@noneill, the height and stripes class are now all set in JavaScript and appear to be working without leading to more scrolling.

dnoneill (approved):
@hudajkhan this looks great. Great work. I am just noting this for later. I want to check with @alundgard but right now the folder heights are 34px and file heights are 38px which is causing a weird experience when you collapse folders. It think that might be a different ticket. I am also not sure if it is on purpose.

hudajkhan:
@dnoneill I agree regarding checking with @alundgard and then we can update as needed

PR #2805, opened 2025-03-17
show unavaliable pdfs

### LINKED ISSUE #2783: Show "download: none" pdfs in the content list in the document viewer
After reviewing the per-file restriction banners in the document viewer with @alundgard, we've decided we should show PDF file in the contents list in the companion window after all. This will resolve the unusual case where different files have different rights. 

Currently, if a file has "download: none", it will be excluded from the contents list. When all files are "download: none", this is not really a big issue because there won't be a contents list at all and the descriptive metadata should describe the whole item for the user to decide whether to request access directly via email or in person.

When an item has files with a mix of rights settings, the effect of hiding the "download: none" files is jarring because the files simply disappear from the list. This gives a misleading view of what files are actually in the item. For example:

There is an item on staging that has 4 files with "download: none" and 9 files in total, [as shown in Argo](https://argo-stage.stanford.edu/view/druid:gm706hs4976). But [in the viewer](https://argo-stage.stanford.edu/view/druid:gm706hs4976), only the 5 files that allow some level of download access appear in the content list.

Image

Now that we support showing a banner on a per-file basis, we should add the "download: none" files to the contents list with the appropriate banner message: "This item cannot be accessed online ..." This will at least give users a more complete picture of what files are contained in an item.

This does mean that multi-file items where all files are "download: none" will have a contents list where all of the links in the list lead to the access restricted display, which could be frustrating to someone who tries every link. But since the banner should be identical for each file, hopefully it will be clear that the restriction applies to all of them.

dnoneill:
closes #2783


jcoyne (approved):

app/viewers/embed/viewer/document_viewer.rb
```diff
@@ -29,13 +29,6 @@ def fullscreen?
...
-      # This indicates if the first PDF is downloadable (though it could be stanford only)
-      # as well as location restricted.  Authorization for other documents will be
-      # checked as the user clicks on items in the content side bar.
-      def available?
-        !document_resource_files.first&.no_download?
-      end
-
       private
 
       def document_resource_files
```
jcoyne, line 34: I think this method can go too.

dnoneill: I think so too, which means we remove pdf_files. I can't quite figure out how the sidebar is getting rendered

dnoneill: Ah. I see it is looking at the purl's file.

dnoneill:
@jcoyne pdf_files, document_resource_files, and spec/lib/embed/viewer/document_viewer_spec.rb. I am assuming this is okay to merge or do you want to do another review?

PR #2824, opened 2025-03-19
consolidate modal functions, fix closeModal to only focus when using the keyboard

### LINKED ISSUE #2788: Share & download tooltip appears after closing modal
The tooltip should only appear when the button is hovered or receives keyboard focus.
<video src="https://github.com/user-attachments/assets/5c74b72b-c0b7-44d8-95a3-b05724ce809d">

dnoneill:
closes #2788


app/javascript/controllers/companion_window_controller.js
```diff
@@ -71,16 +71,17 @@ export default class extends Controller {
     if (!clickWithinDialog) modal.close()
   }
 
-  openShareModal() {
-    this.shareModalTarget.showModal()
+  openModal(event) {
+    const clickedButton = event.currentTarget.closest('button');
+    this.setReturnFocusButton(clickedButton);
+    const modalName = clickedButton.dataset['target'];
```
jcoyne, line 77:
```suggestion
    const modalName = clickedButton.dataset.target
```

app/javascript/controllers/companion_window_controller.js
```diff
@@ -71,16 +71,17 @@ export default class extends Controller {
...
-  openDownloadModal() {
-    this.downloadModalTarget.showModal()
-  }
-
-  displayAccessibility() {
-    this.accessibilityTarget.showModal()
+  setReturnFocusButton(button) {
+    // we want the focus to return to share/download button not to the closed dropdown menu
+    if (button.parentElement.tagName == 'DIALOG') return;
+    this.returnFocusButton = button.matches(':focus-visible') ? button : false;
```
jcoyne, line 84: I don't like mixing types here.  Can it  be

```suggestion
    if (button.parentElement.tagName !== 'DIALOG' && button.matches(':focus-visible'))
        this.returnFocusButton = button
```

app/javascript/controllers/companion_window_controller.js
```diff
@@ -71,16 +71,17 @@ export default class extends Controller {
     if (!clickWithinDialog) modal.close()
   }
 
-  openShareModal() {
-    this.shareModalTarget.showModal()
+  openModal(event) {
+    const clickedButton = event.currentTarget.closest('button');
+    this.setReturnFocusButton(clickedButton);
```
jcoyne, line 76:
```suggestion
    const clickedButton = event.currentTarget.closest('button')
    this.setReturnFocusButton(clickedButton)
```
Remove semicolons.

jcoyne (changes requested):

dnoneill:
Yeah when using the mouse, focus was being put back on the original pressed button. We only want that happening if the user is using the keyboard. I am attaching a video as well.


https://github.com/user-attachments/assets/405f4471-6846-4402-9380-a39072969524

jcoyne (approved):

PR #2825, opened 2025-03-19
Deprecate image_tools parameter in favor of enable_comparison

### LINKED ISSUE #2802: Create new request option for comparison view
We've enabled image_tools for all viewers, but there are still some differences, so a better name would help clarify.

Presently, we want `comparison=true` to work identically to `image_tools=true`, but deprecate `image_tools`.  We can then change exhibits to use `comparison=true` instead.   

See https://github.com/sul-dlss/sul-embed/pull/2789#issuecomment-2722409413

See https://github.com/sul-dlss/exhibits/blob/a63e8c1fded1687a4b242a1790cdf1dbb0007302/app/views/viewers/_mirador3.html.erb#L5

corylown:
Fixes #2802

### DIFF
--- a/app/javascript/src/modules/m3_viewer.js
+++ b/app/javascript/src/modules/m3_viewer.js
@@ -20,7 +20,7 @@ export default {
     const data = el.dataset
     const showAttribution = (data.showAttribution === 'true')
     const hideWindowTitle = (data.hideTitle === 'true')
-    const imageTools = (data.imageTools === 'true')
+    const enableComparison = (data.enableComparison === 'true')
     const cdl = (data.cdl === 'true')
 
     // Determine which panel should be open
@@ -114,7 +114,7 @@ export default {
       },
       workspace: {
         showZoomControls: true,
-        type: imageTools ? 'mosaic' : 'single',
+        type: enableComparison ? 'mosaic' : 'single',
       },
       workspaceControlPanel: {
         enabled: false,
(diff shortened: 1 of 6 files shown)


dnoneill:
@corylown can we get a pr to fix this in exhibits so there isn't that much breakage time

corylown:
@dnoneill I will do that but this change will not (I think) break anything in exhibits.

dnoneill:
Ah, I see. You are checking for the image-tools params. 👍

app/javascript/src/modules/m3_viewer.js
```diff
@@ -20,7 +20,8 @@ export default {
     const data = el.dataset
     const showAttribution = (data.showAttribution === 'true')
     const hideWindowTitle = (data.hideTitle === 'true')
-    const imageTools = (data.imageTools === 'true')
+    // TODO: Remove deprecated data.imageTools once exhibits is updated to use enableComparison
+    const enableComparison = (data.enableComparison === 'true' || data.imageTools === 'true')
```
dnoneill, line 24: I don't think you need data.imageTools here. I am pretty sure there data fields come from the data fields and you changed `data-image-tools` to `data-enable-comparison='<%= @embed_request.enable_comparison? %>'`

corylown: I think you're right. For whatever reason I thought exhibits was supplying its own data attributes, but it's just supplying the parameter.

Fixed.

dnoneill (approved):

PR #2830, opened 2025-03-19
Allowing items with view set to location restriction to show restriction message in the media contents list

### LINKED ISSUE #2815: When rights are "view: location-based" and "download: none" show the "restricted" message in the contents list
The media viewer is not showing "restricted" in the contents list when rights are "view: location-based" and "download: none": https://sul-purl-stage.stanford.edu/jy637fq4076

> Image

The message does show when the files are downloadable at the location: https://sul-purl-stage.stanford.edu/qb529cx7661

> Image

I'm guessing since the "restricted" label probably originates in the download modal, the "download: none" files do not get labeled because they don't appear in the modal at all. Since the banner at the top of the viewer indicates the restriction, it's not a huge problem to lack the label but it is inconsistent with the rest of the display.

hudajkhan:
Closes #2815.

What this pull request does:

Adds a method to return if the view is location based (b/c the file method for location_restricted? looks only at download values). 

Ensures that the wrapper component adds the data attributed for location restriction set to true in the case where view is location restricted (regardless of the value of the download property).  

Updates the wrapper component tests to include a case when file returns location restricted = false (based on download) and view location restricted true.

### DIFF
--- a/app/components/media/wrapper_component.rb
+++ b/app/components/media/wrapper_component.rb
@@ -21,7 +21,7 @@ module Media
                 media_wrapper_index_value: @resource_index,
                 action: 'thumbnail-clicked@window->media-wrapper#toggleVisibility',
                 stanford_only: @file.stanford_only?,
-                location_restricted: @file.location_restricted?,
+                location_restricted: @file.location_restricted? || @file.view_location_restricted?,
                 file_label: @file.label_or_filename,
                 media_tag_target: 'mediaWrapper',
                 thumbnail_url: @thumbnail.presence,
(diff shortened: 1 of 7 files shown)


app/components/media/wrapper_component.rb
```diff
@@ -21,7 +21,7 @@ def call # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
                 media_wrapper_index_value: @resource_index,
                 action: 'thumbnail-clicked@window->media-wrapper#toggleVisibility',
                 stanford_only: @file.stanford_only?,
-                location_restricted: @file.location_restricted?,
+                location_restricted: @file.location_restricted? || @file.view_location_restricted?,
```
jcoyne, line 24: I don't think we want `location_restricted?' we only care about where they can view it here.

hudajkhan: Thanks @jcoyne. I have made that change.

jcoyne (approved):

PR #2841, opened 2025-03-20
don't check probe service if public geo

dnoneill:
We are query probe service for all geo not just restricted geo items. The map rendering isn't tied to probe service so we are just rendering and calling the probe service that doesn't do anything.


hudajkhan (approved):

hudajkhan:
Works as intended. Thanks!

PR #2847, opened 2025-03-21
fix contents list showing with one item

dnoneill:
The sidebar opens even there is only one file for https://sul-purl-uat.stanford.edu/bd989qy8176. That is because it is seeing the supplemental txt file which we don't display. This should fix that issue:
Screenshot 2025-03-21 at 12 09 03 PM


jcoyne (changes requested):
How about adding a test for this case?

dnoneill:
@jcoyne Done

jcoyne (approved):

PR #2849, opened 2025-03-21
default locked image for video without posters and not world viewable, hide lock poster when auth-success

### LINKED ISSUE #2747: "Lock" poster image no longer appearing on restricted media
During the fall 2023 workcycle, we implemented an approach to restricted media where we show a poster image of a "lock" when an item has no poster/thumbnail image and the user is not authorized to view it. (Issue: https://github.com/sul-dlss/sul-embed/issues/1706, PR: https://github.com/sul-dlss/sul-embed/pull/1721

I've checked all of our media samples on stage and I am no longer seeing the lock image anywhere. I think it should be appearing on these items, at least:

- [audio with no poster (citation-only)](https://sul-purl-stage.stanford.edu/ry580bp8547)
- [video with no poster (citation-only)](https://sul-purl-stage.stanford.edu/db343wf9494)
- [audio with no poster (location)](https://sul-purl-stage.stanford.edu/rt655ws0601)
- [video with no poster (location)](https://sul-purl-stage.stanford.edu/qb529cx7661)

dnoneill:
closes #2747 

This is happening because audio have a poster and so if a video and audio item are in the same viewer it thinks they have a poster. Since we want to keep the bar bellow the video, this sets the default to locked if there isn't a poster. Then when there is an auth-success it removed that locked poster icon.


app/components/media_component.html.erb
```diff
@@ -83,13 +83,6 @@
         <%= render LockedStatusComponent.new %>
       </div>
 
-      <div data-controller="locked-audio-poster" data-action="auth-success@window->locked-audio-poster#hide needs-login@window->locked-audio-poster#show auth-denied@window->locked-audio-poster#show" role="presentation" style="flex: 1 0 100%;" hidden>
```
dnoneill, line 86: We no longer use the audio tag for the viewer.

jcoyne: Per https://github.com/sul-dlss/sul-embed/issues/2747#issuecomment-2701535741 I thought we were leaving the audio behavior as is?

dnoneill: Yeah but we aren't actually using the <audio> tag anymore so audio is treated the same way, I am removing code that doesn't run anymore because the <audio> tag isn't used anymore. https://github.com/sul-dlss/sul-embed/pull/2386. So this code isn't doing anything.

PR #2852, opened 2025-03-24
show logged in banner if the user is already logged in

### LINKED ISSUE #2845: Show the "logged in" banner to Stanford users who are already logged in when arriving at a stanford-only item
The media viewer was updated last year to show the "logged in" banner to users who:

- just logged into a stanford-restricted item
- were already logged in before arriving at a stanford-restricted item

Currently, the document, geo, and 3d viewers only show the "logged in" banner when someone just logged into the viewer. Users who arrive already logged in will get access to the resource but will not see that it is stanford-restricted.

We should show the "logged in" banner to all users who have already logged in. If that isn't possible, it would be better to show the log in prompt again, as the goal of persisting the banner is to make it clear to users when they are viewing stanford-restricted items.

dnoneill:
closes #2845

### DIFF
--- a/app/javascript/controllers/file_auth_controller.js
+++ b/app/javascript/controllers/file_auth_controller.js
@@ -16,7 +16,7 @@ export default class extends Controller {
         this.displayAccessTokenError(event.data)
       } else {
         this.cacheToken(event.data.accessToken, event.data.expiresIn)
-        window.dispatchEvent(new CustomEvent('login-success', { detail: {} }))
+        window.dispatchEvent(new CustomEvent('show-message-panel', { detail: {} }))
         this.queryProbeService(event.data.messageId, event.data.accessToken)
           .then((result) => this.renderViewer(result))
           .catch((json) => console.error("no access", json))
@@ -145,7 +145,10 @@ export default class extends Controller {
         const token = this.getCachedToken()
         if (token) {
           this.queryProbeService(messageId, token)
-            .then((result) => this.renderViewer(result))
+            .then((result) => {
+              this.showLoggedInBanner(authResponse)
+              this.renderViewer(result)
+            })
             .catch((json) => {
               console.debug("Probe with cached token failed", json)
               this.queryAccessService(accessService, messageId)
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/file_auth_controller.js
```diff
@@ -157,6 +160,12 @@ export default class extends Controller {
       })
   }
 
+  requiresLogin(authResponse) {
+    if (authResponse.status == 401 && authResponse.heading.en[0].includes('log in')){
+      window.dispatchEvent(new CustomEvent('login-success', { detail: {} }))
```
jcoyne, line 165: Should be a new event type? This is not occurring after they have successfully logged in. This is after authorization.  We're conflating meanings vis-à-vis authN and authZ.

jcoyne (changes requested):

app/javascript/controllers/file_auth_controller.js
dnoneill: I was unsure about this, because this is being called after the application grabs the local token and checks the access which means the user kind of logged in but the application did it for them?

jcoyne: Can we get rid of the existing `login-success` (in `addPostCallbackListener`) and just have this new event?

dnoneill: so instead of login-success something like show logged in banner or something along those lines?

dnoneill: I went with show-message-panel

app/javascript/controllers/file_auth_controller.js
```diff
@@ -16,7 +16,7 @@ export default class extends Controller {
         this.displayAccessTokenError(event.data)
       } else {
         this.cacheToken(event.data.accessToken, event.data.expiresIn)
-        window.dispatchEvent(new CustomEvent('login-success', { detail: {} }))
+        window.dispatchEvent(new CustomEvent('show-message-panel', { detail: {} }))
```
jcoyne, line 19: is this line still needed? will the new one take care of this?

jcoyne (changes requested):

app/javascript/controllers/file_auth_controller.js
dnoneill: Still needed. The new line will not take care of it. This line happens after a user has hit the login button. The other location happens as soon as the user loads the page. They both need to exist.

PR #2853, opened 2025-03-25
preventing chrome error when highlight cue called

hudajkhan:


app/javascript/controllers/transcript_controller.js
```diff
@@ -176,6 +176,12 @@ export default class extends Controller {
   }
 
   highlightCue(evt) {
+    // This method is called when the time-updated event is triggered
+    // In Chrome, the event may be called even before the player is loaded
```
jcoyne, line 180: can you be more specific about what you mean by "the player"?

jcoyne: Is it possible this happens if the video starts before we have the vtt?

PR #2856, opened 2025-03-25
add error listener for image

### LINKED ISSUE #2807: Use icons in place of images when users are not authorized to view image thumbnails in contents list
**Current behavior**

When a user is viewing a restricted media item and the user is not authenticated, the contents sidebar shows either a blank space (Firefox/Chrome) or a question mark placeholder (Safari) where an image would appear if the user is authenticated to see it.

Example 1:

This item has multiple image resources. Only the thumbnail for the first image appears, no subsequent image appears.

Image

Example 2:

This item has an image thumbnail attached to the second resource (the audio), but only the first image shows.

Image

**Desired behavior**

When an image can't be displayed because a user is not authenticated, we would like to show an icon representing the file type instead of a blank space.

There are examples where we already do this for media items that have no attached images. For example, this item has no image thumbnails for the audio and video files. It shows icons in place of images in the contents list:

Image

We would like to do the same thing for items where there are images, but where the user cannot see them.

**Additional information: What if the user is authenticated?**

If the user is authenticated, they should see the actual image instead of the icon.

dnoneill:
closes #2807 

Before login
Screenshot 2025-03-25 at 4 50 39 PM

After login 
Screenshot 2025-03-25 at 4 50 28 PM

### DIFF
--- a/app/javascript/src/modules/thumbnail.js
+++ b/app/javascript/src/modules/thumbnail.js
@@ -19,7 +19,7 @@ export default class {
 
     let thumbnailIcon = ''
     if (this.thumbnailUrl) {
-      thumbnailIcon = `<img class="square-icon" src="${this.thumbnailUrl}" alt="" />`
+      thumbnailIcon = `<img data-controller="image" data-action="error->image#handleError auth-success@window->image#updateImage" class="square-icon" src="${this.thumbnailUrl}" alt="" />`
     } else {
       thumbnailIcon = `<i class="${this.defaultIcon} default-thumbnail-icon"></i>`
     }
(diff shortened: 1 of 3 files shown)


app/javascript/src/modules/thumbnail.js
```diff
@@ -19,7 +19,7 @@ export default class {
 
     let thumbnailIcon = ''
     if (this.thumbnailUrl) {
-      thumbnailIcon = `<img class="square-icon" src="${this.thumbnailUrl}" alt="" />`
+      thumbnailIcon = `<img data-controller="image" data-action="error->image#handleError auth-success@window->image#updateImage" class="square-icon" src="${this.thumbnailUrl}" alt="" />`
```
jcoyne, line 22: What is the error we are trapping for here?  Is it possible this catches some other error that we don't want?

dnoneill: I don't think so. This should be something default to browser that gets triggered when an image fails to load. http://www.w3schools.com/Jsref/event_onerror.asp

jcoyne: Oh, right it's scoped to this element only.

jcoyne (approved):

PR #2898, opened 2025-07-29
fix tooltip breaking map look inside

dnoneill:
https://earthworks.stanford.edu/catalog/stanford-cf800gc3842 is breaking because the feature.properties.label is a number and leaflet does not like that.

https://earthworks.stanford.edu/catalog/stanford-nz504hc9042


dnoneill:
@jcoyne They are but I don't necessarily thing 1234 should break the entire application. That seems kind of weird that we don't allow a label to be a number.

jcoyne:
Yeah, the spec actually doesn't specify anything about the structure of "properties"

> The value of the properties member is an object (any JSON object or a JSON null value)

PR #2905, opened 2025-08-08
load purl, not resource version

dnoneill:


app/javascript/controllers/media_player_controller.js
```diff
@@ -40,6 +40,10 @@ export default class extends Controller {
     })
   }
 
+  fileName(filepath){
```
jcoyne, line 43: Will this work if the file is `path1/file.mov` and `path2/file.mov`?

jcoyne (approved):

PR #2941, opened 2025-10-02
Remove obsolete gem dor-rights-auth

jcoyne:


dnoneill:
@jcoyne Can you put what we used to use this for. Was it cdl?

jcoyne:
@dnoneill it's how we handled authorization in XML prior to using cocina.

PR #2976, opened 2026-01-29
fix stanford only icon css to be after text, not background image

### LINKED ISSUE #2975: display issue with longer title and stanford-only icon in the contents sidebar
See https://purl.stanford.edu/wc835wx0065 

Image

dnoneill:
closes #2975 

Before:
Screenshot 2026-01-29 at 4 53 38 PM
After:

Screenshot 2026-01-30 at 2 00 03 PM

Before:
Screenshot 2026-01-30 at 2 00 25 PM

After:
Screenshot 2026-01-30 at 2 00 15 PM

### DIFF
--- a/app/assets/stylesheets/content_list.css
+++ b/app/assets/stylesheets/content_list.css
@@ -44,6 +44,17 @@
       background: url("stanford_s.svg") no-repeat;
       background-position: bottom left;
       padding-right: 20px;
+
+      .label-text {
+        display: -webkit-box;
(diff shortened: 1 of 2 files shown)


dnoneill:
@jcoyne Yeah I realized that after that is why it is in draft. I aske Alan because I am not sure our current implementation is possible without fully cutting off the title at equalized. It that is solution that is fine but I think I need more context.

PR #3007, opened 2026-03-30
Display geoJSON layers in MapLibre

jcoyne:


app/javascript/controllers/geo_controller.js
```diff
@@ -61,16 +61,46 @@ export default class extends Controller {
...
+      fetch(this.dataAttributes.geoJson)
+        .then(response => response.json())
+        .then(data => this.renderGeoJSON(data))
     } else {
       this.renderReplacementRectangle()
     }
   }
 
+  renderGeoJSON(data) {
+    console.log("adding geojson data")
```
thatbudakguy: Should we take this out?

PR #3038, opened 2026-04-17
Inspect pmtiles

### LINKED ISSUE #3006: Support PMTile inspection in the Geo viewer

jcoyne:
Fixes #3006


app/javascript/controllers/geo_controller.js
```diff
@@ -131,16 +131,79 @@ export default class extends Controller {
...
+  pmtilesInfo(data) {
+    let output = '<div class="pmtiles-info"><div><dl>'
+
+    // Display all properties from the feature
+    Object.entries(data).forEach(([key, value]) => {
+      if (value !== null && value !== undefined && value !== "") {
+        // Format the key to be more readable (capitalize, replace underscores)
+        const formattedKey = key
+          .replace(/_/g, " ")
+          .replace(/\b\w/g, l => l.toUpperCase())
```
thatbudakguy: We might want to run this by our stakeholders. There's a ton of variation in this data, and I could see this producing dubious results sometimes – and sometimes you may want to know the actual attribute names from the table, weird-looking as they are.

That said, it would be cool to eventually use the information from the ISO19110 XML metadata (if the object has one) to transform these programmatically into human-readable names, because I think that file contains extra metadata about the attributes.

thatbudakguy (approved):

PR #3042, opened 2026-04-22
Don't show the IIIF url for index maps in the sidebar unless it's ava…

### LINKED ISSUE #3040: IIIF manifest link rendered for index map sheets that we don't hold
For index map sheets that don't actually correspond to an item we have (shown using red polygon), we're still trying to render a link to a (nonexistent) IIIF manifest.

Visible on kh860qj3406:

Image

jcoyne:
…ilable

Fixes #3040


app/javascript/geo/index_map_renderer.js
```diff
@@ -118,6 +118,7 @@ export class IndexMapRenderer {
   }
 
   indexMapInfo(data) {
+    console.log("data", data)
```
thatbudakguy: Leftover?

jcoyne: yes. removed.

thatbudakguy (approved):

PR #3064, opened 2026-05-05
Toggle map collar

### LINKED ISSUE #3063: Support toggling the georeference mask visibility for warped IIIF maps
Based on stakeholder feedback, it would be nice for warped IIIF maps that have a mask applied (e.g. to remove the "collars" from the map) to be able to toggle the mask on/off, so that patrons could view any useful information (map scale, legend, other metadata) that was contained in the masked area.

bb013fz9675 is an example; you can see from the [parent item record in searchworks](https://searchworks.stanford.edu/view/2941954) that these maps have extra information around the map area which gets cropped out when georeferencing.

We can try adding a custom MapLibre control that toggles the masking behavior on and off. One of the AllMaps developers responded that this should be possible:

> Yes! This is possible using the options, e.g. via your warpedMapLayer's setLayerOptions() (setting the options for all maps in the layer) or setMapOptions() (setting the options for one map, this takes precedence over layer options). You'll want to set the applyMask option, which is true by default, to false. The MapLibre plugin package docs points to the docs of the WarpedMapLayer package, which centralises this (and other) functionality across the three plugins (MapLibre, Leaflet, OpenLayers). You should find docs of these methods and the available options there. If you have any other questions, let us know!

jcoyne:
Fixes #3063

### DIFF
--- a/app/assets/stylesheets/geo.css
+++ b/app/assets/stylesheets/geo.css
@@ -1,6 +1,8 @@
-@import url("companion_window.css");
-@import url("common.css");
-@import url("https://unpkg.com/maplibre-gl@5.21.1/dist/maplibre-gl.css");
+@layer layout, common, framework;
+@import url("companion_window.css") layer(layout);
+@import url("common.css") layer(common);
+@import url("https://unpkg.com/maplibre-gl@5.21.1/dist/maplibre-gl.css")
+layer(framework);
 
 .sul-embed-geo {
   isolation: isolate; /* So that tooltips can hover above Leaflet.js */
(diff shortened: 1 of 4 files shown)


thatbudakguy:
The button is pretty difficult to read when hovered for me:

Screenshot 2026-05-05 at 09 55 01

thatbudakguy:
What do you think about using an icon for the button? Maybe [something like "Crop"](https://mui.com/material-ui/material-icons/?query=frame&selected=Crop)?

jcoyne:
Crop is a destructive operation, so people may avoid it out of fear. I'm not sure how it would toggle either.

thatbudakguy:
It sounds like we have semi-consensus re: usage of the "intersect" and "union" bootstrap icons as [mentioned in slack](https://stanfordlib.slack.com/archives/C0AAK1HNY/p1778000468958059). can we try using those as toggle states, and also showing the collar by default?

thatbudakguy (approved):

PR #3075, opened 2026-05-26
Use the georeference role to find IIIF georeference annotations

### LINKED ISSUE #1754: Incorporate georeference annotations into IIIF manifests using dedicated file role
Currently we search for IIIF image annotations using a combination of the `.json` extension and the `annotations` file role, and incorporate them into the manifest when found:
https://github.com/sul-dlss/purl/blob/c0d7d31be519c68c4f415428fb651074ea4a5d74/app/models/iiif3_presentation_manifest.rb#L281-L285

This creates a problem with georeference annotations, because there's no way to tell at present whether a given file of annotations is annotating the images of the object, georeferencing the object, or both.

The embed viewer needs to know if georeference annotations are present so that it can decide to render a map, and the indexers need to know if georeference annotations are present so they can represent that in the record metadata for searching.

This ticket proposes adding (really just looking for, since there is no validation on the file role, called `use` in Cocina, in SDR) a file role called `georeference`, which would be picked up by similar logic to the above and incorporated into the manifest `AnnotationList`.

This would allow distinguishing image annotations from georeference annotations, as well as the ability to update the two separately from each other, e.g. for a historical map that was both georeferenced and annotated.

thatbudakguy:
See https://github.com/sul-dlss/purl/issues/1754

### DIFF
--- a/app/models/embed/purl.rb
+++ b/app/models/embed/purl.rb
@@ -77,6 +77,10 @@ module Embed
       downloadable_files.any?(&:annotations?)
     end
 
+    def iiif_georeference_annotations?
+      downloadable_files.any?(&:georeference_annotations?)
+    end
+
     def purl_url
       return "#{Settings.purl_url}/#{@druid}" if @version_id.blank?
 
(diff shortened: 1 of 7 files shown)


thatbudakguy:
Can be tested on https://sul-purl-stage.stanford.edu/bm355ph4409.

jcoyne (approved):

PR #3083, opened 2026-06-17
Add PMTile layers for points and lines

### LINKED ISSUE #3081: PMTiles with point geometry aren't rendered in Geo viewer
Visible on [kx858my6814](https://sul-purl-stage.stanford.edu/kx858my6814).

This is likely because we hard-code the MapLibre layer type to `fill`:
https://github.com/sul-dlss/sul-embed/blob/e322b21be5ff3b19634e39ade3be2dab5914f31d/app/javascript/geo/pmtiles_renderer.js#L48

The fix is probably to add separate layers for fill, point, line, etc. and then add filters to each corresponding to the relevant geometry type. For an example of this approach, see the [PMTiles viewer source](https://github.com/protomaps/PMTiles/blob/8b8ddea4dbff1b0104cf2bebf2f7ff35c91b41d5/app/src/PageMap.tsx#L138-L188), noting the `filter` expressions for each layer.

jcoyne:
Fixes #3081

### DIFF
--- a/app/javascript/geo/pmtiles_renderer.js
+++ b/app/javascript/geo/pmtiles_renderer.js
@@ -42,25 +42,15 @@ export class PmtilesRenderer {
         url: `pmtiles://${this.pmtilesUrl}`
       })
 
-      // use the first layer found
-      this.map.addLayer({
-        id: "pmtiles-layer",
-        type: "fill",
-        source: "pmtiles-source",
-        "source-layer": metadata.vector_layers[0].id,
-        paint: {
-          "fill-color": "steelblue",
-          "fill-opacity": 0.75
-        }
-      })
+      const interactiveLayers = this.drawLayers(
+        metadata.vector_layers.map(layer => layer.id)
+      )
 
       // Hover tooltip popup
       const popup = new maplibregl.Popup({
         closeButton: false,
         closeOnClick: false
       })
-
-      const interactiveLayers = ["pmtiles-layer"]
 
       // Show a tooltip on mouse move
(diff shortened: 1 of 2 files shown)


app/javascript/geo/pmtiles_renderer.js
```diff
@@ -85,14 +75,78 @@ export class PmtilesRenderer {
...
+          const feature = e.features[0]
+          this.highlightFeature(feature)
+          this.inspection(feature.properties)
+        })
+      })
+    })
+  }
+
+  drawLayers(layers) {
+    // use the first layer found
```
thatbudakguy:
```suggestion
```
No longer true?

config/settings/development.yml
```diff
@@ -113,8 +113,12 @@ sandbox:
     #   purl: sf815vr1246
     - label: Index map
       purl: kh860qj3406
-    - label: PMTiles
+    - label: PMTiles fill
```
thatbudakguy:
```suggestion
    - label: PMTiles polygons
```
"fill" is a maplibre style term; the data themselves are polygons

thatbudakguy (commented):
This works great for hf224mw4004 (the color combination looks very trippy), but it still doesn't display the points object for me in local dev – I only see its bounding box.

Image

jcoyne:
@thatbudakguy the points object doen't yet have a PMTile generated. Waiting on the unzip script.

thatbudakguy (approved):
Points PMTile successfully unzipped; have confirmed this works.

Issue #3094, opened 2026-06-25
COGs don't render because of missing importmap pins


thatbudakguy:
This causes COGs to not render. Visible in the browser console on https://purl.stanford.edu/bb223nv3920.

thatbudakguy:
Might be able to be fixed by just pinning it in the importmap?

PR #3098, opened 2026-06-29
Add pin for lzw and utils

jcoyne:


thatbudakguy (approved):
I updated #3094 because this removed one error, but now I see:
> TypeError: The specifier “utils” was a bare specifier, but was not remapped to anything.

So I think we need to keep pinning until records like bb223nv3920 display.

PR #3106, opened 2026-07-10
Add prettier to format javascript

jcoyne:


marlo-longley:
@jcoyne Adding Prettier seems like a fine move but this brings up the same question I had on the Mirador implementation. Here, running `lint` won't catch Prettier issues. Someone has to know to run the format command to keep it in sync, and CI/lint doesn't show issues. That is why I added this PR, which brings it in to the existing flow and enforces the rules: https://github.com/ProjectMirador/mirador/pull/4325

Thoughts?

jcoyne:
@marlo-longley doesn't https://github.com/sul-dlss/sul-embed/pull/3106/changes#diff-ce0dee93a528f6c7648f4cd671a3ec7be6a80f976c88f3aa1c322b7a80828fbaR40 take care of it?

marlo-longley:
@jcoyne I see, then this is brining up a larger question for me. Seems like JS tests (npm run test) and lint (npm run lint) are never run in CI. You have to know to run them locally and they aren't enforced? I would expect Prettier to be part of an overall JS test and lint flow, that is able to be run locally and also enforced in CI. The ruby flow runs `['test:prepare', :spec, :rubocop]` together. I'm not trying to overcomplicate this Prettier PR but trying to understand the overall JS testing picture.

I could merge this but wonder if there is follow on work.

jcoyne:
@marlo-longley this PR is only about adding prettier. It does not attempt to answer any of the other questions you raised.

thatbudakguy (approved):

PR #3119, opened 2026-07-29
Use access methods from cocina-display

### LINKED ISSUE #3111: Refactor to use cocina_display more thoroughly
Most of this is concentrated in app/models, in particular `Purl`, `PurlJsonLoader`, and anything to do with files.

If there are methods here that aren't in cocina_display that might be useful for other apps, we should move them into the gem.

jcoyne:
Fixes #3111


app/models/embed/purl_json_loader.rb
```diff
@@ -64,29 +64,27 @@ def restricted_location
...
-      json.dig('access', 'download')
+      cocina_record.download_rights
     end
 
     def view
-      json.dig('access', 'view')
+      cocina_record.view_rights
     end
 
     def controlled_digital_lending
```
thatbudakguy, line 73: hasn't CDL been removed? I thought we were no longer doing it

app/models/embed/purl_json_loader.rb
```diff
@@ -115,32 +113,26 @@ def layer_type
...
-
-      Array(json.dig('structural', 'contains')).map do |file_set_json|
-        Purl::ResourceJsonDeserializer.new(@druid, file_set_json).deserialize
+      cocina_record.filesets.map do |file_set|
+        Purl::ResourceJsonDeserializer.new(@druid, file_set.cocina).deserialize
       end
     end
 
     # @return [Array<String>] the list of identifiers for virtual object constituents
     def constituents
```
thatbudakguy: I think this is `CocinaRecord#virtual_object_members`

app/models/embed/purl_json_loader.rb
jcoyne: 🤷

jcoyne: It's still used by the mirador viewer in sul-embed.

thatbudakguy (approved):

PR #3131, opened 2026-08-25
Add tab with georeferenced maps to Mirador viewer instead of using Geo viewer

### LINKED ISSUE #3124: Improve viewer for georeferenced IIIF maps
Note this issue only applies to items with content-type `map` for which there is an accompanying IIIF georeference JSON file with the appropriate `georeference` file role. We currently only have one: https://purl.stanford.edu/bb013fz9675, which was created for testing. However, there are a huge number of David Rumsey maps for which we have georeference data, so this could be massively expanded in the future.

The test object https://purl.stanford.edu/bb013fz9675 is part of a [single map series in Searchworks](https://searchworks.stanford.edu/view/2941954) and is not released on its own. This is notable because it displays using the map viewer, which sets it apart visually from its siblings in the digital collection display.

After discussion with stakeholders, including staff from Rumsey and Branner, we would like to improve the display for these items by defaulting to the regular IIIF map viewer, so that there's less obvious contrast between siblings that are/are not georeferenced. The map is still a nice to have, though, and we discussed doing it using a tabbed interface. ogm-viewer does something like this for records that have multiple different previews:

Image

One consequence of this is that the metadata shown in the viewer sidebar could potentially differ unless we decide how to handle it. When we're showing a IIIF object, the metadata in the viewer sidebar comes from the manifest – but where does it come from when we show the map viewer? The Cocina? Care should be taken so this doesn't create confusion.

Something that deserves more thought: how to handle paged objects with georeference annotations?

thatbudakguy:
- **Make importmap use package.json for its ogm-viewer pin**
  This prevents Mirador's usage of the map and the Geo viewer's
  usage from drifting out of sync.
  
  Also updates ogm-viewer to 1.0.0.
  

- **Remove responsibility for georeferenced maps from the Geo viewer**
  

- **Render georeferenced maps as a second tab in Mirador viewer**
  Closes #3124

### DIFF
--- a/config/importmap.rb
+++ b/config/importmap.rb
@@ -42,18 +42,15 @@ pin "@google/model-viewer", to: "https://cdn.jsdelivr.net/npm/@google/model-view
 pin "three", to: "https://ga.jspm.io/npm:three@0.163.0/build/three.module.js"
 pin "popper", to: 'https://cdn.jsdelivr.net/npm/@popperjs/core@2.11.8/dist/esm/popper.js'
 
-# Two pins that go into the same location: one is the components, the other is
-# the classes used to build Resources and Previewers to put into them. Pinning
-# into the same path means we only get one copy of maplibre-gl.
-#
-# The version is read from package.json rather than written here, because Mirador's georeference tab
-# imports the same library as a bundled dependency and Vite resolves that from package.json. Declared
-# twice, a bump to one alone would serve two different viewers to two views of one object. Vendoring
-# would remove the URL entirely, but this library's dist is 64 chunks importing each other by
-# relative path, and Propshaft digests filenames without rewriting JS imports - so every chunk but
-# the entry would 404 once precompiled.
+# The mirador viewer needs a local copy of ogm-viewer in node_modules to bundle,
+# but the geo viewer uses importmaps, and Propshaft can't serve the node_modules
+# version in a way that plays nice in the browser. Since we're stuck with two
+# copies, make the package.json version authoritative, so they don't get out of sync.
 ogm_viewer_version = JSON.parse(Rails.root.join('package.json').read).fetch('dependencies').fetch('ogm-viewer')
 ogm_viewer = "https://cdn.jsdelivr.net/npm/ogm-viewer@#{ogm_viewer_version}/dist/components"
 
+# Two pins that go into the same location: one is the components, the other is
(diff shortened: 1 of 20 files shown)


thatbudakguy:
# First tab
Screenshot 2026-08-25 at 15 06 06

# Second tab
Screenshot 2026-08-25 at 15 06 12

config/importmap.rb
```diff
@@ -45,5 +45,15 @@
...
-pin "ogm-viewer", to: "https://cdn.jsdelivr.net/npm/ogm-viewer@1.2.0/dist/components/ogm-viewer.js"
-pin "ogm-viewer/lib", to: "https://cdn.jsdelivr.net/npm/ogm-viewer@1.2.0/dist/components/index.js"
+#
+# The version is read from package.json rather than written here, because Mirador's georeference tab
+# imports the same library as a bundled dependency and Vite resolves that from package.json. Declared
+# twice, a bump to one alone would serve two different viewers to two views of one object. Vendoring
+# would remove the URL entirely, but this library's dist is 64 chunks importing each other by
+# relative path, and Propshaft digests filenames without rewriting JS imports - so every chunk but
+# the entry would 404 once precompiled.
+ogm_viewer_version = JSON.parse(Rails.root.join('package.json').read).fetch('dependencies').fetch('ogm-viewer')
```
jcoyne, line 49: I'm having trouble understanding this. One version is from jsdeliver CDN and the other version is packaged by vite (from the NPM registry), aren't they effectively different packages even if they started from the same src?  Why is this so important?

jcoyne (changes requested):

config/importmap.rb
thatbudakguy: It's just consistency so that the pins don't drift out of sync. We're stuck with two versions because Mirador needs the local one to bundle, but the Geo viewer uses importmaps, and Propshaft can't serve the local one in a way that plays nice in the browser.

thatbudakguy: I picked the package.json to be the authoritative version, which is what this does.

jcoyne: That makes sense. Can you replace the current comments here with what you just said here?

thatbudakguy: Took a crack at rewriting this. The comment above it also applied to the stuff that's now below, so I moved that too.

jcoyne (approved):

